PR #10, opened 2018-10-17
Bump to crd

ingvagabund:
@enxebre @vikaschoudhary16 FYI

With this vendored in @enxebre 's PR (https://github.com/openshift/cluster-api-provider-aws/pull/67) with minor modifications I was able to run the first e2e.


pkg/e2e/framework/clusterapi.go
```diff
@@ -89,103 +95,84 @@ func (f *Framework) DeployClusterAPIStack(clusterAPINamespace, actuatorImage, ac
...
 
-	f.By("Deleting api server certs")
-	err = WaitUntilDeleted(func() error {
-		return f.KubeClient.CoreV1().Secrets(certsSecret.Namespace).Delete(certsSecret.Name, &metav1.DeleteOptions{})
-	}, func() error {
-		_, err := f.KubeClient.CoreV1().Secrets(certsSecret.Namespace).Get(certsSecret.Name, metav1.GetOptions{})
-		return err
-	})
-	f.ErrNotExpected(err)
+	// Do not delete the CRDs as they are always in the default namespace.
```
enxebre: can we just remove this before merging?

ingvagabund: It still could be a valid use case. Later, I would like to if condition this part with:
```go
if deleteCRDs {
...
}
```

and `DestroyClusterAPIStack(clusterAPINamespace, actuatorImage, actuatorPrivateKey string, deleteCRDs bool)`. So rather keeping the code in comment for now.

pkg/manifests/manifests.go
```diff
@@ -5,239 +5,885 @@ import (
...
+			Name:     "manager-role",
+		},
+	}
+}
+
+func ManagerManifest(clusterAPINamespace, managerImage string) *appsv1.StatefulSet {
+	var terminationGracePeriodSeconds int64 = 10
+	return &appsv1.StatefulSet{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "StatefulSet",
```
enxebre: why does this need to be an statefulset? couldn't it be a deployment

ingvagabund: Copy of upstream yaml. Not sure what was the reason to change deployment to statefulset.

PR #16, opened 2018-10-31
Fix machine crd

vikaschoudhary16:
Include `Status` as a subresource in machine object


pkg/manifests/manifests.go
```diff
@@ -174,6 +174,9 @@ func MachineCRDManifest() *v1beta1.CustomResourceDefinition {
 				Kind:   "Machine",
 			},
 			Scope: "Namespaced",
+			Subresources: &v1beta1.CustomResourceSubresources{
```
enxebre: can we please add the same for machineSet and machineDeployment

enxebre:
follow up here https://github.com/openshift/cluster-api-actuator-pkg/pull/18
/approve

PR #20, opened 2018-11-19
Allow the machine-role clusterrole to list secrest as well

ingvagabund:
The machine controller reads a user data from a secret.
Thus, the pod running the machine controller must be allowed to list secrets.

/hold
until verified in https://github.com/openshift/cluster-api-provider-libvirt/pull/92

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -674,6 +674,19 @@ func ClusterRoleManifest() *rbacv1.ClusterRole {
 					"nodes",
 				},
 			},
+			{
+				Verbs: []string{
+					"get",
+					"list",
+					"watch",
+				},
+				APIGroups: []string{
+					"",
+				},
+				Resources: []string{
+					"secrets",
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


ingvagabund:
/hold cancel
Just verified with https://ci.openshift.redhat.com/jenkins/job/pull-ci-openshift-cluster-api-provider-libvirt-master-e2e/130/console

pkg/manifests/manifests.go
```diff
@@ -674,6 +674,19 @@ func ClusterRoleManifest() *rbacv1.ClusterRole {
 					"nodes",
 				},
 			},
```
enxebre: What's the motivation behind this? Shoudn't this be already implicit by https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/controller_policy.go#L231 and https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/controller_policy.go#L161

ingvagabund: We don't run the bootstrap policy in actuator e2e. Thus, we need to create the clusterrole explicitly.

PR #25, opened 2019-02-01
Openshift machine api

enxebre:
We are mirroring the API types under machine.openshift.io group. For the sake of transparency we are making this explicit in the import path. This allows both libraries upstream and openshift ot be consumed simultaneously by other tooling. We aim to consilidate back with upstream once there's a relase stable enough


enxebre:
cc @ingvagabund

enxebre:
Once this get merge we can update the harcoded changes in the vendor via https://github.com/openshift/cluster-api-provider-aws/pull/147

PR #26, opened 2019-02-01
Update left pointers from clusterv1alpha1 to machinev1beta1

enxebre:
Follow up for https://github.com/openshift/cluster-api-actuator-pkg/pull/25


enxebre:
cc @bison @ingvagabund

ingvagabund:
Grep says those are exactly the last bits left

/approve
/lgtm

PR #28, opened 2019-02-20
E2e

enxebre:
Unify machine api e2e suites
See Readme https://github.com/openshift/cluster-api-actuator-pkg/pull/28/commits/39881247af730c61da184e66369f811336985336


enxebre:
cc @ingvagabund @bison

pkg/e2e/openshift/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Non Goals
+
+- Validate behaviour of any component of the system completely unrelated to the machine API domain
+
+- Pre create Openshift/kubernetes clusters to run the expectations against it
+
+## Implementation
+
+Currently pure golang is used for running suite expectations but this can be moved to a framework eg. ginkgo as requirement rises
```
ingvagabund: `s/eg./e.g./`

ingvagabund:
`pkg/e2e/openshift` is to restrictive, `pkg/e2e` would serve the same purpose. The goal is to be able to run the tests over any Kubernetes distribution (not just OpenShift) where machine API stack is deployed.

I would expect the following layout:
```
- pkg/e2e/machine-api-operator
- pkg/e2e/cluster-autoscaler-operator
- pkg/e2e/actuator
- pkg/e2e/machine-healthcheck-controllers or pkg/e2e/machinepool
- pkg/e2e/common for remaining tests (e.g. nodelink controller) until there is enough tests to create separate category/component
```

pkg/e2e/openshift/README.md
ingvagabund: What's the benefit of the current implementation over ginkgo framework? We are already at a point where it's time to reuse ginkgo:
- to generate junit files
- to specify subset of tests to be run
- better readability of the code (everything else is part of framework initialization)
- community is already familiar with the technology
- support for various assertions we don't need to implement
- order of tests can by randomized before running
- tests can be run in parallel (which may discover potential flakes/bugs) compare to running them in series as we have now

enxebre:
@ingvagabund to clarify, the goal is to take any duplicated e2e testing we have today and centralised it here.
So we can verify that any product user expectation is not broken by one or more components that belong to the machine API domain

This must be invariant. These tests must run across all repos with no option to be disabled in order to validate the holistic expected behaviour of the system which might be broken by any subcomponent. If your component interfere some how with machine API domain and it's not running this, that's bad

The implementation is intentionally simple. Once we merge this and all repos are consuming from here a second iteration for implementation details, e.g layout, ginkgo... can take place

ingvagabund:
> This must be invariant. These tests must run across all repos with no option to be disabled in order to validate the holistic expected behaviour of the system which might be broken by any subcomponent. If your component interfere some how with machine API domain and it's not running this, that's bad

Not in disagreement here. That's why I am for `Conformance` tests that will be run over every PR of every repo (including this) we have. However, you can't run tests written for libvirt over aws. Thus, the need for saying which tests are not to be run and which are.

> The implementation is intentionally simple. Once we merge this and all repos are consuming from here a second iteration for implementation details, e.g layout, ginkgo... can take place

When is the second iteration going to take place? The later we do it, the harder it will be to integrate the tests into ginkgo since the number of tests will grow over time. I am still in favor of deciding if we are going to use ginkgo now or not before we merge the PR. And if we decide to go with ginkgo, start with ginkgo from scratch.

enxebre:
> When is the second iteration going to take place? The later we do it, the harder it will be to integrate the tests into ginkgo since the number of tests will grow over time. I am still in favor of deciding if we are going to use ginkgo now or not before we merge the PR. And if we decide to go with ginkgo, start with ginkgo from scratch.

The implementation of the tests is totally unrelated to this PR. Again the problem this is solving is "to take any duplicated e2e testing we have today and centralised it here. So we can verify that any product user expectation is not broken by one or more components that belong to the machine API domain"

Once we equalise here the coverage we have today duplicated across repos and this is consumed everywhere, it works, it's stable and we are confident, then any implementation change can take place

ingvagabund:
> The implementation of the tests is totally unrelated to this PR. Again the problem this is solving is "to take any duplicated e2e testing we have today and centralised it here.

The purpose of the PR can be changed any time.

> So we can verify that any product user expectation is not broken by one or more components that belong to the machine API domain"

We can reach the same goal even with the ginkgo framework.

> Once we equalise here the coverage we have today duplicated across repos and this is consumed everywhere, it works, it's stable and we are confident, then any implementation change should take place

Every time we change the underlying framework (pure golang, ginkgo, I-do-not-care-what-the-name-of-the-framework-is framework), we will need to stabilize and get confident again. Let's do it now when we can start from scratch and iteratively add more tests.

And in the opposite way. First, let's run the "new" tests (or all tests grabbed from all the repos) just here (using both kubernetes+kubemark and openshift). Once all the tests are green and running, we can start vendoring the new framework into other repos.

bison:
Looks like the `go vet` run is picking up stuff from vendor.

ingvagabund:
/lgtm

for @bison

PR #29, opened 2019-02-20
More tests with ginkgo

ingvagabund:


pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,441 @@
+package actuators
+
```
enxebre: This is called actuators but the fact that this is driven by the actuator is circumstantial today, this could be e.g the machine controller. This should belong to the same suite there's in infra and may be rename it to "elastic compute" or similar

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,441 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	workerRoleLabel = "node-role.kubernetes.io/worker"
```
enxebre: these values are duplicated in operator common. Can we have some common utils?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,371 @@
+package infra
```
enxebre: infra? is this an accidental carry over?
In any case is there any reason to have operator/autoscalers/infra/actuators.
I think they could all belong to same package e.g "elastic compute" and then have independent files and a common utils

pkg/e2e/framework/framework.go
```diff
@@ -9,18 +10,20 @@ import (
...
 	appsv1beta2 "k8s.io/api/apps/v1beta2"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	// apiregistrationclientset "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
```
enxebre: can we remove this comment?

pkg/e2e/operators/common.go
```diff
@@ -0,0 +1,9 @@
+package operators
+
+import "time"
+
+const (
```
enxebre: this is duplicated across different packages.

enxebre:
thanks!! Dropped some comments. I don't see the need for having different packages, I'd be in favour of having all the suite in the same package e.g "elastic compute" and different files one common utils.
Since this is overlapping `e2e/openshift` can you remove those? Also could you elaborate a bit the motivation and improvements of this PR in the description?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,371 @@
...
+				Name:      "default",
+				Namespace: e2e.TestContext.MachineApiNamespace,
+			},
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "ClusterAutoscaler",
+				APIVersion: "autoscaling.openshift.io/v1alpha1",
+			},
+			Spec: caov1alpha1.ClusterAutoscalerSpec{
+				ScaleDown: &caov1alpha1.ScaleDownConfig{
+					Enabled:           true,
```
enxebre: the autoscaler is expecting pointers to strings here, so this will break

Gopkg.lock
```diff
@@ -326,6 +326,9 @@
     "pkg/client/clientset_generated/clientset/scheme",
```
enxebre: could you put the vendoring commits in its own one?

Makefile
```diff
@@ -25,7 +25,7 @@ check: fmt vet lint test ## Check your code
 
 .PHONY: test
 test: # Run unit test
-	$(DOCKER_CMD) go test -race -cover ./pkg/...
+	$(DOCKER_CMD) go test -race -cover `go list ./... | grep -v github.com/openshift/cluster-api-actuator-pkg/pkg/e2e`
```
frobware: Why don't we be explicit here? Just run the named package?

Makefile
```diff
@@ -39,6 +39,27 @@ fmt: ## Go fmt your code
...
+		-v github.com/openshift/cluster-api-actuator-pkg/pkg/e2e \
+		-kubeconfig $${KUBECONFIG:-~/.kube/config} \
+		-machine-api-namespace openshift-machine-api \
+		-ginkgo.v \
+		-args -v 5 -logtostderr true
+
+.PHONY: k8s-e2e
+k8s-e2e: ## Run k8s specific e2e test
+	# KUBECONFIG and SSH_PK dirs needs to be mounted inside a container if tests are run in containers
+	go test -timeout 30m \
```
frobware: I think the `test-e2e` and the `k8s-e2e` only differ in the namespace. We could create a Makefile macro here and reduce the duplication.

pkg/e2e/actuators/actuators.go
frobware: nit: Anything but `common` or `utils`. Something with more explicit naming please.

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+	for key := range nodeDrainLabels {
+		if _, exists := machine.Spec.ObjectMeta.Labels[key]; exists {
+			continue
+		}
+		machine.Spec.ObjectMeta.Labels[key] = nodeDrainLabels[key]
+	}
+	return machine
+}
+
+func isNodeReady(node *corev1.Node) bool {
```
frobware: Can't believe you bailed on the 67MB import from kube for this... :-D

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+	for _, c := range node.Status.Conditions {
+		if c.Type == corev1.NodeReady {
+			return c.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
+
+func waitUntilNodesAreReady(client runtimeclient.Client, listOpt *runtimeclient.ListOptions, nodeCount int) error {
```
frobware: I think we should pass in a `Context` here - start with that in mind so that we can do cancellation.

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+	return wait.PollImmediate(retryMedium, waitLong, func() (bool, error) {
+		nodes := corev1.NodeList{}
+		if err := client.List(context.TODO(), listOpt, &nodes); err != nil {
+			glog.Errorf("Error querying api for Node object: %v, retrying...", err)
+			return false, nil
+		}
+		// expecting nodeGroupSize nodes
+		nodeCounter := 0
```
frobware: How about `readyNodes`?

pkg/e2e/actuators/actuators.go
ingvagabund: I did not want to overload the word controller. It's true machine controller is "just" a wrapper around an actuator. Here, `actuator` is shorter than `machinecontroller` or `elasticcompute`. Plus, elastic compute is process oriented term compared to actuator which is component oriented. To goal was to break the tests into component-wise buckets. Infra homes all tests of which I was not sure what would be the right component to put them to. More or less following what I described in the test design google doc.

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+
+			nodeCounter++
+		}
+
+		if nodeCounter < nodeCount {
+			glog.Errorf("Expecting %v nodes with %#v labels in Ready state, got %v", nodeCount, nodeDrainLabels, nodeCounter)
+			return false, nil
+		}
+
+		glog.Infof("Expected number (%v) of nodes with %v label in Ready state found", nodeCount, nodeDrainLabels)
```
frobware: Shouldn't `nodeCount` be `nodeCounter`?

pkg/e2e/actuators/actuators.go
ingvagabund: Yeah, I will create `pkg/e2e/common` package.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: > infra? is this an accidental carry over?
Right, it should be `autoscaler`.

> In any case is there any reason to have operator/autoscalers/infra/actuators.
I think they could all belong to same package e.g "elastic compute" and then have independent files and a common utils

We are not exposing new API so in this case I prefer to allocate one directory/package for each important component. I still prefer smaller packages with few files rather than one package with a lots of files.

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+
+			// The point of the test is to make sure majority of the pods is rescheduled
+			// to other nodes. Pod disruption budget makes sure at most one pod
+			// owned by the RC is not Ready. So no need to test it. Though, usefull to have it printed.
+			glog.Infof("RC ReadyReplicas/Replicas: %v/%v", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
+
+			// This makes sure at most one replica is not ready
+			if rcObj.Status.Replicas-rcObj.Status.ReadyReplicas > 1 {
+				return false, fmt.Errorf("pod disruption budget not respecpted, node was not properly drained")
```
frobware: typo: "respected"

pkg/e2e/actuators/actuators.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+
+			// Depends on timing though a machine can be deleted even before there is only
+			// one pod left on the node (that is being evicted).
+			if podCounter > 2 {
+				glog.Infof("Expecting at most 2 pods to be scheduled to drained node %v, got %v", machine.Status.NodeRef.Name, podCounter)
```
frobware: nit: `%q` for names.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund:
```go
type ScaleDownConfig struct {
	Enabled           bool   `json:"enabled"`
	DelayAfterAdd     string `json:"delayAfterAdd"`
	DelayAfterDelete  string `json:"delayAfterDelete"`
	DelayAfterFailure string `json:"delayAfterFailure"`
	UnneededTime      string `json:"unneededTime,omitempty"`
}
```

Gopkg.lock
ingvagabund: Sure

pkg/e2e/actuators/actuators.go
enxebre: the suite you are reimplementing belongs all together. It's validating holistic features that result from the interaction of one or more of the components that provide the e2e machine API experience. Actuator is an implementation detail that we don't care about in this context cc 
We should have only one suite pkg e.g "elastic compute", may be different files and a common utils
@bison @frobware

pkg/e2e/autoscaler/autoscaler.go
enxebre: Not sure where you got that from. Please see https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go#L67

pkg/e2e/actuators/actuators.go
ingvagabund: We should have to set `BeforeEach` for each `g.Describe` like this (most of the code can be put into a function and called, yet still needs to be repeated):

```go
	var describeContext context.Context

	g.BeforeEach(func() {
		var cancel context.CancelFunc

		g.By("Creating context")

		stopCh := signals.SetupSignalHandler()
		describeContext, cancel = context.WithCancel(context.Background())
		go func() {
			select {
			case <-stopCh:
				cancel()
			case <-describeContext.Done():
			}
		}()
	})
```

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Changed by this commit [1]. The `vendor` dir has old version of CAO code.

[1] https://github.com/openshift/cluster-autoscaler-operator/commit/2a05e7b63d9eedcfb084262bbdd138c57a4723da#diff-10d8058f9e03d0da119d441627f8ac4d

pkg/e2e/actuators/actuators.go
bison: I tend to agree that these are meant to be high-level E2E, so things should be named more after groups of features / use-cases / stories than components. Not really sure what the right thing is exactly though. Any example in upstream Kubernetes or something we can look to?

pkg/e2e/autoscaler/autoscaler.go
enxebre: yes, so it needs a newer version so we can run the tests against cao

pkg/e2e/actuators/actuators.go
ingvagabund: https://github.com/kubernetes/kubernetes/tree/master/test/e2e

Makefile
ingvagabund: other packages might have unit tests worth running so I am just skipping anything but e2e

Makefile
ingvagabund: Any example?

pkg/e2e/actuators/actuators.go
ingvagabund: > We should have only one suite pkg e.g "elastic compute", may be different files and a common utils

There is only one test suite. Under the hood broken into more packages for better readability and manipulation.

1. To be clear the conflict of interest is to decide if we are going to break the single test suite component-wise or process-wise. The test suite for me here is "Conformance". We might even go with `pkg/e2e/conformance/...` layout to explicitly say the tests fall under the category (though I don't consider that necessary).

   I am fine with both (component and process). Any hints on what would be the names/identifiers we might use for holistic process-wise categorization? I could imagine:
   - `autoscaling`
   - `machinereconciling`
   - anything else?

   `elasticcompute` is too generic buzzword that does not say anything.

2. Does the conflict also includes the question of putting the tests under multiple files under one package instead of how it is currently done?

   I still want to break the individual tests into packages so you can import subset of tests in different test suite (e.g. `Performance`). If you check the import part of `e2e_test.go`, you can see:

   ```go
   import (
   	...
   	_ "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/actuators"
   	_ "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/autoscaler"
   	_ "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/infra"
   	_ "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/operators"
   )
   ```

   I could have `pkg/e2e/performance/e2e_test.go` that will import just operators (to assert both operators are available and working). With all tests under the same package (just different files) you can import either all the tests or no tests. And I don't want to import all conformance tests into the performance test suite.

   And again, it's easier to see intention of a set of tests if they are imported from e.g. `actuators` or `autocaler` then to just have:
   ```go
   import (
   	...
   	_ "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e"
   )
   ```

---
*Bottom line*: On the other hand, one could argue we can put all the tests under one huge test suite and use `[Conformance]` and `[Performance]`. Doing that, we would need to put the keyword to every `Describe` or spec part to make sure only conformance, resp. performance tests are run when `--ginkgo.focus` is set to `[Conformance]`, resp. `[Performance]` focus. Though, this can be discussed later once we have more test suites.

enxebre: 1 - This must be process/feature wise, it does not care about components. Conformance/machines/smoke/elasticCompute I don't have strong opinions on the naming

2 - "you can import either all the tests or no tests." this is good and it's exactly the invariant we want to force. This suite is the gate for any of repo which might have an impact on the user feature validation

ingvagabund: > 1 - This must be process/feature wise, it does not care about components. Conformance/machines/smoke/elasticCompute I don't have strong opinions on the naming

`Conformance` is the test suite name, `machines`, `autoscaling`, ... would be test suite groups. The question is what would be the group names?

> 2 - "you can import either all the tests or no tests." this is good and it's exactly the invariant we want to force. This suite is the gate for any of repo which might have an impact on the user feature validation

This is orthogonal to the question. All tests are to be run as part of `Conformance` test suite. Inseparably from each other. The intent is to:
- allow other test suites to reuse subset of tests
- decompose one huge test suite into smaller chunks (groups) so it's easier to maintain, read and extend the code

pkg/e2e/actuators/actuators.go
ingvagabund: `nodeCount` is the expected number (it could be even more in real though it needs to be at least `nodeCount`)

pkg/e2e/actuators/actuators.go
ingvagabund: Let's discuss this and expectations in the next iteration of the tests. Right now, I am not sure how we should approach this.

Makefile
ingvagabund: We might have a single `make e2e-operator` or similar and run `make e2e-operator NAMESPACE=kube-system` so there is only a single point of entry. This will require to update the Prow job first. Let's do it in the next iteration so we can start adding new tests to the new framework.

Gopkg.lock
ingvagabund: Done

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Done

enxebre:
This looks good enough to me. Once we merge we can sync all repos to run this and follow up with any farther improvement

PR #36, opened 2019-02-28
[autoscaling]: print which machines actually reference a node

ingvagabund:
To be able to count how many machines are in a node group
and how many nodes are actually linked. Autoscaling scale out
is failing either due to a node taking too much time to register
or for node labels not being properly propagated from machines.


ingvagabund:
```

PR #37, opened 2019-03-01
Operators

enxebre:
- Revendor for using recommended openshift/library-go instead of old cvo to manipulate status
- Use common functions for checking deployments and status


pkg/e2e/operators/machine-api-operator.go
```diff
@@ -8,158 +8,69 @@ import (
...
+		o.Expect(isDeploymentAvailable(client, "machine-api-operator")).To(o.BeTrue())
 	})
 
 	g.It("reconcile controllers deployment", func() {
 		var err error
 		client, err := e2e.LoadClient()
 		o.Expect(err).NotTo(o.HaveOccurred())
 
+		controllersDeployment := "clusterapi-manager-controllers"
 		key := types.NamespacedName{
```
frobware: I think this would be better if we substituted `key` for `controllersDeployment`. Then where we consume/log we can use `controllersDeployment.Name`, et al.

PR #43, opened 2019-03-06
Scaling stories

enxebre:
Add Scaling stories

### DIFF
--- a/pkg/e2e/infra/infra.go
+++ b/pkg/e2e/infra/infra.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/e2e/infra/utils.go
```diff
@@ -118,6 +126,60 @@ func getMachineSetList(client runtimeclient.Client) (*mapiv1beta1.MachineSetList
...
+		return nil, err
+	}
+	return &machineList, nil
+}
+
+// getMachinesFromMachineSet returns an array of machines owned by a fiven machineSet
+func getMachinesFromMachineSet(client runtimeclient.Client, machineSet mapiv1beta1.MachineSet) ([]mapiv1beta1.Machine, error) {
+	machineList, err := getMachineList(client)
+	if err != nil {
+
```
frobware: Fill in the blanks!

enxebre: thanks, fixed

pkg/e2e/infra/infra.go
```diff
@@ -112,119 +111,88 @@ var _ = g.Describe("[Feature:Machines] Machines should", func() {
...
 
-	g.It("be replaced with new one when deleted", func() {
+	g.It("recover from deleted worker machines", func() {
 		var err error
 		client, err := e2e.LoadClient()
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		// Assume cluster state
+		// Expect for cluster to cool down from previous tests
 		err = e2e.WaitUntilAllNodesAreReady(client)
```
ingvagabund: This is still no assurance. Some nodes may turn to NotReady state after this line if the corresponding machine is deleted in a previous test. As long we we don't remove a node right before a machine is deleted, this will never be 100%.

pkg/e2e/infra/utils.go
```diff
@@ -118,6 +126,79 @@ func getMachineSetList(client runtimeclient.Client) (*mapiv1beta1.MachineSetList
 	return &machineSetList, nil
 }
 
+// getMachineSetList returns a MachineSetList object filtered by machineRoleLabel
+func getWorkerMachineSetList(client runtimeclient.Client) (*mapiv1beta1.MachineSetList, error) {
```
frobware: Unless we need access to either `TypeMeta` or `ListMeta` why not return the slice (i.e., `[]MachineSet`). 

And ditto for the other functions that return `<Foo>List`.

pkg/e2e/infra/utils.go
```diff
@@ -118,6 +126,79 @@ func getMachineSetList(client runtimeclient.Client) (*mapiv1beta1.MachineSetList
...
+	machineSetList := mapiv1beta1.MachineSetList{}
+	listOptions := runtimeclient.ListOptions{
+		Namespace: e2e.TestContext.MachineApiNamespace,
+	}
+
+	// See https://github.com/openshift/cluster-api-provider-kubemark/blob/master/examples/worker-machinesets.yaml#L45
+	// We plan to update the installer and consumers to rely only on applying 'node-role.kubernetes.io/worker'to machineSets
+	// in favour of 'machine.openshift.io/cluster-api-machine-role'
+	listOptions.MatchingLabels(map[string]string{nodeWorkerRoleLabel: ""})
```
ingvagabund: The worker node label is part of a machine spec, not machineset labels. This will not find any machineset.

pkg/e2e/infra/utils.go
```diff
@@ -118,6 +126,79 @@ func getMachineSetList(client runtimeclient.Client) (*mapiv1beta1.MachineSetList
...
+		return true, nil
+	}); err != nil {
+		glog.Errorf("Error calling getMachineSetListWorkers: %v", err)
+		return nil, err
+	}
+	if len(machineSetList.Items) > 0 {
+		return &machineSetList, nil
+	}
+
+	listOptions.MatchingLabels(map[string]string{machineRoleLabel: "worker"})
```
frobware: I think this function should be generalised as we have identical code here. Make the generalised form takes the `labels` you want to match then create helper functions for each label type or a helper that calls the generalised form and concatenates the results.

pkg/e2e/infra/utils.go
```diff
@@ -176,3 +291,57 @@ func getWorkerNode(client runtimeclient.Client) (*corev1.Node, error) {
...
+		return fmt.Errorf("error calling getScaleClient %v", err)
+	}
+
+	scale, err := scaleClient.Scales(e2e.TestContext.MachineApiNamespace).Get(schema.GroupResource{Group: machineAPIGroup, Resource: "MachineSet"}, name)
+	if err != nil {
+		return fmt.Errorf("error calling scaleClient.Scales get: %v", err)
+	}
+
+	scaleUpdate := scale.DeepCopy()
+	scaleUpdate.Spec.Replicas = int32(replicas)
```
frobware: Shouldn't this be assigning a pointer? If so, how does the test currently work?

enxebre: `scaleUpdate.Spec.Replicas` is int32 type, pointer is for machine type

pkg/e2e/infra/utils.go
```diff
@@ -155,6 +236,40 @@ func deleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) er
...
+	if machine.Status.NodeRef == nil {
+		glog.Infof("machine %q has no NodeRef", machine.Name)
+		return nil, nil
+	}
+	key := runtimeclient.ObjectKey{Namespace: machine.Status.NodeRef.Namespace, Name: machine.Status.NodeRef.Name}
+	if err := client.Get(context.Background(), key, &node); err != nil {
+		return nil, fmt.Errorf("error getting node %q: %v", node.Name, err)
+	}
+
+	glog.Infof("Machine %q is backing node %q", machine.Name, node.Name)
```
frobware: nit: 'is backed by'

pkg/e2e/infra/utils.go
```diff
@@ -176,3 +291,57 @@ func getWorkerNode(client runtimeclient.Client) (*corev1.Node, error) {
 	}
 	return &nodeList.Items[0], nil
 }
+
+// areNodesReady returns true if an array of nodes are all ready
+func areNodesReady(nodes []*corev1.Node) bool {
```
frobware: nit: if you change the name parts around it can read as:

```go
if nodesAreReady(...) {
```

pkg/e2e/infra/utils.go
enxebre: isn't the machine backing the node, and the node backed by the machine?

pkg/e2e/infra/utils.go
enxebre: just getting all machineSets now

pkg/e2e/infra/utils.go
enxebre: just getting all machineSets now

pkg/e2e/infra/utils.go
frobware: OK; I read it three ways. Happy as-is.

pkg/e2e/infra/infra.go
```diff
@@ -151,6 +153,98 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
+			if len(nodes) != scaleOut {
+				return false
+			}
+			return nodesAreReady(nodes)
+		}, e2e.WaitLong, 5*time.Second).Should(o.BeTrue())
+
+		o.Eventually(func() bool {
+			nodes, err := getNodesFromMachineSet(client, machineSet1)
+			o.Expect(err).NotTo(o.HaveOccurred())
+			if len(nodes) != scaleOut {
```
frobware: I think this can be reduced to: `return len(nodes) == scaleOut && nodesAreReady(nodes)`.

pkg/e2e/infra/utils.go
```diff
@@ -155,6 +209,41 @@ func deleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) er
...
+}
+
+// getNodeFromMachine returns the node object referenced by machine.Status.NodeRef
+func getNodeFromMachine(client runtimeclient.Client, machine mapiv1beta1.Machine) (*corev1.Node, error) {
+	var node corev1.Node
+	if machine.Status.NodeRef == nil {
+		glog.Errorf("Machine %q has no NodeRef", machine.Name)
+		return nil, fmt.Errorf("machine %q has no NodeRef", machine.Name)
+	}
+	key := runtimeclient.ObjectKey{Namespace: machine.Status.NodeRef.Namespace, Name: machine.Status.NodeRef.Name}
```
ingvagabund: Nit: nodes are cluster scope, no need to specify the namespace

PR #44, opened 2019-03-07
Improve Taints validation

enxebre:
[Feature:Machines] Managed cluster should - have ability to additively reconcile taints from machine to nodes
- Add nodes snapshot
- Update test to rely on emerging DSL api
- Ensure original taints are restored at the end of the test
Goes after https://github.com/openshift/cluster-api-actuator-pkg/pull/43

### DIFF
--- a/pkg/e2e/infra/infra.go
+++ b/pkg/e2e/infra/infra.go
@@ -9,9 +9,7 @@ import (
 	g "github.com/onsi/ginkgo"
 	o "github.com/onsi/gomega"
 	e2e "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/framework"
-	mapiv1beta1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


pkg/e2e/infra/utils.go
```diff
@@ -319,3 +319,32 @@ func getScaleClient() (scale.ScalesGetter, error) {
 	}
 	return scaleClient, nil
 }
+
+// nodesSnapShot logs the state of all the nodes in the cluster
+func nodesSnapShot(client runtimeclient.Client) error {
```
frobware: Rename this to capture its intent; it was only when I read the comment that I see it is related to gathering logs. `snapshotNodeLogs()`?

pkg/e2e/infra/utils.go
```diff
@@ -319,3 +319,32 @@ func getScaleClient() (scale.ScalesGetter, error) {
...
+		return fmt.Errorf("error calling getMachineSetList: %v", err)
+	}
+	for key := range nodeList.Items {
+		glog.Infof("Node %q. Ready: %t. Unschedulable: %t", nodeList.Items[key].Name, e2e.IsNodeReady(&nodeList.Items[key]), nodeList.Items[key].Spec.Unschedulable)
+	}
+	return nil
+}
+
+// getNodeList returns a nodeList object
+func getNodeList(client runtimeclient.Client) (*corev1.NodeList, error) {
```
frobware: As per some of the other recent changes it we don't need the MetaType info from NodeList then return a slice of nodes.

pkg/e2e/infra/infra.go
```diff
@@ -31,84 +29,68 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
 
-		g.By("Verify machine taints are getting applied to node")
-		err = func() error {
-			listOptions := runtimeclient.ListOptions{
-				Namespace: e2e.TestContext.MachineApiNamespace,
-			}
-			machineList := mapiv1beta1.MachineList{}
+		machines, err := getMachines(client)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		machine := machines[0]
```
frobware: Is this always guaranteed to have 1 machine? We may not error, but equally `len(machines)` may be zero.

enxebre: It's guaranty https://github.com/openshift/cluster-api-actuator-pkg/pull/44/files#diff-a8f8018153bd2312214f96573fe48e57R166

frobware: > It's guaranteed https://github.com/openshift/cluster-api-actuator-pkg/pull/44/files#diff-a8f8018153bd2312214f96573fe48e57R166

Then I think it is worth adding an assertion before accessing `machine[0]`. It makes the intent clearer (for new readers of the code). At this point in the test I know there is at least 1 machine.

ingvagabund: It should be up to consumer of `getMachines` to decide if zero number of machines is valid or invalid.

pkg/e2e/infra/utils.go
```diff
@@ -157,6 +157,18 @@ func getMachineList(client runtimeclient.Client) (*mapiv1beta1.MachineList, erro
 	return &machineList, nil
 }
 
+// getMachines returns the list of machines or an error
+func getMachines(client runtimeclient.Client) ([]mapiv1beta1.Machine, error) {
+	machineList, err := getMachineList(client)
+	if err != nil {
+		return nil, fmt.Errorf("error getting machineList: %v", err)
+	}
+	if len(machineList.Items) == 0 {
```
frobware: On reflection, this doesn't seem right. There's no actual error here. There's just no machines. And that's perfectly legitimate.

ingvagabund:
Looks fine

pkg/e2e/infra/utils.go
ingvagabund: As per https://github.com/openshift/cluster-api-actuator-pkg/pull/44/files#r263715365, it's better to have consumer of `getMachines` to decide if this is an error or not.

pkg/e2e/infra/utils.go
```diff
@@ -319,3 +319,41 @@ func getScaleClient() (scale.ScalesGetter, error) {
...
+		}
+		return true, nil
+	}); err != nil {
+		glog.Errorf("Error calling getMachineSetList: %v", err)
+		return nil, err
+	}
+	return &nodeList, nil
+}
+
+// getMachines returns the list of machines or an error
```
frobware: Should be getNodes.

frobware: And should be "returns the list of nodes, or an error".

frobware:
/lgtm

and as @ingvagabund as been through this as well

/approve

PR #49, opened 2019-03-13
Code improvements

ingvagabund:
- simplifying the code
- making the code more readable

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -237,13 +237,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 
 		glog.Info("Get nodeList")
 		nodeList := corev1.NodeList{}
-		err = wait.PollImmediate(1*time.Second, e2e.WaitMedium, func() (bool, error) {
-			if err := client.List(context.TODO(), runtimeclient.MatchingLabels(map[string]string{nodeTestLabel: ""}), &nodeList); err != nil {
-				glog.Errorf("error querying api for nodeList object: %v, retrying...", err)
-				return false, err
-			}
-			return true, nil
-		})
+		err = client.List(context.TODO(), runtimeclient.MatchingLabels(map[string]string{nodeTestLabel: ""}), &nodeList)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		nodeGroupInitialTotalNodes := len(nodeList.Items)
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -236,7 +236,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 		err = client.Create(context.TODO(), clusterAutoscaler)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		initialNumberOfReplicas := pointer.Int32PtrDerefOr(targetMachineSet.Spec.Replicas, 0)
+		initialNumberOfReplicas := pointer.Int32PtrDerefOr(targetMachineSet.Spec.Replicas, e2e.DefaultMachineSetReplicas)
 
 		g.By("Creating MachineAutoscaler objects")
 		machineAutoscaler := machineAutoscalerResource(&targetMachineSet, 1, initialNumberOfReplicas+1)
@@ -266,8 +266,8 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 				glog.Errorf("error querying api for machineset object: %v, retrying...", err)
 				return false, nil
 			}
-			glog.Infof("MachineSet %s. Initial number of replicas: %d. Current number of replicas: %d", targetMachineSet.Name, initialNumberOfReplicas, pointer.Int32PtrDerefOr(ms.Spec.Replicas, 0))
-			return pointer.Int32PtrDerefOr(ms.Spec.Replicas, 0) > initialNumberOfReplicas, nil
+			glog.Infof("MachineSet %s. Initial number of replicas: %d. Current number of replicas: %d", targetMachineSet.Name, initialNumberOfReplicas, pointer.Int32PtrDerefOr(ms.Spec.Replicas, e2e.DefaultMachineSetReplicas))
+			return pointer.Int32PtrDerefOr(ms.Spec.Replicas, e2e.DefaultMachineSetReplicas) > initialNumberOfReplicas, nil
 		})
 		o.Expect(err).NotTo(o.HaveOccurred())
 
@@ -315,7 +315,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 				glog.Errorf("error querying api for machineSet object: %v, retrying...", err)
 				return false, nil
 			}
-			msReplicas := pointer.Int32PtrDerefOr(ms.Spec.Replicas, 0)
+			msReplicas := pointer.Int32PtrDerefOr(ms.Spec.Replicas, e2e.DefaultMachineSetReplicas)
 			glog.Infof("Initial number of replicas: %d. Current number of replicas: %d", initialNumberOfReplicas, msReplicas)
 			if msReplicas > initialNumberOfReplicas {
 				return false, nil
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -268,7 +268,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 			}
 			ms := &mapiv1beta1.MachineSet{}
 			if err := client.Get(context.TODO(), msKey, ms); err != nil {
-				glog.Errorf("error querying api for clusterAutoscaler object: %v, retrying...", err)
+				glog.Errorf("error querying api for machineset object: %v, retrying...", err)
 				return false, nil
 			}
 			glog.Infof("MachineSet %s. Initial number of replicas: %d. Current number of replicas: %d", targetMachineSet.Name, initialNumberOfReplicas, pointer.Int32PtrDerefOr(ms.Spec.Replicas, 0))
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -251,13 +251,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 
 		glog.Info("Create workload")
 		workLoad := newWorkLoad()
-		err = wait.PollImmediate(1*time.Second, e2e.WaitMedium, func() (bool, error) {
-			if err := client.Create(context.TODO(), workLoad); err != nil {
-				glog.Errorf("error querying api for workLoad object: %v, retrying...", err)
-				return false, err
-			}
-			return true, nil
-		})
+		err = client.Create(context.TODO(), workLoad)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
 		glog.Info("Waiting for cluster to scale out number of replicas")
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -186,9 +187,9 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 		}
 
 		g.By(fmt.Sprintf("Create ClusterAutoscaler and MachineAutoscaler objects. Targeting machineSet %s", targetMachineSet.Name))
-		initialNumberOfReplicas := targetMachineSet.Spec.Replicas
+		initialNumberOfReplicas := pointer.Int32PtrDerefOr(targetMachineSet.Spec.Replicas, 0)
 		clusterAutoscaler := clusterAutoscalerResource()
-		machineAutoscaler := machineAutoscalerResource(&targetMachineSet, 1, *initialNumberOfReplicas+1)
+		machineAutoscaler := machineAutoscalerResource(&targetMachineSet, 1, initialNumberOfReplicas+1)
 		err = wait.PollImmediate(1*time.Second, e2e.WaitMedium, func() (bool, error) {
 			if err := client.Create(context.TODO(), clusterAutoscaler); err != nil {
 				if !strings.Contains(err.Error(), "already exists") {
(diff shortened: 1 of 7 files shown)


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -207,40 +218,24 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
...
-			wait.PollImmediate(1*time.Second, e2e.WaitShort, func() (bool, error) {
-				if err := client.Delete(context.TODO(), machineAutoscaler); err != nil {
-					glog.Errorf("error querying api for machineAutoscaler object: %v, retrying...", err)
-					return false, nil
-				}
-				return true, nil
-			})
+			err = e2e.DeleteObjectsByLabels(context.TODO(), client, map[string]string{autoscalingTestLabel: ""}, &caov1alpha1.MachineAutoscalerList{})
+			if err != nil {
+				glog.Warning(err)
```
frobware: It is OK to continue in the face of an error?

pkg/e2e/framework/common.go
```diff
@@ -9,3 +20,43 @@ const (
 	WaitLong        = 10 * time.Minute
 	RetryMedium     = 5 * time.Second
 )
+
+// DeleteObjectsByLabels list all objects of a given kind by labels and deletes them.
+// Currently supported kinds:
+// -
```
frobware: To be continued... ?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -197,25 +196,16 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
...
 		}
 
-		g.By(fmt.Sprintf("Create ClusterAutoscaler and MachineAutoscaler objects. Targeting machineSet %s", targetMachineSet.Name))
-		initialNumberOfReplicas := *targetMachineSet.Spec.Replicas
+		g.By("Create ClusterAutoscaler object")
 		clusterAutoscaler := clusterAutoscalerResource()
+		err = client.Create(context.TODO(), clusterAutoscaler)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		initialNumberOfReplicas := *targetMachineSet.Spec.Replicas
```
frobware: This is a general comment for the PR: we should choose to deref safely using `pointer.Int32PtrDerefOr()`.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -117,175 +121,151 @@ func labelMachineSetNodes(client runtimeclient.Client, ms *mapiv1beta1.MachineSe
 	})
 }
 
+// Build default CA resource to allow fast scaling up and down
+func clusterAutoscalerResource() *caov1alpha1.ClusterAutoscaler {
```
enxebre: can we move this into utils?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -117,175 +121,151 @@ func labelMachineSetNodes(client runtimeclient.Client, ms *mapiv1beta1.MachineSe
...
-			}
-			return true, nil
-		})
+		g.By("Create ClusterAutoscaler object")
+		clusterAutoscaler := clusterAutoscalerResource()
+		err = client.Create(context.TODO(), clusterAutoscaler)
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		workLoad := newWorkLoad()
+		initialNumberOfReplicas := *targetMachineSet.Spec.Replicas
```
enxebre: can we use `pointer.Int32PtrDerefOr`

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -117,175 +121,151 @@ func labelMachineSetNodes(client runtimeclient.Client, ms *mapiv1beta1.MachineSe
...
+		g.By("Waiting for cluster to scale out number of replicas")
 		err = wait.PollImmediate(5*time.Second, e2e.WaitLong, func() (bool, error) {
-			msKey := types.NamespacedName{
-				Namespace: e2e.TestContext.MachineApiNamespace,
-				Name:      targetMachineSet.Name,
-			}
 			ms := &mapiv1beta1.MachineSet{}
-			if err := client.Get(context.TODO(), msKey, ms); err != nil {
-				glog.Errorf("error querying api for clusterAutoscaler object: %v, retrying...", err)
+			if err := client.Get(context.TODO(), targetMachineSetKey, ms); err != nil {
```
enxebre: can we create `getMachineSet` in utils?

pkg/e2e/infra/infra.go
```diff
@@ -34,10 +35,15 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
 		client, err := e2e.LoadClient()
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		machines, err := getMachines(client)
+		machineList := mapiv1beta1.MachineList{}
```
enxebre: can we update getMachines to use `client.List` and keep this as it was?

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: I prefer to keep it in the same file as it is autoscaling specific. Unless it is needed in a different `e2e/PACKAGE` suite. The less we expose/export the better.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Wondering if that would help. I would rather panic in case `targetMachineSet.Spec.Replicas == nil` rather than set some default value and then debug unexpected behaviour caused by it for 1h.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: I removed the function on purpose since it can be replaced by the one-liner you are commenting onto.

pkg/e2e/infra/infra.go
ingvagabund: I removed `getMachines` function on purpose. As long as it can be replaced by a single invocation, we should not hide it in another function/method.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: It's in the defer part where all the created resources are deleted. So, if this one fails, there are still other resources to be deleted.

pkg/e2e/framework/common.go
ingvagabund: Yes, I will fill it up.

pkg/e2e/autoscaler/autoscaler.go
frobware: Maybe good to add:

```Go
// if this one fails, there are still other resources to be deleted.
```

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: addressed

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -143,55 +190,18 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
...
-				MinReplicas: 1,
-				ScaleTargetRef: caov1alpha1.CrossVersionObjectReference{
-					Name:       targetMachineSet.Name,
-					Kind:       "MachineSet",
-					APIVersion: "machine.openshift.io/v1beta1",
-				},
-			},
-		}
+		clusterAutoscaler := clusterAutoscalerResource()
+		machineAutoscaler := machineAutoscalerResource(&targetMachineSet, 1, *initialNumberOfReplicas+1)
```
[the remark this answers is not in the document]
ingvagabund: addressed

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: done

pkg/e2e/autoscaler/autoscaler.go
frobware: I think there's a lot to be said in having it scoped locally if it is only used locally, so +1 from me.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -348,8 +349,8 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 				glog.Errorf("error querying api for machineSet object: %v, retrying...", err)
 				return false, nil
 			}
-			glog.Infof("Initial number of replicas: %d. Current number of replicas: %d", *initialNumberOfReplicas, *ms.Spec.Replicas)
-			if *ms.Spec.Replicas > *initialNumberOfReplicas {
+			glog.Infof("Initial number of replicas: %d. Current number of replicas: %d", initialNumberOfReplicas, *ms.Spec.Replicas)
+			if *ms.Spec.Replicas > initialNumberOfReplicas {
```
frobware: We should use deref here too.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -362,7 +363,7 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 			scaledNodesLen := int32(len(scaledNodes.Items))
 			glog.Infof("Current number of replicas: %d. Current number of nodes: %d", *ms.Spec.Replicas, scaledNodesLen)
 			// get all linked nodes (so we can wait later on their deletion)
-			return scaledNodesLen <= *ms.Spec.Replicas && scaledNodesLen <= *initialNumberOfReplicas, nil
+			return scaledNodesLen <= *ms.Spec.Replicas && scaledNodesLen <= initialNumberOfReplicas, nil
```
frobware: We should use deref here too.

pkg/e2e/infra/utils.go
```diff
@@ -138,8 +138,8 @@ func getMachineSets(client runtimeclient.Client) ([]mapiv1beta1.MachineSet, erro
 	return machineSetList.Items, nil
 }
 
-// getMachineList returns a MachineList object
-func getMachineList(client runtimeclient.Client) (*mapiv1beta1.MachineList, error) {
+// getMachines returns the list of machines or an error
+func getMachines(client runtimeclient.Client) ([]mapiv1beta1.Machine, error) {
 	machineList := mapiv1beta1.MachineList{}
```
enxebre: why removing `getMachineList` instead of reusing it?

frobware: I read this as the same thing, though instead of returning the `List` it returns a slice - the latter seems better IMO. And I'm sure I've requested similar changes in other PRs this week which do the same thing.

pkg/e2e/infra/utils.go
```diff
@@ -332,12 +314,12 @@ func getScaleClient() (scale.ScalesGetter, error) {
 
 // nodesSnapShotLogs logs the state of all the nodes in the cluster
 func nodesSnapShotLogs(client runtimeclient.Client) error {
-	nodes, err := getNodes(client)
-	if err != nil {
-		return fmt.Errorf("error calling getNodes: %v", err)
+	nodeList := corev1.NodeList{}
```
enxebre: why this change, we are consistently trying to use get*s abstraction rather than list

pkg/e2e/infra/utils.go
enxebre: based on the changes afterwards we probably want to drop this commit https://github.com/openshift/cluster-api-actuator-pkg/pull/49/commits/82456b63439e03c76fbc48b8a612fae3e24bbbb7?

pkg/e2e/infra/infra.go
```diff
@@ -144,10 +145,14 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
-		machineSets, err := getMachineSets(client)
+		machineSetList := &mapiv1beta1.MachineSetList{}
+		err = client.List(
+			context.TODO(),
+			(&runtimeclient.ListOptions{}).InNamespace(e2e.TestContext.MachineApiNamespace),
+			machineSetList)
 		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(len(machineSets)).To(o.BeNumerically(">", 2))
-		machineSet := machineSets[0]
+		o.Expect(len(machineSetList.Items)).To(o.BeNumerically(">", 2))
```
enxebre: why are we removing getMachineSets?

pkg/e2e/infra/utils.go
```diff
@@ -138,18 +138,11 @@ func getMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
-	if err := wait.PollImmediate(1*time.Second, time.Minute, func() (bool, error) {
-		if err := client.Get(context.TODO(), key, &machine); err != nil {
-			glog.Errorf("Error querying api for nodeList object: %v, retrying...", err)
-			return false, nil
-		}
-		return true, nil
-	}); err != nil {
-		glog.Errorf("Error calling getMachineFromNode: %v", err)
-		return nil, err
+	if err := client.Get(context.TODO(), runtimeclient.ObjectKey{Namespace: namespace, Name: machineName}, &machine); err != nil {
```
enxebre: don't we want to keep this short retry for transient network issues?

pkg/e2e/infra/infra.go
ingvagabund: Put back

pkg/e2e/infra/utils.go
ingvagabund: Put back

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Done

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Done

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: Done

pkg/e2e/autoscaler/autoscaler.go
enxebre: I was suggesting pkg/e2e/autoscaler/utils.go so no details are exposed on the ginkgo file, so it's smaller and just reads user stories. Not a blocker for me though

ingvagabund:
This obviously is not right:
```
@@ -162,14 +165,16 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
 
 		g.By(fmt.Sprintf("waiting for cluster to grow %d nodes. Size should be %d", clusterGrowth, intermediateClusterSize))
-		waitForClusterSizeToBeHealthy(intermediateClusterSize)
+		err = waitForClusterSizeToBeHealthy(client, initialClusterSize)
+		o.Expect(err).NotTo(o.HaveOccurred())
 
 		g.By(fmt.Sprintf("scaling in %q machineSet to %d replicas", machineSet.Name, scaleIn))
 		err = scaleMachineSet(machineSet.Name, scaleIn)
 
 		g.By(fmt.Sprintf("waiting for cluster to decrease %d nodes. Final size should be %d nodes", clusterDecrease, finalClusterSize))
-		waitForClusterSizeToBeHealthy(finalClusterSize)
+		err = waitForClusterSizeToBeHealthy(client, initialClusterSize)
+		o.Expect(err).NotTo(o.HaveOccurred())
```

Just changed back to `intermediateClusterSize` and `finalClusterSize`.

PR #50, opened 2019-03-15
Move node draining under infra

ingvagabund:
Node draining is now part of the machine controller. Thus, `pkg/e2e/actuator` is no longer needed.


pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+		if err := client.Get(context.TODO(), key, &rcObj); err != nil {
+			glog.Errorf("Error querying api RC %q object: %v, retrying...", rc.Name, err)
+			return false, nil
+		}
+		if rcObj.Status.ReadyReplicas == 0 {
+			glog.Infof("Waiting for at least one RC ready replica (%v/%v)", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
+			return false, nil
+		}
+		glog.Infof("Waiting for RC ready replicas (%v/%v)", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
+		if rcObj.Status.Replicas != rcObj.Status.ReadyReplicas {
```
frobware: nit: Collapse the remainder to this expression.

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+		key := types.NamespacedName{
+			Namespace: rc.Namespace,
+			Name:      rc.Name,
+		}
+		if err := client.Get(context.TODO(), key, &rcObj); err != nil {
+			glog.Errorf("Error querying api RC %q object: %v, retrying...", rc.Name, err)
+			return false, nil
+		}
+		if rcObj.Status.ReadyReplicas == 0 {
+			glog.Infof("Waiting for at least one RC ready replica (%v/%v)", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
```
frobware: Nit: add some context to `%v/%v`. e.g., `ready: %v, actual %v`.

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+		}
+		if err := client.Get(context.TODO(), key, &machine); err != nil {
+			glog.Errorf("Error querying api machine %q object: %v, retrying...", targetMachine.Name, err)
+			return false, nil
+		}
+		if machine.Status.NodeRef == nil || machine.Status.NodeRef.Kind != "Node" {
+			glog.Error("Machine %q not linked to a node", machine.Name)
+		}
+
+		drainedNodeName = machine.Status.NodeRef.Name
```
frobware: We deref `status.NodeRef` which may be nil. Is the previous `if` expression supposed to return if it is `nil` or not a node?

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+
+		pods := corev1.PodList{}
+		listOpt := &runtimeclient.ListOptions{}
+		listOpt.MatchingLabels(rc.Spec.Selector)
+		if err := client.List(context.TODO(), listOpt, &pods); err != nil {
+			glog.Errorf("Error querying api for Pods object: %v, retrying...", err)
+			return false, nil
+		}
+
+		// expecting nodeGroupSize nodes
```
frobware: This comment talks about nodes yet we are counting pods. Is the comment out of date? Is there a 1:1 between #nodes and #pods?

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+			Name:      rc.Name,
+		}
+		if err := client.Get(context.TODO(), key, &rcObj); err != nil {
+			glog.Errorf("Error querying api RC %q object: %v, retrying...", rc.Name, err)
+			return false, nil
+		}
+
+		// The point of the test is to make sure majority of the pods is rescheduled
+		// to other nodes. Pod disruption budget makes sure at most one pod
+		// owned by the RC is not Ready. So no need to test it. Though, usefull to have it printed.
```
frobware: typo: "useful".

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+		key = types.NamespacedName{
+			Namespace: rc.Namespace,
+			Name:      rc.Name,
+		}
+		if err := client.Get(context.TODO(), key, &rcObj); err != nil {
+			glog.Errorf("Error querying api RC %q object: %v, retrying...", rc.Name, err)
+			return false, nil
+		}
+
+		// The point of the test is to make sure majority of the pods is rescheduled
```
frobware: nit: "are" rescheduled

pkg/e2e/infra/utils.go
```diff
@@ -422,3 +424,138 @@ func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclien
...
+		// to other nodes. Pod disruption budget makes sure at most one pod
+		// owned by the RC is not Ready. So no need to test it. Though, usefull to have it printed.
+		glog.Infof("RC ReadyReplicas/Replicas: %v/%v", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
+
+		// This makes sure at most one replica is not ready
+		if rcObj.Status.Replicas-rcObj.Status.ReadyReplicas > 1 {
+			return false, fmt.Errorf("pod disruption budget not respected, node was not properly drained")
+		}
+
+		// Depends on timing though a machine can be deleted even before there is only
```
frobware: If it is dependent on timing, is this teat reliable/repeatable and/or still deterministic?

pkg/e2e/infra/utils.go
ingvagabund: right, added `return false, nil`

pkg/e2e/infra/utils.go
ingvagabund: Left over, comment removed

pkg/e2e/infra/utils.go
ingvagabund: It's still reliable. The thing is a machine is delete right after the drain operation succeeds. And the drain operation succeeds when there is no pod running on a node. So in case the operation is quick (e.g. it's easy to delete pod that is running just `sleep` command), the retry period of `e2e.RetryMedium` (5 seconds) can be too long to capture a case where there is only one running pod left. One could argue even those two pods could be deleted in the interval. Thus the reason why there is 10 pods to be evicted so the chance is minimized.

pkg/e2e/infra/utils.go
```diff
@@ -437,14 +437,11 @@ func waitUntilAllRCPodsAreReady(client runtimeclient.Client, rc *corev1.Replicat
 			return false, nil
 		}
 		if rcObj.Status.ReadyReplicas == 0 {
-			glog.Infof("Waiting for at least one RC ready replica (%v/%v)", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
+			glog.Infof("Waiting for at least one RC ready replica, ReadyReplicas: %v, Replicas: %v", rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
```
frobware: Do the same change in a few lines below for consistency now.

ingvagabund: Done

pkg/e2e/infra/utils.go
```diff
@@ -336,3 +339,220 @@ func waitUntilAllNodesAreSchedulable(client runtimeclient.Client) error {
...
+	for key := range nodeDrainLabels {
+		if _, exists := machine.Spec.ObjectMeta.Labels[key]; exists {
+			continue
+		}
+		machine.Spec.ObjectMeta.Labels[key] = nodeDrainLabels[key]
+	}
+	return machine
+}
+
+func waitUntilNodesAreReady(client runtimeclient.Client, listOpt *runtimeclient.ListOptions, nodeCount int) error {
```
enxebre: there's already a lib function to check nodes are ready. We should use that and rely on ".eventually" on the ginkgo script. So we keep this library for manipulating reosurces atomically as free as possible from wait operations cc @ingvagabund @frobware

pkg/e2e/infra/utils.go
```diff
@@ -336,3 +339,220 @@ func waitUntilAllNodesAreSchedulable(client runtimeclient.Client) error {
...
+			glog.Errorf("Expecting %v nodes with %#v labels in Ready state, got %v", nodeCount, nodeDrainLabels, readyNodes)
+			return false, nil
+		}
+
+		glog.Infof("Expected number (%v) of nodes with %v label in Ready state found", nodeCount, nodeDrainLabels)
+		return true, nil
+	})
+}
+
+func waitUntilNodesAreDeleted(client runtimeclient.Client, listOpt *runtimeclient.ListOptions) error {
```
enxebre: there's already lib functions to ask for nodes and similar operations. We should use that and rely on ".eventually" on the ginkgo script. So we keep this library for manipulating reosurces atomically as free as possible from wait operations cc @ingvagabund @frobware

pkg/e2e/infra/utils.go
```diff
@@ -336,3 +339,220 @@ func waitUntilAllNodesAreSchedulable(client runtimeclient.Client) error {
...
+		}
+
+		glog.Info("Expected result: all pods from the RC up to last one or two got scheduled to a different node while respecting PDB")
+		return true, nil
+	})
+
+	return drainedNodeName, err
+}
+
+func waitUntilNodeDoesNotExists(client runtimeclient.Client, nodeName string) error {
```
enxebre: there's already a lib function to check nodes are ready. We should use that and rely on ".eventually" on the ginkgo script. So we keep this library for manipulating resources atomically as free as possible from wait operations cc @ingvagabund @frobware

PR #47, opened 2019-03-12
Machine healthcheck tests

cynepco3hahue:
Tests:
- [Feature:MachineHealthCheck] MachineHealthCheck controller
  with node-unhealth-conditions configmap should delete unhealthy machine
- [Feature:MachineHealthCheck] MachineHealthCheck controller
  should delete the unhealthy machine
    
Copy tests from the https://github.com/openshift/machine-api-operator/pull/204

PR #51, opened 2019-03-17
Refactor machine-health-check tests

cynepco3hahue:
- move setup logic to separate methods
- add TODO on the kubemark skip condition

### DIFF
--- a/pkg/e2e/framework/common.go
+++ b/pkg/e2e/framework/common.go
@@ -15,11 +15,11 @@ import (
 )
 
 const (
-	WorkerRoleLabel = "node-role.kubernetes.io/worker"
-	WaitShort       = 1 * time.Minute
-	WaitMedium      = 3 * time.Minute
-	WaitLong        = 10 * time.Minute
-	RetryMedium     = 5 * time.Second
+	WorkerNodeRoleLabel = "node-role.kubernetes.io/worker"
+	WaitShort           = 1 * time.Minute
+	WaitMedium          = 3 * time.Minute
+	WaitLong            = 10 * time.Minute
+	RetryMedium         = 5 * time.Second
 
 	// DefaultMachineSetReplicas is the default number of replicas of a machineset
 	// if MachineSet.Spec.Replicas field is set to nil
--- a/pkg/e2e/framework/nodes.go
+++ b/pkg/e2e/framework/nodes.go
@@ -1,0 +1,33 @@
+package framework
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetWorkerNodes returns all nodes with the nodeWorkerRoleLabel label
+func GetWorkerNodes(client runtimeclient.Client) ([]corev1.Node, error) {
+	listOptions := runtimeclient.ListOptions{
+		Namespace: TestContext.MachineApiNamespace,
+	}
+	listOptions.MatchingLabels(map[string]string{WorkerNodeRoleLabel: ""})
+	workerNodes := &corev1.NodeList{}
+	err := client.List(context.TODO(), &listOptions, workerNodes)
+	if err != nil {
+		return nil, err
+	}
+	return workerNodes.Items, nil
+}
+
+// FilterReadyNodes fileter the list of nodes and returns the list with ready nodes
+func FilterReadyNodes(nodes []corev1.Node) []corev1.Node {
+	var readyNodes []corev1.Node
+	for _, n := range nodes {
+		if IsNodeReady(&n) {
+			readyNodes = append(readyNodes, n)
+		}
+	}
+	return readyNodes
+}
--- a/pkg/e2e/infra/infra.go
+++ b/pkg/e2e/infra/infra.go
@@ -26,8 +26,8 @@ import (
 )
 
 var nodeDrainLabels = map[string]string{
-	e2e.WorkerRoleLabel:  "",
-	"node-draining-test": string(uuid.NewUUID()),
+	e2e.WorkerNodeRoleLabel: "",
+	"node-draining-test":    string(uuid.NewUUID()),
 }
 
 func replicationControllerWorkload(namespace string) *corev1.ReplicationController {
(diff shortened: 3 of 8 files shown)


ingvagabund:
We do not share code between sub test suites. Either, if the code is specific for `pkg/e2e/machinehealthcheck`, please put the code under utils.go of the package and make methods/functions private. In case of defining methods/functions that can be reused among the sub test-suites, please put them under `pkg/e2e/framework`.

cynepco3hahue:
@ingvagabund I thought it what @enxebre meant under the comment https://github.com/openshift/cluster-api-actuator-pkg/pull/47#discussion_r265631793

ingvagabund:
> I thought it what @enxebre meant under the comment #47 (comment)

I see. `pkg/e2e/infra/utils.go` is primarily meant to be used only by `pkg/e2e/infra` package. Given all the `generic` functions are there, it's natural to put your functions there as well. However, I would like to avoid exporting anything but the infra sub test-suite itself from `pkg/e2e/infra`. The ultimate goal is to move anything common under either `pkg/e2e/framework` or under `pkg/e2e/new-and-to-be-created` package.

We already have functionality of `GetMachineByNode` under `getMachineFromNode` [1]. Plus `GetReadyWorkerNodes` could be implemented as combination of `getWorkerNode` [2] and `nodesAreReady` [3]. Though, `getWorkerNode` and the place where it is invoked will need to be modified a bit. Thus, I would instead suggest to move `getWorkerNode` [2] and `nodesAreReady` [3] under `pkg/e2e/framework/common.go` file and make the functions exported. Then importing them in `pkg/e2e/machinehealthcheck` and `pkg/e2e/infra`.

[1] https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/infra/utils.go#L125
[2] https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/infra/utils.go#L194
[3] https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/infra/utils.go#L216

cynepco3hahue:
@ingvagabund I re-worked the PR, can you please review again and run tests on it?

pkg/e2e/framework/nodes.go
```diff
@@ -0,0 +1,39 @@
...
+	workerNodes := &corev1.NodeList{}
+	err := client.List(context.TODO(), &listOptions, workerNodes)
+	if err != nil {
+		return nil, err
+	}
+	return workerNodes.Items, nil
+}
+
+// GetWorkerReadyNodes returns all ready nodes with the nodeWorkerRoleLabel label
+func GetWorkerReadyNodes(client runtimeclient.Client) ([]corev1.Node, error) {
```
ingvagabund: Can you rename the function to `FilterReadyNodes` and pass a list of nodes instead? We already have a use case where we verify all nodes (including master nodes) are ready.

pkg/e2e/framework/nodes.go
```diff
@@ -0,0 +1,39 @@
...
+	corev1 "k8s.io/api/core/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetWorkerNodes returns all nodes with the nodeWorkerRoleLabel label
+func GetWorkerNodes(client runtimeclient.Client) ([]corev1.Node, error) {
+	listOptions := runtimeclient.ListOptions{
+		Namespace: TestContext.MachineApiNamespace,
+	}
+	listOptions.SetLabelSelector(fmt.Sprintf("%s=", NodeWorkerRoleLabel))
```
ingvagabund: Please, use `listOptions.MatchingLabels(map[string]string{nodeWorkerRoleLabel: ""})` instead.

pkg/e2e/infra/infra.go
```diff
@@ -195,9 +195,22 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
+			workerNodes, err := e2e.GetWorkerNodes(client)
+			if err != nil {
+				return false
+			}
+			if len(workerNodes) == 0 {
+				return false
+			}
+			workerNode = &workerNodes[0]
+			return true
+		}, time.Minute, time.Second).ShouldNot(o.BeTrue())
```
ingvagabund: Should not be true? I would expect the opposite.

pkg/e2e/infra/infra.go
```diff
@@ -195,9 +195,22 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
 
-		workerNode, err := getWorkerNode(client)
+		g.By("getting worker node")
+		var workerNode *corev1.Node
+		o.Eventually(func() bool {
+			workerNodes, err := e2e.GetWorkerNodes(client)
+			if err != nil {
+				return false
+			}
+			if len(workerNodes) == 0 {
```
ingvagabund: Is this fixing a known issue? If not, I would still prefer to use the original form `workerNode, err := e2e.GetWorkerNodes(client)`. If we want to assert the fact that there is at least one worker machine, it needs to go outside of `o.Eventually` so the test can fail in case there is no worker node.

pkg/e2e/framework/common.go
```diff
@@ -15,11 +15,11 @@ import (
 )
 
 const (
-	WorkerRoleLabel = "node-role.kubernetes.io/worker"
-	WaitShort       = 1 * time.Minute
-	WaitMedium      = 3 * time.Minute
-	WaitLong        = 10 * time.Minute
-	RetryMedium     = 5 * time.Second
+	NodeWorkerRoleLabel = "node-role.kubernetes.io/worker"
```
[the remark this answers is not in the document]
cynepco3hahue: done

pkg/e2e/framework/nodes.go
cynepco3hahue: Done
Can you point me where can I replace the code with the `FilterReadyNodes`?

pkg/e2e/framework/nodes.go
cynepco3hahue: done

pkg/e2e/infra/infra.go
cynepco3hahue: nice catch, done

pkg/e2e/infra/infra.go
cynepco3hahue: I just saw that `getWorkerNode` used `wait.PollImmediate`, so I entered the `e2e.GetWorkerNodes(client)` under the `Eventually`, but if `wait.PollImmediate` was some redundant code I will happily remove the `Eventually`.

PR #53, opened 2019-03-20
Update readme implementation details

enxebre:
Update readme implementation details

### DIFF
--- a/pkg/e2e/README.md
+++ b/pkg/e2e/README.md
@@ -34,12 +34,18 @@ https://github.com/openshift/autoscaler
 
 ## Implementation
 
-We use [ginkgo](https://onsi.github.io/ginkgo/)
+- We use [ginkgo](https://onsi.github.io/ginkgo/)
 
-These tests assume the existence of the system to be validated, i.e an OpenShift cluster and a Kubeconfig file
+- These tests assume the existence of the system to be validated, i.e an OpenShift cluster and a Kubeconfig file
 
-Product expectations should be as implementation agnostic as possible e.g "When the workload increases I expect my cluster to grow".
+- Product expectations should be as implementation agnostic as possible e.g "When the workload increases I expect my cluster to grow".
 The API Group used for this to happen and the implementation details might vary. The user expectation remains the same
+
+- We are developing a reusable library for manipulating machine API resources that should be leveraged for writing e2e expectations
+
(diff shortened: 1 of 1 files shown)


enxebre:
cc @frobware @ingvagabund

pkg/e2e/README.md
```diff
@@ -34,13 +34,19 @@ https://github.com/openshift/autoscaler
...
+- These tests assume the existence of the system to be validated, i.e an OpenShift cluster and a Kubeconfig file
 
-Product expectations should be as implementation agnostic as possible e.g "When the workload increases I expect my cluster to grow".
+- Product expectations should be as implementation agnostic as possible e.g "When the workload increases I expect my cluster to grow".
 The API Group used for this to happen and the implementation details might vary. The user expectation remains the same
 
+- We are developing a reusable library for manipulating machine API resources that should be leveraged for writing e2e expectations
+
+- Product expectations should be simple, easy to read and avoid details. To this end:
+  - They must leverage the common library for manipulating machine API resources so this is done in a consistent manner across all e2e, e.g https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/infra/utils.go#L113
```
ingvagabund: Link to `utils.go` with a specific commit please.

pkg/e2e/README.md
```diff
@@ -34,13 +34,19 @@ https://github.com/openshift/autoscaler
...
+  - Polling logic should be expressed at the Ginkgo level when possible. Higher polling abstractions can be defined here leveraging the common library. E.g https://github.com/openshift/cluster-api-actuator-pkg/blob/a086b6d5db86e7e48a6ebce24343e461be233440/pkg/e2e/infra/infra.go#L184
```
ingvagabund: - I would explicitly rename every ginkgo construct that we allow to be used (Describe, It, By, assertions). Plus mentioning other can be used but they need to be explicitly justified.
- If `o.Eventually` needs to be used, there's a high chance the code inside the assertion can be moved into a function where `wait.PollImmediate`can be used instead.
- All the ginkgo constructs can be used only on the high level. No function/method is allowed to use the constructs unless explicitly justified.

enxebre: this is not trying to be restrictive with `Describe, It, By,` assertions just yet.

PR #58, opened 2019-04-01
Wait longer for clusterapi-manager-controllers deployment do re-appear

ingvagabund:
Sometimes a clusterapi-manager-controllres-XXXX pod stucks in Terminated
state before a deployment is allowed to be created again. Thus, it takes
longer than 1 minute for a deleted deployment to be re-created by
the machine api operator.


enxebre:
why they do `stucks in Terminated`?
/approve

ingvagabund:
> why they do stucks in Terminated?

Sometime it takes longer for a pod to get Terminated. E.g. takes longer to stop a container.

PR #59, opened 2019-04-02
e2e/operators/utils: Bump isDeploymentAvailable wait to long

bison:
This often takes a bit longer than the medium wait time.


enxebre:
do we know why this take more than 3 minutes?

bison:
> do we know why this take more than 3 minutes?

No idea. One issue is that this may be used in different situations that call for different timeouts, e.g. some containers have longer graceful shutdown times etc. Happy to close this and we can work on getting more info about specific cases. I just got tired of typing `/retest` yesterday.

PR #60, opened 2019-04-12
e2e-aws-operator: Fix machinehealthcheck test

vikaschoudhary16:
In the current test,  [BeforeEach()](https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/machinehealthcheck/machinehealthcheck.go#L90-#L114) gets executed for [this `It`](https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/machinehealthcheck/machinehealthcheck.go#L90-#L114). This Before reach deletes configmap as well and when, actually `[stopKubeletAndValidateMachineDeletion](https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/machinehealthcheck/machinehealthcheck.go#L117)` gets invoked again, there is no configmap and it fails to validate node deletion. And thus test failure!!

I verified this behavior running these tests locally and it occurs every time. 

This PR is simplifying the nesting of BeforeEach and AfterEach. If the intention of the existing changes and this nesting is to execute stopKubeletAndValidateMachineDeletion twice for two machines, we can achieve that by simply copy pasting `IT` block in this PR.

### DIFF
--- a/pkg/e2e/machinehealthcheck/machinehealthcheck.go
+++ b/pkg/e2e/machinehealthcheck/machinehealthcheck.go
@@ -113,8 +113,10 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck controller", f
 		})
 	})
 
-	It("should delete unhealthy machine", func() {
-		stopKubeletAndValidateMachineDeletion(workerNode, workerMachine, 6*time.Minute)
+	Context("without node-unhealthy-conditions configmap", func() {
+		It("should delete unhealthy machine", func() {
+			stopKubeletAndValidateMachineDeletion(workerNode, workerMachine, 7*time.Minute)
+		})
 	})
 
 	AfterEach(func() {


enxebre:
ping @cynepco3hahue

cynepco3hahue:
@vikaschoudhary16 Can you provide the link where does it fail?

cynepco3hahue:
From logs, I can see that machine was deleted by the machine health check controller, but the machine-api failed to remove finalizer
```
Failed to remove finalizer from machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": Operation cannot be fulfilled on machines.machine.openshift.io "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": the object has been modified; please apply your changes to the latest version and try again
...
I0414 10:49:26.531912       1 controller.go:128] Reconciling Machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7"
I0414 10:49:26.531931       1 controller.go:300] Machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7" in namespace "cluster.k8s.io/cluster-name" doesn't specify "openshift-machine-api" label, assuming nil cluster
I0414 10:49:26.531947       1 controller.go:185] Reconciling machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7" triggers delete
```

cynepco3hahue:
I created the PR - https://github.com/openshift/cluster-api-actuator-pkg/pull/62, it will remove finalizer when it needs to be removed, but I think it looks like the bug under the `machine-api` code.
@ingvagabund @enxebre Guys, what do you think regarding https://github.com/openshift/cluster-api-actuator-pkg/pull/60#issuecomment-482961091? Does it bug under the machine-api controller?

ingvagabund:
> E0414 10:49:25.525278       1 controller.go:219] Failed to remove finalizer from machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": Operation cannot be fulfilled on machines.machine.openshift.io "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": the object has been modified; please apply your changes to the latest version and try again

@cynepco3hahue this basically says the finalizer could not be removed since the machine object on the apiserver is never than the one the machine controller actually works with. That error is expected and the machine is reconciled again. It will get removed in the next reconciliation cycle.

What's important are those two lines:

```
E0414 10:49:25.525278       1 controller.go:219] Failed to remove finalizer from machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": Operation cannot be fulfilled on machines.machine.openshift.io "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7": the object has been modified; please apply your changes to the latest version and try again
...
I0414 10:49:26.531947       1 controller.go:185] Reconciling machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7" triggers delete
```

The first line reporting the unsuccessful Update operation. The other one saying the machine got reconciled again after some time and is to be deleted. Can you share more logs? I would expect the machine controller to eventually log something like: `Machine "ci-op-5r3bjbhs-79b09-fvxxn-worker-us-east-1a-ls4t7" deletion successful`.

vikaschoudhary16:
@cynepco3hahue For me it fails consistently. Checking the machine controller logs, i can see that finalizer gets removed and machine gets deleted eventually though the test fails.

When i increased test case timeout to 7 minutes, it worked. Since this failure has been seen at CI as well, i propose to increase the time out and also wrapping the test for default case in proper context message.

ingvagabund:
@vikaschoudhary16 as you say: https://github.com/openshift/machine-api-operator/pull/285#issuecomment-483220463

@cynepco3hahue it takes more than 6 minutes. In my case it took 8 minutes (without 3 seconds) to wait until a machine gets deleted.

pkg/e2e/machinehealthcheck/machinehealthcheck.go
```diff
@@ -113,8 +113,10 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck controller", f
 		})
 	})
 
-	It("should delete unhealthy machine", func() {
-		stopKubeletAndValidateMachineDeletion(workerNode, workerMachine, 6*time.Minute)
+	Context("without node-unhealthy-conditions configmap", func() {
+		It("should delete unhealthy machine", func() {
+			stopKubeletAndValidateMachineDeletion(workerNode, workerMachine, 7*time.Minute)
```
ingvagabund: I don't see how putting the `It` clause under `Context` help with to fix this test. It just might be the timeout that got increased that helped. We might bump the timeout to even 10 minutes. Though, that will only hide the real issue. This might also be a bug in the HC itself.

vikaschoudhary16: putting IT under context is just adding correct description about test case. 

5 minutes is timeout in the configmap  so machine deletion get triggered after 5 minutes. With keeping test case timeout as 7 min, effectively it is waiting for 2 mins. If we still see I failures, we can debug where exactly its taking delayed. For now, i would prefer to try out 7 min timeout.

enxebre:
this test is disabled now until we get back CI stability, this change should be introduced as we enable it back so it gives a green

vikaschoudhary16:
This got obsoleted and superceded by other PRs so closing it.

PR #63, opened 2019-04-15
Accommodate for new deployment name: machine-api-controllers

enxebre:
Accomodate for better controllers deployment naming https://github.com/openshift/machine-api-operator/pull/278 so it doesn't break the suite in any repo while the payload does not include #278

### DIFF
--- a/pkg/e2e/operators/machine-api-operator.go
+++ b/pkg/e2e/operators/machine-api-operator.go
@@ -8,6 +8,10 @@ import (
 	e2e "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/framework"
 )
 
+var (
+	deploymentDeprecatedName = "clusterapi-manager-controllers"
+)
+
 var _ = g.Describe("[Feature:Operators] Machine API operator deployment should", func() {
 	defer g.GinkgoRecover()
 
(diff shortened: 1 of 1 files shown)


pkg/e2e/operators/machine-api-operator.go
```diff
@@ -23,9 +27,13 @@ var _ = g.Describe("[Feature:Operators] Machine API operator deployment should",
 		client, err := e2e.LoadClient()
 		o.Expect(err).NotTo(o.HaveOccurred())
 
-		deploymentName := "clusterapi-manager-controllers"
+		deploymentName := "machine-api-controllers"
 		initialDeployment, err := getDeployment(client, deploymentName)
-		o.Expect(err).NotTo(o.HaveOccurred())
+		if err != nil {
```
ingvagabund: `if apierrors.IsNotFound(err) {...} else {o.Expect(err).NotTo(o.HaveOccurred())}`

ingvagabund: `apierrors "k8s.io/apimachinery/pkg/api/errors"`

enxebre: realised the error is actually a polling time out

PR #64, opened 2019-04-16
Separate operators from functional testing

michaelgugino:
Operator behavior should be tested independently
from the functional behavior of the components
that the operators deploy.

The primary reason for this is to preserve logs
for troubleshooting test failures and flakes in
the functional tests.

Implements: https://jira.coreos.com/browse/CLOUD-460


Makefile
```diff
@@ -46,7 +46,13 @@ build-e2e:
 .PHONY: test-e2e
 test-e2e: ## Run openshift specific e2e test
 	go test -timeout 90m \
-		-v github.com/openshift/cluster-api-actuator-pkg/pkg/e2e \
+		-v github.com/openshift/cluster-api-actuator-pkg/pkg/e2e-operators \
+		-kubeconfig $${KUBECONFIG:-~/.kube/config} \
+		-machine-api-namespace $${NAMESPACE:-openshift-machine-api} \
+		-ginkgo.v \
+		-args -v 5 -logtostderr true
```
vikaschoudhary16: why logtostderr?

Makefile
```diff
@@ -46,7 +46,13 @@ build-e2e:
 .PHONY: test-e2e
 test-e2e: ## Run openshift specific e2e test
```
ingvagabund: Would
```
define test =
	go test -timeout 90m \
		-v github.com/openshift/cluster-api-actuator-pkg/pkg/e2e \
		-kubeconfig $${KUBECONFIG:-~/.kube/config} \
		-machine-api-namespace $${NAMESPACE:-openshift-machine-api} \
		-args -v 5 -logtostderr \
		$1 $2 $3 $4 $5 $6 $7 $8 $9
endef

.PHONY: test-e2e
	$(call test,-ginkgo.v,-ginkgo.noColor=true,-ginkgo.focus "Feature:Operators")
	$(call test,-ginkgo.v,-ginkgo.noColor=true,-ginkgo.skip "Feature:Operators")
```
serve the same purpose?

Makefile
ingvagabund: without the flag the logs get printed into a file instead of stderr

frobware: We don't need the trailing `true`.

Makefile
frobware: What's the second case supposed to do if the first focuses on the same thing? Is this a typo?

vikaschoudhary16: https://jira.coreos.com/browse/CLOUD-458

Makefile
vikaschoudhary16: Also why not same flag at L#58?

michaelgugino: This is a copied from what already exists.  In any case, I believe it's because that's how you capture the logs from pods.

Makefile
ingvagabund: The first has `gingko.focus` which just runs anything that complies to `Feature:Operators` regex. The second one has `gingko.skip` which runs anything but specs that complies to `Feature:Operators` regex.

ingvagabund: @michaelgugino can you also add a comment into the Makefile explaining why this is needed? E.g. `//The primary reason for this is to preserve logs for troubleshooting test failures and flakes in the tests. Feature:Operator tests remove deployments. Thus loosing all the logs previously acquired.`

michaelgugino: @ingvagabund done.

enxebre:
couldn't we ensure the operator test run first programatically? if so why is this approach preferred?

ingvagabund:
http://onsi.github.io/ginkgo/#spec-permutation

michaelgugino:
Closing in favor of https://github.com/openshift/cluster-api-actuator-pkg/pull/69

PR #67, opened 2019-04-17
Deprecate status=Failing in favor of status=Degraded

michaelgugino:
This commit updates tests and openshift/api to allow
for both failing and degraded status, after all comonents
are updated, we should update to latest openshift/api
and remove traces of status=Failing


Gopkg.toml
```diff
@@ -41,6 +41,10 @@ required = [
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1",
     ]
 
+[[constraint]]
+  name = "github.com/openshift/api"
+  revision = "8e476cb7322e59919cbb6482fd076ec5a214df25"
```
michaelgugino: This is the specific commit that has both Failing and Degraded.  Latest master already has failing removed, so this will be a multi-step process.

PR #66, opened 2019-04-17
Use v1beta1 and v1 versions of cluster-autoscaler-operator resources

bison:
Switches cluster-autoscaler-operator resources from v1alpha1 to v1beta1 and v1 versions.

PR #68, opened 2019-04-18
Temporarily disable k8s e2e

frobware:


frobware:
Close as we have a solution for #66 now.

PR #77, opened 2019-05-01
e2e/autoscaler: scale out in parallel

frobware:
Scale out all machinesets at the same time. The previous
behaviour persisted with scaling one machine set at a time but we
want to validate across all available machine sets, in parallel.

This change also switches to using events to control test
termination. The intent is to implement the test using
cluster-autoscaler primitives and not use implementation
details (e.g., the replica count) on a scalable resource. We
have other e2e tests that validate the behaviour of
increasing/decreasing the replica count for low-level scaling.

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -28,9 +28,6 @@ const (
 )
 
 func newWorkLoad() *batchv1.Job {
-	backoffLimit := int32(4)
-	completions := int32(50)
-	parallelism := int32(50)
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "workload",
@@ -72,9 +69,9 @@ func newWorkLoad() *batchv1.Job {
 					},
 				},
 			},
-			BackoffLimit: &backoffLimit,
-			Completions:  &completions,
-			Parallelism:  &parallelism,
+			BackoffLimit: pointer.Int32Ptr(4),
+			Completions:  pointer.Int32Ptr(50),
+			Parallelism:  pointer.Int32Ptr(50),
 		},
 	}
 }
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -3,6 +3,8 @@ package autoscaler
 import (
 	"context"
 	"fmt"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -16,15 +18,21 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	autoscalingTestLabel = "test.autoscaling.label"
+	autoscalingTestLabel                  = "test.autoscaling.label"
+	clusterAutoscalerComponent            = "cluster-autoscaler"
+	clusterAutoscalerObjectKind           = "ConfigMap"
+	clusterAutoscalerScaledUpGroup        = "ScaledUpGroup"
+	clusterAutoscalerScaleDownEmpty       = "ScaleDownEmpty"
+	clusterAutoscalerMaxNodesTotalReached = "MaxNodesTotalReached"
+	pollingInterval                       = 3 * time.Second
 )
 
 func newWorkLoad() *batchv1.Job {
@@ -121,7 +129,7 @@ func labelMachineSetNodes(client runtimeclient.Client, ms *mapiv1beta1.MachineSe
 }
 
 // Build default CA resource to allow fast scaling up and down
-func clusterAutoscalerResource() *caov1.ClusterAutoscaler {
+func clusterAutoscalerResource(maxNodesTotal int) *caov1.ClusterAutoscaler {
 	tenSecondString := "10s"
 	return &caov1.ClusterAutoscaler{
 		ObjectMeta: metav1.ObjectMeta{
@@ -142,6 +150,9 @@ func clusterAutoscalerResource() *caov1.ClusterAutoscaler {
 				DelayAfterDelete:  &tenSecondString,
 				DelayAfterFailure: &tenSecondString,
 				UnneededTime:      &tenSecondString,
+			},
+			ResourceLimits: &caov1.ResourceLimits{
+				MaxNodesTotal: pointer.Int32Ptr(int32(maxNodesTotal)),
 			},
 		},
 	}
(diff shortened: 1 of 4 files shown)


frobware:
/hold

I may add the additional test that validates `maxNodesTotal` as it would save having to revendor this so many times.

frobware:
This needs https://github.com/openshift/kubernetes-autoscaler/pull/90

frobware:
This will fail until openshift/kubernetes-autoscaler#90 merges.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,166 +139,246 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
+			err := e2e.DeleteObjectsByLabels(context.TODO(), client, map[string]string{autoscalingTestLabel: ""}, &batchv1.JobList{})
+			o.Expect(err).NotTo(o.HaveOccurred())
+
+			if len(cleanupObjects) > 1 && cleanupObjects[len(cleanupObjects)-1] == workload {
+				// workload already deleted above
+				cleanupObjects = cleanupObjects[:len(cleanupObjects)-1]
+			}
+		})
+		g.By(fmt.Sprintf("Waiting for cluster to scale down %d more nodes", atomic.LoadUint64(&scaleDownCount)))
+		return atomic.LoadUint64(&scaleDownCount) == 0, nil
```
vikaschoudhary16: Why scaleDownCount should be 0? clusterExpansionSize has master machines + worker nodes + (extra scaled out - 1). Why it is being asserted that all initial worker nodes and master nodes will also be scaled down?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,166 +139,246 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
 
-			err = e2e.DeleteObjectsByLabels(context.TODO(), client, map[string]string{autoscalingTestLabel: ""}, &caov1.ClusterAutoscalerList{})
-			if err != nil {
-				// if this one fails, there is no point of returning an error as this is the last resource deletion action
-				glog.Warning(err)
-			} else {
-				glog.Info("Deleted clusterAutoscaler object")
-			}
-		}()
+	var maxNodesTotalStartTimeLock sync.Mutex
```
vikaschoudhary16: Why lock is required?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,166 +139,246 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
-			return scaledNodesLen <= msReplicas && scaledNodesLen <= initialNumberOfReplicas, nil
-		})
+		}
+	}()
+
+	g.By("Getting machinesets")
+	machineSets, err := e2e.GetMachineSets(context.TODO(), client)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	// We need at least two machinesets to so that we can scale up
+	// to maxNodesTotal-1.
```
vikaschoudhary16: we need two machinesets because min & max values in asr are such that (max - min == 1). If Max was let say 3, then even one machineset would have been sufficient to carry out the test. Right?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,166 +139,246 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
-			glog.Infof("Current number of replicas: %d. Current number of nodes: %d", msReplicas, scaledNodesLen)
-			return scaledNodesLen <= msReplicas && scaledNodesLen <= initialNumberOfReplicas, nil
-		})
+		}
+	}()
+
+	g.By("Getting machinesets")
+	machineSets, err := e2e.GetMachineSets(context.TODO(), client)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	// We need at least two machinesets to so that we can scale up
```
vikaschoudhary16: s/to so that/so that

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,166 +139,246 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
+	// to assert that we don't grow beyond a specified maximium;
+	// without the maximum the cluster could grow to that.
+	maxNodesTotal := len(nodes) + clusterExpansionSize - 1
+	g.By(fmt.Sprintf("Create ClusterAutoscaler object with maxNodesTotal:%v", maxNodesTotal))
+	clusterAutoscaler := clusterAutoscalerResource(maxNodesTotal)
+	err = client.Create(context.TODO(), clusterAutoscaler)
+	o.Expect(err).NotTo(o.HaveOccurred())
+	cleanupObjects = append(cleanupObjects, runtime.Object(clusterAutoscaler))
+
+	var scaleUpStartTimeLock sync.Mutex
```
vikaschoudhary16: s/scaleUpStartTimeLock/scaleUpCountLock

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -173,165 +182,144 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
+		workload := newWorkLoad()
+		o.Expect(client.Create(context.TODO(), workload)).Should(o.Succeed())
+		cleanupObjects = append(cleanupObjects, runtime.Object(workload))
+		o.Eventually(func() uint32 {
+			v := scaleUpCounter.get()
+			glog.Infof("Waiting for %s to generate %v more %q events", clusterAutoscalerComponent, v, clusterAutoscalerScaledUpGroup)
+			return v
+		}, e2e.WaitLong, 3*time.Second).Should(o.BeZero())
+
+		maxNodesTotalReachedCounter := newMaxNodesTotalReachedCounter(eventWatcher, 1)
```
ingvagabund: What if you observe `Max total nodes in cluster reached` event before the last `Scale-up: setting group`? Then, this will timeout.

frobware: Then that would imply a bug in the cluster-autoscaler.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -173,165 +182,144 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
-		err = client.Create(context.TODO(), newWorkLoad())
+		g.By("Getting nodes")
+		nodes, err := e2e.GetNodes(client)
 		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(len(nodes)).To(o.BeNumerically(">=", 1))
+
+		g.By(fmt.Sprintf("Creating %v machineautoscalers", len(machineSets)))
+		var clusterExpansionSize int
+		for i := range machineSets {
+			o.Expect(pointer.Int32PtrDerefOr(machineSets[i].Spec.Replicas, 0)).To(o.BeNumerically("==", 1))
```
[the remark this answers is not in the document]
frobware: I am not expecting that and want to stop if it happens, but happy to drop this assertion.

Longer term I think we should create our own machine sets with known replica counts, et al, as part of the test.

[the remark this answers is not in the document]
frobware: Dropped in https://github.com/openshift/cluster-api-actuator-pkg/pull/77/commits/9d3759f1a44ac317b028eccda5185b6a0922589c

pkg/e2e/autoscaler/autoscaler.go
frobware: The CA will keep sending `MaxNodesTotalReached` events - it is not just a one-shot event. Everytime `--scan-interval` it will try to scale up but this condition will be met and so send another event.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,165 +136,149 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
+		o.Eventually(func() uint32 {
+			v := scaleUpCounter.get()
+			glog.Infof("Waiting for %s to generate %v more %q events", clusterAutoscalerComponent, v, clusterAutoscalerScaledUpGroup)
+			return v
+		}, e2e.WaitLong, 3*time.Second).Should(o.BeZero())
+
+		// The cluster-autoscaler will keep on generating
+		// MaxNodesTotalReached events because the workload
+		// has many more PODs to run. We only need to assert
+		// this happens at most once.
```
vikaschoudhary16: s/at most/ at least/s

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -176,165 +136,149 @@ func machineAutoscalerResource(targetMachineSet *mapiv1beta1.MachineSet, minRepl
...
+		g.By("Getting nodes")
+		nodes, err := e2e.GetNodes(client)
 		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(len(nodes)).To(o.BeNumerically(">=", 1))
+
+		g.By(fmt.Sprintf("Creating %v machineautoscalers", len(machineSets)))
+		var clusterExpansionSize int
+		for i := range machineSets {
+			min := pointer.Int32PtrDerefOr(machineSets[i].Spec.Replicas, 1)
+			max := min + 1
```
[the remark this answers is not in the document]
frobware: L222 needs to be dynamic. As we saw recently there is no guarantee that min is 1.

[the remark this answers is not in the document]
frobware: Done in https://github.com/openshift/cluster-api-actuator-pkg/pull/77/commits/4910e6120f20c5d683fcee02cb2bfb6ccda1b657

PR #78, opened 2019-05-02
Relax machineSets exepectations

enxebre:
Relax the expected number of machineSets and replicas based on https://github.com/openshift/installer/issues/1698#issuecomment-488658119 and https://github.com/openshift/release/pull/3615 to temporary reduce CI cloud burden


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -316,7 +316,6 @@ var _ = g.Describe("[Feature:Machines] Autoscaler should", func() {
 		initialNumberOfReplicas0 := pointer.Int32PtrDerefOr(targetMachineSet0.Spec.Replicas, e2e.DefaultMachineSetReplicas)
 		initialNumberOfReplicas1 := pointer.Int32PtrDerefOr(targetMachineSet1.Spec.Replicas, e2e.DefaultMachineSetReplicas)
 		glog.Infof("initialNumberOfReplicas0 %d, initialNumberOfReplicas1 %d", initialNumberOfReplicas0, initialNumberOfReplicas1)
-		o.Expect(initialNumberOfReplicas0).To(o.BeNumerically("==", initialNumberOfReplicas1))
```
frobware: If we have two machinesets what prevents us from asserting this still holds true?

enxebre: One of the sets will have 2 replicas. The goal of openshift/release#3615 was to alleviate CI cloud rate limits by dropping one az (hence one machineSet) but keeping number of machines

frobware: Ah, of course. My [current] PR has removed that assumption.

PR #80, opened 2019-05-24
Sync with cluster api 0.1.0, update k8s to 1.14, sigs.k8s.io/controller-runtime to the master HEAD

ingvagabund:
SSIA


Gopkg.toml
```diff
@@ -64,20 +64,20 @@ required = [
...
 
 [[override]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.13.1"
+  version = "kubernetes-1.14.1"
 
 [[override]]
   name = "k8s.io/kube-aggregator"
-  branch = "release-1.13"
+  branch = "release-1.14"
```
cynepco3hahue: Does it any reason why do we use branch here and not the version?

ingvagabund: Copy-paste from a different Gopkg.toml and never changed since. Feel free to change the branch to version. So far it has not have any negative impact on the functionality.

PR #84, opened 2019-06-04
Increase unneeded time to scale down a node

ingvagabund:
Choose a time that is at least twice as the sync period
and that has high least common multiple to avoid a case
when a node is considered to be empty even if there are
pods already scheduled and running on the node.


frobware:
/lgtm
/approve

Interesting debugging session. 
Hope the fix is the same for the upstream PR.

PR #85, opened 2019-06-05
e2e/autoscaler: consume MaxNodesTotalReached event

frobware:
The previous behaviour was to scale out by waiting for the expected
scale up events to be generated on the node groups. After that had
completed we then waited for a further 60s to ensure that there were
no further scale up attempts as we had explicitly capped the cluster
size.

This commit changes the behaviour from waiting for 60s to explicitly
waiting for the MaxNodesTotalReached event to be generated as well. We use
that event to assert that the cluster will not grow beyond the
specified maximum.

This PR also helps to validate that we carry two upstream fixes that
were made with respect to calculating when the maximum size of the
cluster is reached:
    - https://github.com/kubernetes/autoscaler/pull/1516
    - https://github.com/kubernetes/autoscaler/pull/1771

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -183,6 +183,19 @@ func newScaleDownCounter(w *eventWatcher, v uint32) *eventCounter {
 	return c
 }
 
+func newMaxNodesTotalReachedCounter(w *eventWatcher, v uint32) *eventCounter {
+	isAutoscalerMaxNodesTotalEvent := func(event *corev1.Event) bool {
+		return event.Source.Component == clusterAutoscalerComponent &&
+			event.Reason == clusterAutoscalerMaxNodesTotalReached &&
+			event.InvolvedObject.Kind == clusterAutoscalerObjectKind &&
+			strings.HasPrefix(event.Message, "Max total nodes in cluster reached")
+	}
+
+	c := newEventCounter(w, isAutoscalerMaxNodesTotalEvent, v, increment)
+	c.enable()
+	return c
+}
+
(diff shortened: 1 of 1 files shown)


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -282,21 +295,15 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
...
-		// capped size.
-		//
-		// TODO(frobware): switch to matching on
-		// MaxNodesTotalReached when that is available in the
-		// cluster-autoscaler image.
+		// cluster size with maxNodesTotal. We check that we
+		// received the max cluster size reached event.
+		maxNodesTotalReachedCounter := newMaxNodesTotalReachedCounter(eventWatcher, 0)
 		testDuration = time.Now().Add(time.Duration(e2e.WaitShort))
-		o.Consistently(func() bool {
```
enxebre: isn't this complementary check?

frobware: > isn't this complementary check?

Yes, it's back in.

frobware:
/hold cancel

This merged: https://github.com/openshift/kubernetes-autoscaler/pull/102

enxebre:
test in this repo were running against out of date release payload hence not picking latest autoscaler changes.
This should be fixed now https://github.com/openshift/release/pull/4000

frobware:
/hold

don't want this to merge just yet.

frobware:
/hold 

Until I work out why this is failing on k8s-e2e.

frobware:
/hold

Waiting for https://github.com/openshift/kubernetes-autoscaler/pull/103

PR #86, opened 2019-06-10
Dump autoscaler logs before deleting the resource

vikaschoudhary16:
https://jira.coreos.com/browse/CLOUD-537

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
 	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -187,6 +188,30 @@ func remaining(t time.Time) time.Duration {
 	return t.Sub(time.Now()).Round(time.Second)
 }
 
+func dumpClusterAutoscalerLogs(client runtimeclient.Client, restClient *rest.RESTClient) {
+	pods := corev1.PodList{}
+	caLabels := map[string]string{
+		"app": "cluster-autoscaler",
+	}
+	if err := client.List(context.TODO(), &pods, []runtimeclient.ListOptionFunc{runtimeclient.MatchingLabels(caLabels)}...); err != nil {
+		glog.Errorf("Error querying api for clusterAutoscaler pod object: %v", err)
+		return
+	}
+	// We're only expecting one pod but let's log from all that
+	// are found. If we see more than one that's indicative of
+	// some unexpected problem and we may as well dump its logs.
+	for i, pod := range pods.Items {
+		req := restClient.Get().Namespace(e2e.TestContext.MachineApiNamespace).Resource("pods").Name(pod.Name).SubResource("log")
+		res := req.Do()
+		raw, err := res.Raw()
+		if err != nil {
+			glog.Errorf("Unable to get pod logs: %v", err)
+			continue
+		}
+		glog.Infof("\n\nDumping pod logs: %d/%d, logs from %q:\n%v", i, len(pods.Items), pod.Name, string(raw))
+	}
+}
+
 var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 	g.It("scale up and down", func() {
 		defer g.GinkgoRecover()
@@ -198,11 +223,19 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 		client, err = e2e.LoadClient()
 		o.Expect(err).NotTo(o.HaveOccurred())
 
+		var restClient *rest.RESTClient
+		restClient, err = e2e.LoadRestClient()
+		o.Expect(err).NotTo(o.HaveOccurred())
+
 		// Anything we create we must cleanup
 		var cleanupObjects []runtime.Object
 		defer func() {
 			cascadeDelete := metav1.DeletePropagationForeground
 			for _, obj := range cleanupObjects {
+				switch obj.(type) {
+				case *caov1.ClusterAutoscaler:
+					dumpClusterAutoscalerLogs(client, restClient)
+				}
 				if err = client.Delete(context.TODO(), obj, func(opt *runtimeclient.DeleteOptions) {
 					opt.PropagationPolicy = &cascadeDelete
 				}); err != nil {
--- a/pkg/e2e/framework/framework.go
+++ b/pkg/e2e/framework/framework.go
@@ -17,9 +17,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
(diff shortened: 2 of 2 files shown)


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -260,7 +260,6 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 		g.By(fmt.Sprintf("Creating ClusterAutoscaler configured with maxNodesTotal:%v", maxNodesTotal))
 		clusterAutoscaler := clusterAutoscalerResource(maxNodesTotal)
 		o.Expect(client.Create(context.TODO(), clusterAutoscaler)).Should(o.Succeed())
-		cleanupObjects = append(cleanupObjects, runtime.Object(clusterAutoscaler))
```
ingvagabund: If you remove this, CA will keep running which may have negative impact on other tests. Instead, we might just dump ca logs right before CA resource is removed.

vikaschoudhary16: or what about removing the CA at the end of the test, if completed successfully?

ingvagabund: The test run after the failed test will be potentially impacted if the CA resource is not removed.

frobware:
PR title needs updating given the latest changes.

pkg/e2e/autoscaler/autoscaler.go
ingvagabund: `cleanupObjects = append(cleanupObjects, runtime.Object(clusterAutoscaler))` still needs to be kept in the test.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -209,6 +218,29 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 					glog.Errorf("error deleting object: %v", err)
 				}
 			}
+			if clusterAutoscaler != nil {
```
ingvagabund: Please, move the log printing before iteration through `cleanupObjects`.

pkg/e2e/autoscaler/autoscaler.go
vikaschoudhary16: If we append CA ressource to the cleanupObjects, then again will have to find out CA resource at cleaning up to print the logs. Since CA resource is one special case where we are printing logs, thought of not adding it in cleanupObjects and deleting it specifically after handling cleanupObjects in the defer func. 
Ofcourse, what you suggesting is also another approach. We can add CA to cleanupObjects and then at deletion time , check object type for CA resource. If found, print logs.

I prefered first thinking that its only one resource which needs special handling so may not be good idea to first add and then filter out. WDYT?

ingvagabund: You already have the CA object assigned to `clusterAutoscaler` variable.

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -209,6 +218,29 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
...
+				pods := corev1.PodList{}
+				if err := client.List(context.TODO(), &pods, []runtimeclient.ListOptionFunc{runtimeclient.MatchingLabels(caLabels)}...); err != nil {
+					glog.Errorf("Error querying api for clusterAutoscaler pod object: %v", err)
+				}
+				if len(pods.Items) > 1 {
+					glog.Errorf("Got more than one autoscaler pods, which is unexpected!!!")
```
frobware: Equally, if we get zero it goes boom.

pkg/e2e/autoscaler/autoscaler.go
frobware: Is there a reason we can't handle this as a special case in the for loop?

pkg/e2e/autoscaler/autoscaler.go
frobware: > I prefered first thinking that its only one resource which needs special handling so may not be good idea to first add and then filter out. WDYT?

I took this approach: https://github.com/openshift/cluster-api-actuator-pkg/pull/88/commits/eab13a740ef6d5941ca905485118d06017225ef9

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -259,8 +293,8 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 
 		g.By(fmt.Sprintf("Creating ClusterAutoscaler configured with maxNodesTotal:%v", maxNodesTotal))
 		clusterAutoscaler := clusterAutoscalerResource(maxNodesTotal)
-		o.Expect(client.Create(context.TODO(), clusterAutoscaler)).Should(o.Succeed())
 		cleanupObjects = append(cleanupObjects, runtime.Object(clusterAutoscaler))
+		o.Expect(client.Create(context.TODO(), clusterAutoscaler)).Should(o.Succeed())
```
ingvagabund: Any reason to move the line down?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -187,6 +188,31 @@ func remaining(t time.Time) time.Duration {
...
+		glog.Errorf("Error querying api for clusterAutoscaler pod object: %v", err)
+	}
+	if len(pods.Items) > 1 {
+		glog.Errorf("Got more than one autoscaler pods, which is unexpected!!!")
+	}
+	// We're only expecting one pod but let's log from all that
+	// are found. If we see more than one that's indicative of
+	// some unexpected problem and we may as well dump its logs.
+	for _, pod := range pods.Items {
+		req := restClient.Get().Namespace(e2e.TestContext.MachineApiNamespace).Resource("pods").Name(pod.Name).SubResource("log")
```
enxebre: is this not supported on controller runtime yet?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -187,6 +188,31 @@ func remaining(t time.Time) time.Duration {
 	return t.Sub(time.Now()).Round(time.Second)
 }
 
+func dumpClusterAutoscalerLogs(client runtimeclient.Client, restClient *rest.RESTClient) {
+	pods := corev1.PodList{}
+	caLabels := map[string]string{
+		"app": "cluster-autoscaler",
+	}
+	if err := client.List(context.TODO(), &pods, []runtimeclient.ListOptionFunc{runtimeclient.MatchingLabels(caLabels)}...); err != nil {
+		glog.Errorf("Error querying api for clusterAutoscaler pod object: %v", err)
```
frobware: Why do we not return here?

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -187,6 +188,31 @@ func remaining(t time.Time) time.Duration {
...
+	// are found. If we see more than one that's indicative of
+	// some unexpected problem and we may as well dump its logs.
+	for _, pod := range pods.Items {
+		req := restClient.Get().Namespace(e2e.TestContext.MachineApiNamespace).Resource("pods").Name(pod.Name).SubResource("log")
+		res := req.Do()
+		raw, err := res.Raw()
+		if err != nil {
+			glog.Errorf("Unable to get pod logs: %v", err)
+		}
+		glog.Infof("\n\n Pod %q logs:\n\n%v", pod.Name, string(raw))
```
frobware: Nit: drop the [:space:] between newline and 'Pod'.

pkg/e2e/framework/framework.go
```diff
@@ -426,6 +428,24 @@ func LoadClient() (runtimeclient.Client, error) {
 	return runtimeclient.New(config, runtimeclient.Options{})
 }
 
+func LoadRestClient() (*rest.RESTClient, error) {
```
frobware: Can we turn this change into its own discrete commit?

pkg/e2e/autoscaler/autoscaler.go
frobware: To capture how many Pods were found consider logging as:

```go
glog.Infof("Dumping pod logs: %d/%d, logs:\n%v", i, len(pods), pod.Name, string(raw))
```

You could then dispense with the conditional:

```go
if len(pods.Items) > 1 {
	glog.Errorf("Got more than one autoscaler pods, which is unexpected!!!")
}
```

pkg/e2e/autoscaler/autoscaler.go
vikaschoudhary16: AFAIU, No!

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -187,6 +188,29 @@ func remaining(t time.Time) time.Duration {
...
+	}
+	// We're only expecting one pod but let's log from all that
+	// are found. If we see more than one that's indicative of
+	// some unexpected problem and we may as well dump its logs.
+	for i, pod := range pods.Items {
+		req := restClient.Get().Namespace(e2e.TestContext.MachineApiNamespace).Resource("pods").Name(pod.Name).SubResource("log")
+		res := req.Do()
+		raw, err := res.Raw()
+		if err != nil {
+			glog.Errorf("Unable to get pod logs: %v", err)
```
frobware: Should we not `continue` here?

frobware: We need to handle the case where this can fail. e.g., https://github.com/openshift/cluster-api-actuator-pkg/pull/88/commits/2c1c380483f4d90d324122f50d50b3373995a919#diff-88da310e1da5f7c75bfb19be49e0a89fR207

frobware:
@vikaschoudhary16 - sorry, clicked wrong button.

/lgtm

PR #93, opened 2019-06-27
Increase workload and timeouts for Azure

ingvagabund:
It just takes longer and worker nodes have more resource capacity.

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -77,8 +77,8 @@ func newWorkLoad() *batchv1.Job {
 				},
 			},
 			BackoffLimit: pointer.Int32Ptr(4),
-			Completions:  pointer.Int32Ptr(50),
-			Parallelism:  pointer.Int32Ptr(50),
+			Completions:  pointer.Int32Ptr(100),
+			Parallelism:  pointer.Int32Ptr(100),
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -77,8 +77,8 @@ func newWorkLoad() *batchv1.Job {
 				},
 			},
 			BackoffLimit: pointer.Int32Ptr(4),
-			Completions:  pointer.Int32Ptr(50),
-			Parallelism:  pointer.Int32Ptr(50),
+			Completions:  pointer.Int32Ptr(100),
```
ingvagabund: As @frobware suggested, we need to dynamically estimate size of the workload based on available node resources.

enxebre: adding a TODO would be nice

pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -372,5 +377,15 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 				remaining(testDuration), len(nodes), len(currentNodes))
 			return len(currentNodes)
 		}, e2e.WaitMedium, pollingInterval).Should(o.Equal(len(nodes)))
+
+		g.By("Waiting for scaled up machines to be deleted")
```
enxebre: what was the motivation for this addtion?

ingvagabund: Since it takes longer to delete a vm in Azure.

PR #94, opened 2019-07-02
pkg/e2e/operators: Add tests for rejecting invalid CAO resources

bison:
This adds two small tests to ensure that invalid ClusterAutoscaler and
MachineAutoscaler resources are rejected early by the validating
webhooks deployed by the cluster-autoscaler-operator.


bison:
This should not pass until https://github.com/openshift/cluster-autoscaler-operator/pull/108 merges.

PR #95, opened 2019-07-05
README: Add instructions for running autoscaler tests under minikube

frobware:
Add broad instructions for running the cluster autoscaler e2e tests using a local minikube cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,241 @@
 # cluster-api-actuator-pkg
 
 Shared packages for Cluster API actuators.
+
+## Runing the cluster autoscaler e2e tests locally using minikube
+
+You can run the autoscaler e2e tests locally using minikube with the
+following setup:
+
+### Start minikube
+
+```console
+$ minikube start --vm-driver kvm2 --kubernetes-version v1.13.4 --v 5 --memory=25165
+* minikube v1.2.0 on linux (amd64)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,77 @@
...
+--- PASS: TestE2E (0.00s)
+PASS
+ok      github.com/openshift/cluster-api-actuator-pkg/pkg/e2e   0.037s
+```
+
+Adjust `-ginkgo.focus` as appropriate.
+
+### Cleanup
+
+The tests (during development) will / can fail and I use the following
```
vikaschoudhary16: s/I/@frobware? :)

README.md
```diff
@@ -1,3 +1,77 @@
 # cluster-api-actuator-pkg
 
 Shared packages for Cluster API actuators.
+
+## Runing the e2e tests locally using minikube
```
enxebre: how would this work without running the mao which is a requirement for cao and the clusteroperator CRD?

README.md
frobware: Done in f60a9484...bef6ae54

README.md
frobware: > how would this work without running the mao which is a requirement for cao and the clusteroperator CRD?

Made the comment more specific - now only caters for autoscaler tests.

frobware:
/hold 

Updating instructions with output from running the individual commands.

PR #97, opened 2019-07-18
Add e2e for node drain

vikaschoudhary16:
This PR is adding a new e2e test case for verification that before machine delete, all pods are drained from the node.
Existing test case, which is actually verifying PDB feature, is named as node drain test. This naming is being fixed.

Node drain test does following:
1. Creates two Machines and wait for nodes to get ready
2. Creates a server pod on machine2
3. Creates a replication controller with 20 replicas. container has a preStop hook, which posts its hostname at the server.
4. Wait for replicas to get ready
5. determine how many pods are on machine1
6. Delete machine1
7. Ensure that all the pods which were on machine1, posted their hostname to the  server. which means all pods gets deleted gracefully before machine delete.


Gopkg.lock
```diff
@@ -1214,6 +1223,7 @@
     "k8s.io/client-go/tools/clientcmd",
     "k8s.io/client-go/tools/clientcmd/api",
     "k8s.io/code-generator/cmd/deepcopy-gen",
+    "k8s.io/kubernetes/test/utils/image",
```
michaelgugino: Let's just copy this in tree so we're no importing 2.2 million lines of code.

PR #100, opened 2019-07-26
e2e/autoscaler: dynamically compute workload size

frobware:
Rather than guess at the number of jobs to create--which triggers the
scale up---we now compute that value dynamically.

The computation is based on the number of nodes in the cluster and to
ensure that one we only get one pod per node the memory requirement is
set at 70% of a node's overall memory capacity.

The motivation for this change is to reduce the overall number of
events generated. As this test explicitly requires explicit events to
be generated--and consumed--we now stand a better chance of not having
critical events (for the test) dropped.

When test runs we now see a limited set of events related to scale-up:

```console
cluster-autoscaler-status: Scale-up: setting group openshift-machine-api/amcdermo-st695-worker-us-east-2a size to 2
cluster-autoscaler-status: Max total nodes in cluster reached: 2
cluster-autoscaler-status: Scale-up: group openshift-machine-api/amcdermo-st695-worker-us-east-2a size set to 2
workload-txh5f: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-lz7fs: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-b45pn: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-s85hv: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-5vk42: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-c7brq: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-4z26f: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-qlnjx: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
workload-ntmph: pod triggered scale-up: [{openshift-machine-api/amcdermo-st695-worker-us-east-2a 1->2 (max: 2)}]
[19m27s remaining] Expecting 2 "ScaledUpGroup" events; observed 1
[19m24s remaining] Expecting 2 "ScaledUpGroup" events; observed 1
[19m21s remaining] Expecting 2 "ScaledUpGroup" events; observed 1
cluster-autoscaler-status: Max total nodes in cluster reached: 1
```

Prior to this change we would get hundreds of `pod triggered scale-up`
events reported.

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -35,7 +35,7 @@ const (
 	pollingInterval                       = 3 * time.Second
 )
 
-func newWorkLoad() *batchv1.Job {
+func newWorkLoad(njobs int32, memoryRequest resource.Quantity) *batchv1.Job {
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "workload",
@@ -59,7 +59,7 @@ func newWorkLoad() *batchv1.Job {
 							},
 							Resources: corev1.ResourceRequirements{
 								Requests: corev1.ResourceList{
-									"memory": resource.MustParse("500Mi"),
+									"memory": memoryRequest,
 									"cpu":    resource.MustParse("500m"),
 								},
 							},
@@ -78,8 +78,8 @@ func newWorkLoad() *batchv1.Job {
 				},
 			},
 			BackoffLimit: pointer.Int32Ptr(4),
-			Completions:  pointer.Int32Ptr(100),
-			Parallelism:  pointer.Int32Ptr(100),
+			Completions:  pointer.Int32Ptr(njobs),
+			Parallelism:  pointer.Int32Ptr(njobs),
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/e2e/autoscaler/autoscaler.go
```diff
@@ -385,7 +394,10 @@ var _ = g.Describe("[Feature:Machines][Serial] Autoscaler should", func() {
 		}
 		scaleUpCounter := newScaleUpCounter(eventWatcher, 0, scaledGroups)
 		maxNodesTotalReachedCounter := newMaxNodesTotalReachedCounter(eventWatcher, 0)
-		workload := newWorkLoad()
+		bytes, ok := memCapacity.AsInt64()
+		o.Expect(ok).Should(o.BeTrue())
+		// set job load to 70% of available memory per instance.
+		workload := newWorkLoad(int32(maxNodesTotal+1), resource.MustParse(fmt.Sprintf("%v", bytes*70/100)))
```
ingvagabund: Not: I think it's safe to directly use `0.7` coefficient instead of explicitly describing 70%.

Otherwise, 70% is sufficient. Enough capacity and not too much to cover existing and "to-be-created" use cases.

PR #104, opened 2019-08-06
e2e/autoscaler: create all machinesets AOT

frobware:
The autoscaler test now creates its own machinesets Ahead-Of-Time
rather than relying on what is available on the cluster.

The intent is for the test to be more resilient to other tests failing
in the e2e suite which typically leave nodes/machines hanging around
and that perturbs the expectations for scale out (i.e., we end up with
$N more nodes that are eligible for scheduling than expected).

### DIFF
--- a/pkg/e2e/autoscaler/autoscaler.go
+++ b/pkg/e2e/autoscaler/autoscaler.go
@@ -33,6 +33,7 @@ const (
 	clusterAutoscalerScaleDownEmpty       = "ScaleDownEmpty"
 	clusterAutoscalerMaxNodesTotalReached = "MaxNodesTotalReached"
 	pollingInterval                       = 3 * time.Second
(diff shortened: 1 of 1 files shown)


frobware:
Examples of where this occurs today:
- https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/periodic-ci-openshift-cluster-autoscaler-operator-e2e-azure-master/39/

ingvagabund:
Sad to see the tests will run longer. Though, don't see any better way.

frobware:
> Sad to see the tests will run longer. Though, don't see any better way.

I see this as possibly a temporary step. I would rather get a clean bill of health more often whilst looking at which tests fail to clean up. The alternative? We see it flake and just /retest.

frobware:
/hold

I want to understand whether #107 has had a much bigger impact on reliable testing.

PR #108, opened 2019-08-09
e2e/infra: dump POD status in waitUntilAllRCPodsAreReady

frobware:
Sometimes this will timeout because Status.Replicas !=
Status.ReadyReplicas. Print the state of all the pods for debugging
purposes so we can distinguish between the cases when it works and
those rarer cases when it doesn't.

### DIFF
--- a/pkg/e2e/infra/utils.go
+++ b/pkg/e2e/infra/utils.go
@@ -2,6 +2,7 @@ package infra
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/e2e/infra/utils.go
```diff
@@ -398,6 +399,21 @@ func waitUntilAllRCPodsAreReady(client runtimeclient.Client, rc *corev1.Replicat
 		glog.Infof("[%s remaining] Waiting for RC ready replicas, ReadyReplicas: %v, Replicas: %v", remainingTime(endTime), rcObj.Status.ReadyReplicas, rcObj.Status.Replicas)
 		return rcObj.Status.Replicas == rcObj.Status.ReadyReplicas, nil
 	})
+
+	// Sometimes this will timeout because Status.Replicas !=
+	// Status.ReadyReplicas. Print the state of all the pods for
+	// debugging purposes so we can distinguish between the cases
+	// when it works and those rare cases when it doesn't.
+	pods := corev1.PodList{}
```
enxebre: can we filter to get only the pods owned by the replicaset?

frobware: Listing on rc.Namespace is too wide?

enxebre: filtering to the controller seems more appropriate for the `func waitUntilAllRCPodsAreReady` but I'm fine leaving to the ns too

frobware: > filtering to the controller seems more appropriate for the `func waitUntilAllRCPodsAreReady` but I'm fine leaving to the ns too

Updated.

PR #118, opened 2019-11-02
e2e: Add MachineHealthCheck tests for v1beta1

bison:
This (re)adds an E2E test suite and associated helpers to cover the MachineHealthCheck feature. This now works with version v1beta1 of the MachineHealthCheck API. The tests operate on a dedicated MachineSet, which should allow things to be paralellized more easily in the future.

### DIFF
--- a/pkg/e2e/framework/nodes.go
+++ b/pkg/e2e/framework/nodes.go
@@ -2,7 +2,9 @@ package framework
 
 import (
 	"context"
+	"fmt"
 
+	mapiv1beta1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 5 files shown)


pkg/e2e/framework/nodes.go
```diff
@@ -15,6 +17,22 @@ func AddNodeCondition(c client.Client, node *corev1.Node, cond corev1.NodeCondit
 	return c.Status().Patch(context.Background(), nodeCopy, client.MergeFrom(node))
 }
 
+// GetNodeForMachine retrieves the node backing the given Machine.
+func GetNodeForMachine(c client.Client, m *mapiv1beta1.Machine) (*corev1.Node, error) {
```
enxebre: There're helpers for this in infra/utils `getNodeFromMachine` already
We'll want to consolidate the reusable library in the upcoming follow ups.

bison: Yeah for sure. That's one of the frustrating things I ran into. There's also a bunch of helpers there, but almost none of them are exported for some reason.

PR #119, opened 2019-11-08
Allow setting BUILD_DEST and add Dockerfile

bison:
This adds a Dockerfile so Prow can build an image containing the test suite.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-api-actuator-pkg
+COPY . .
+ENV NO_DOCKER=1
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-api-actuator-pkg
+COPY . .
+ENV NO_DOCKER=1
+ENV BUILD_DEST=/go/bin/cluster-api-e2e
+RUN unset VERSION && GOPROXY=off make build-e2e
+CMD ["/usr/bin/cluster-api-e2e"]
```
enxebre: should this path match `BUILD_DEST`?

bison: WHOOPS. Fixed.

PR #121, opened 2019-11-15
Remove unused Framework and TestContext types

bison:
This removes the unused Framework struct and its methods. It also removes the TestContext type and the single global instance of it, which was just a container for a few mostly unused variables.


bison:
/hold

It looks to me like the removed `Framework` type is only used by cluster-api-provider-libvirt at this point. See [here](https://github.com/openshift/cluster-api-provider-libvirt/blob/master/test/machines/machines_test.go#L16). I'm not sure CI is even running those tests though.

@zeenix, are we good to remove this? This bit is not maintained, so if you are using those tests, I'd plan to move them off this.

ingvagabund:
@bison that CI is dead. Jenkins job is no longer running. libvirt-actuator uses CI provided by openshift (e2e-libvirt) instead.

zeenix:
> @zeenix, are we good to remove this? This bit is not maintained, so if you are using those tests, I'd plan to move them off this.

Yeah, sure. I don't have `lgtm` power here it seems so someone else will have to do this bit. :)

PR #122, opened 2019-11-20
A few small helper refactors

bison:
This updates the vendored dependencies after having removed a bunch of code. It also uses controller-runtime for loading client configs, which removes the last command-line flag we were parsing here. Finally, none of the shared helpers accept a context now. There wasn't a single instance where we passed an existing context that we might use for cancellation, so let's make the function signatures smaller.


bison:
/hold cancel

The hold was an accident. Oops.

PR #123, opened 2019-11-22
Add a random ID to machine names in infra tests

bison:
Adding random ID to machine names in infra tests.


bison:
Closing in favor of using Prow to cherry-pick.

PR #128, opened 2019-12-05
Bug 1796386: Create different machineset for every test and enable parallelism

alexander-demichev:
**Enable parallelism**
- Create different machineset for every test within the infra package.
- Update ginkgo and gomega and fetch ginkgo CLI.
- Run tests in parallel.

**Before**:
AWS: ~31min
GCP: ~30min
Azure: ~1h20min

**After**:
AWS: ~16min
GCP: ~18min
Azure: ~55min

**Follow up:**
- Enable Autoscaler to run in parallel with the other `[Feature:Machines]` specs.
- Enable and expose JUnit output on prow dashboard.
- Remove all unused code.
- Remove two levels folder-tree in `pkg/`.
- Move any util to manipulate resources into the framework pkg and let all other packages consume from there as they see fit.


pkg/e2e/infra/infra.go
```diff
@@ -196,77 +236,43 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
+		err = deleteMachine(client, machine)
+		Expect(err).NotTo(HaveOccurred())
+
+		e2e.WaitForMachineSet(client, machineSet.GetName())
 	})
 
-	g.It("grow and decrease when scaling different machineSets simultaneously", func() {
-		client, err := e2e.LoadClient()
-		o.Expect(err).NotTo(o.HaveOccurred())
+	It("grow and decrease when scaling different machineSets simultaneously", func() {
```
alexander-demichev: Looks like this test requires 2 different machinesets, I'll find a way to make it work

pkg/e2e/infra/infra.go
```diff
@@ -6,9 +6,10 @@ import (
 	"strings"
 	"time"
 
+	. "github.com/onsi/ginkgo"
```
enxebre: why is this change?
If it's needed this should be in a separate commit, otherwise this introduce a lot of noise in this PR due to all the references to ginkgo/gomega making more difficult to review the real logic

pkg/e2e/infra/infra.go
```diff
@@ -328,75 +330,34 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
-		})
-		o.Expect(err).NotTo(o.HaveOccurred())
-
-		g.By("Creating two new machines, one for node about to be drained, other for moving workload from drained node")
+		By("Creating two new machines, one for node about to be drained, other for moving workload from drained node")
 		// Create two machines
 		randID := e2e.RandomString()
-		machine1 := machineFromMachineset(&machinesets.Items[0])
+		machine1 := machineFromMachineset(machineSet)
 		machine1.Name = "machine1-" + randID
```
enxebre: there's no need to create these machines now, you already got available the one created by the test scoped machineSet

pkg/e2e/infra/utils.go
```diff
@@ -265,8 +256,20 @@ func waitForClusterSizeToBeHealthy(client runtimeclient.Client, targetSize int)
 	}
```
alexander-demichev: `waitForClusterSizeToBeHealthy` is not used anymore, not sure if we should keep it for future

pkg/e2e/infra/infra.go
```diff
@@ -123,40 +121,74 @@ func invalidMachinesetWithEmptyProviderConfig() *mapiv1beta1.MachineSet {
...
+	g.AfterEach(func() {
+		g.By("Deleting the new MachineSet")
+		err := client.Delete(context.Background(), machineSet)
+		o.Expect(err).ToNot(o.HaveOccurred())
+
+		e2e.WaitForMachineSetDelete(client, machineSet)
+	})
+
+	g.It("have machines linked with nodes", func() {
+		selector := machineSet.Spec.Selector
```
enxebre: I actually don't think we need this test anymore since this is validated every time we `WaitForMachineSet` beforeEach test

pkg/e2e/infra/infra.go
```diff
@@ -197,232 +229,88 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
-				&targetMachineSet.Spec.Template.Spec.ProviderSpec,
-				0) // zero replicas
-			machineSets[i].Spec.Template.Spec.ObjectMeta.Labels = map[string]string{
-				e2e.WorkerNodeRoleLabel: "",
-			}
-			o.Expect(client.Create(context.TODO(), machineSets[i])).Should(o.Succeed())
-			cleanupObjects = append(cleanupObjects, runtime.Object(machineSets[i]))
-		}
+		g.By("Creating a second MachineSet")
+		machineSetParams := buildMachineSetParams(client, 0)
```
enxebre: I think this whole test can be simplified by something like:
```
machineSet2, err := e2e.CreateMachineSet(client, machineSetParams)
e2e.WaitForMachineSet(client, machineSet2.GetName())
scaleMachineSet(machineSet1, 0)
scaleMachineSet(machineSet2, 3)
e2e.WaitForMachineSet(client, machineSet1.GetName())
e2e.WaitForMachineSet(client, machineSet2.GetName())
```
Once we run the tests in parallel we can probably just drop it

pkg/e2e/infra/infra.go
```diff
@@ -197,232 +229,88 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
-			delObjects["machine2"] = machine2
-
-			return nil
-		}()
-		o.Expect(err).NotTo(o.HaveOccurred())
-
-		g.By("Waiting until both new nodes are ready")
-		err = waitUntilNodesAreReady(client, []runtimeclient.ListOption{runtimeclient.MatchingLabels(nodeDrainLabels)}, 2)
-		o.Expect(err).NotTo(o.HaveOccurred())
+		machine := machines[0]
```
enxebre: you need to apply the `nodeDrainLabels` to the machines used for the test here.
Also should probably use machines with no machineSet to be able to rely on `verifyNodeDraining` otherwise a new machine will come up as you delete it.

pkg/e2e/infra/infra.go
```diff
@@ -197,232 +223,63 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
+		// Add node draining labels to params
+		for k, v := range nodeDrainLabels {
+			machineSetParams.Labels[k] = v
+		}
 
-			if err := client.Create(context.TODO(), machine2); err != nil {
-				return fmt.Errorf("unable to create machine %q: %v", machine2.Name, err)
-			}
-			delObjects["machine2"] = machine2
+		machine, err := e2e.CreateMachine(client, machineSetParams)
```
enxebre: seems this is only creating one machine with `nodeDrainLabels`. So there's only one machine which satisfies the `replicationControllerWorkload` `NodeSelector`. So when draining the eviction API won't be able to find another suitable node.
Instead of creating a new machine, couldn't we use 2 machines of the ones created by the beforeEach machineSet and apply them the `nodeDrainLabels`? So
- Fetch 2 machines from machineSet
- Apply `nodeDrainLabels`
- Run workload (same as now)
- Delete one machine (same as now)
- verifyNodeDraining (same as now)

pkg/e2e/framework/common.go
```diff
@@ -273,6 +273,61 @@ type MachineSetParams struct {
...
+
+	return machine, nil
+}
+
+// WaitForMachine waits for the machine to enter the "Running" phase and for node to be ready
+func WaitForMachine(c client.Client, name string) {
+	Eventually(func() error {
+		machine, err := GetMachine(c, name)
+		Expect(err).ToNot(HaveOccurred())
+
```
enxebre: I think we should to validate the machine has no deletion timestamp here, and also probably check phase is running.
But base on https://github.com/openshift/cluster-api-actuator-pkg/pull/128/files#r363689909 may be we don't need this two methods

pkg/e2e/infra/infra.go
```diff
@@ -123,40 +121,68 @@ func invalidMachinesetWithEmptyProviderConfig() *mapiv1beta1.MachineSet {
...
+func buildMachineSetParams(client runtimeclient.Client, replicas int) e2e.MachineSetParams {
+	// Get the current workers MachineSets so we can copy a ProviderSpec
+	// from one to use with our new dedicated MachineSet.
+	workers, err := e2e.GetWorkerMachineSets(client)
+	o.Expect(err).ToNot(o.HaveOccurred())
+
+	providerSpec := workers[0].Spec.Template.Spec.ProviderSpec.DeepCopy()
+	clusterName := workers[0].Spec.Template.Labels[e2e.ClusterKey]
+
+	msName := e2e.RandomString()
```
enxebre: This needs rebase and follow https://github.com/openshift/cluster-api-actuator-pkg/pull/131/files#diff-798770aa2c88fce8a6789bb252bcb627R394

enxebre:
can you please squash commits reasonable or rephrase the one which says "More e2e test improvements"?

pkg/e2e/infra/infra.go
enxebre: this is marked as resolved but the comment still applies

pkg/e2e/framework/common.go
```diff
@@ -41,23 +39,9 @@ const (
 
 // RandomString returns a random 6 character string.
 func RandomString() string {
-	randID := string(uuid.NewUUID())
+	randID := "e2e-" + string(uuid.NewUUID())
```
enxebre: can we please make sure we follow this pattern https://github.com/openshift/cluster-api-actuator-pkg/pull/131/files#diff-798770aa2c88fce8a6789bb252bcb627R394-R397

pkg/e2e/infra/infra.go
```diff
@@ -197,243 +223,70 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
+		o.Expect(err).ToNot(o.HaveOccurred())
 
-		g.By("Waiting until both new nodes are ready")
-		err = waitUntilNodesAreReady(client, []runtimeclient.ListOption{runtimeclient.MatchingLabels(nodeDrainLabels)}, 2)
-		o.Expect(err).NotTo(o.HaveOccurred())
+		err = client.Update(context.TODO(), machines[1])
+		o.Expect(err).ToNot(o.HaveOccurred())
 
 		g.By("Creating RC with workload")
 		rc := replicationControllerWorkload("default")
```
enxebre: can we please make sure this is always deleted before the end of the test as it was via `delObjects`

pkg/e2e/infra/infra.go
```diff
@@ -197,243 +223,70 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
 		err = client.Create(context.TODO(), rc)
 		o.Expect(err).NotTo(o.HaveOccurred())
-		delObjects["rc"] = rc
 
 		g.By("Creating PDB for RC")
 		pdb := podDisruptionBudget("default")
```
enxebre: can we please make sure this is always deleted before the end of the test as it was via delObjects

pkg/e2e/infra/infra.go
```diff
@@ -485,22 +316,10 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
 		g.By("Delete PDB")
 		err = client.Delete(context.TODO(), pdb)
 		o.Expect(err).NotTo(o.HaveOccurred())
```
enxebre: wouldn't this test always get stuck deleting the machineSet during afterEach as the only machine running the  pods and suitable for it will get stuck draining the node

pkg/e2e/infra/utils.go
```diff
@@ -30,25 +30,15 @@ const (
 	machineAPIGroup  = "machine.openshift.io"
 )
 
-func isOneMachinePerNode(client runtimeclient.Client) bool {
-	machineList := mapiv1beta1.MachineList{}
+func isOneMachinePerNode(client runtimeclient.Client, machines []*mapiv1beta1.Machine) bool {
```
enxebre: why this change, this does not seems to be used at all?

pkg/e2e/infra/utils.go
```diff
@@ -265,8 +256,20 @@ func waitForClusterSizeToBeHealthy(client runtimeclient.Client, targetSize int)
 	}
 
 	glog.Infof("waiting for each node to be backed by a machine")
```
enxebre: Why this change, this does not seem to be used anymore? should be deleted

enxebre:
can you please rephrase the commit message?

pkg/e2e/framework/common.go
```diff
@@ -40,10 +40,10 @@ const (
 )
 
 // RandomString returns a random 6 character string.
-func RandomString() string {
+func RandomString(clusterName string) string {
 	randID := string(uuid.NewUUID())
 
-	return randID[:6]
+	return clusterName + randID[:6]
```
enxebre: can we `fmt.Sprintf("%s-%s",
			c clusterName,
			randID,
		)`
to make it more readable and be consistent with https://github.com/openshift/installer/blob/master/pkg/asset/machines/gcp/machinesets.go#L45

hack/ci-integration.sh
```diff
@@ -1,7 +1,9 @@
...
-  -kubeconfig ${KUBECONFIG:-~/.kube/config} \
-  -args -v 5 -logtostderr \
-  "$@"
+go run ./vendor/github.com/onsi/ginkgo/ginkgo \
+    -timeout 90m \
+    -p -stream \
+    -v \
+    -failFast \
+    "$@" \
+    ./pkg/e2e/ -- --alsologtostderr -v 4 -kubeconfig ${KUBECONFIG:-~/.kube/config}
```
JoelSpeed: Before, the tests were using `-ginkgo.noColor=true`, should that be included that here? (Does it make a difference?)

pkg/e2e/infra/infra.go
```diff
@@ -123,40 +122,90 @@ func invalidMachinesetWithEmptyProviderConfig() *mapiv1beta1.MachineSet {
...
-		var err error
-		client, err := e2e.LoadClient()
-		o.Expect(err).NotTo(o.HaveOccurred())
+		selector := machineSet.Spec.Selector
+		machines, err := e2e.GetMachines(client, &selector)
+		o.Expect(err).ToNot(o.HaveOccurred())
 
-		machines, err := e2e.GetMachines(client)
-		o.Expect(err).NotTo(o.HaveOccurred())
-		o.Expect(len(machines)).To(o.BeNumerically(">", 0))
```
JoelSpeed: Might this check be worth keeping before accessing a potentially empty list on the next line? Alternatively `o.Expect(machines).ToNot(o.BeEmpty())` is probably a nicer way to do it

pkg/e2e/infra/infra.go
```diff
@@ -197,231 +246,73 @@ var _ = g.Describe("[Feature:Machines] Managed cluster should", func() {
...
-		var err error
-		client, err := e2e.LoadClient()
-		o.Expect(err).NotTo(o.HaveOccurred())
+		g.By("Create a machine for node about to be drained")
 
-		g.By("checking existing cluster size")
-		existingClusterSize, err := getClusterSize(client)
-		o.Expect(err).NotTo(o.HaveOccurred())
+		selector := machineSet.Spec.Selector
+		machines, err := e2e.GetMachines(client, &selector)
```
JoelSpeed: This test expects machines to contain at least two items, should check that `o.Expect(len(machines)).To(o.BeNumerically(">=", 2))`

JoelSpeed:
/lgtm

Happy to have my comments addressed in a follow up

enxebre:
thanks @JoelSpeed 
I'll do the honours on your behalf seems the bot complains.
/lgtm
@alexander-demichev mind making sure we put a follow up right away for:
https://github.com/openshift/cluster-api-actuator-pkg/pull/128#discussion_r372846347
https://github.com/openshift/cluster-api-actuator-pkg/pull/128#discussion_r372843786
https://github.com/openshift/cluster-api-actuator-pkg/pull/128#discussion_r372839593

enxebre:
Addressed https://github.com/openshift/cluster-api-actuator-pkg/pull/128#discussion_r372839593
/lgtm

pkg/e2e/infra/infra.go
alexander-demichev: :+1:

PR #129, opened 2020-01-10
[OCPCLOUD-763] Update instructions for running e2e tests

JoelSpeed:
I've removed the instructions for running the tests against Minikube as these are out of date and do not work.

In their place I have added instructions for users to set up a cluster using IPI, disable CVO, deploy their test code (in the cluster or locally) and then have left the instructions at the bottom for the building and running the e2e tests that were there before

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,192 +2,63 @@
 
 Shared packages for Cluster API actuators.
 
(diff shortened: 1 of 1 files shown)


enxebre:
awesome @JoelSpeed thanks! /approve

PR #136, opened 2020-02-03
Remove import aliases to make tests more readable

alexander-demichev:
Remove import aliases to make tests more readable.


enxebre:
wouldn't this more difficult with an IDE to check by autocompleting which functions provide each library?
Also if we do the change we'd want to do it everywhere consistently, not only in infra.

alexander-demichev:
I don't think it can be a problem for IDE. Let me see where else these aliases can be removed.

JoelSpeed:
/lgtm

They're `.` imported in most places I've ever seen it so no reason not to be the same

PR #138, opened 2020-02-24
Cleanup: Remove unused code and restructure

alexander-demichev:
This PR: 
- Removes `e2e/` directory and places all test packages right inside `pkg/`
- Moves code from `utils.go` to `framework` package, so we can have a unique place for helper functions
- Intoduces `machines.go` and `machinesets.go` inside `framework` as a place for code that handles machines and machinesets
- Moves all code from `common.go` inside `framework` to other appropriate files.
- Some other minor code reorganizations in `framework`
- Removes `manifests` and `types` packages
- Removes unused functions from  `utils.go`


enxebre:
It does not need to go in this PR but https://github.com/openshift/cluster-api-actuator-pkg/tree/master/pkg/manifests and https://github.com/openshift/cluster-api-actuator-pkg/tree/master/pkg/types are probably no needed anymore. Also there're unused functions we can get rid of here https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/e2e/infra/utils.go

alexander-demichev:
I think I removed all unused functions from 'utils.go'.

enxebre:
/lgtm
on behalf of @JoelSpeed https://github.com/openshift/cluster-api-actuator-pkg/pull/138#issuecomment-591886778

cfergeau:
`cluster-api-provider-libvirt` still has some test code making use of `manifests` https://github.com/openshift/cluster-api-provider-libvirt/blob/master/cmd/libvirt-actuator/main.go#L233-L247 is there a replacement for the API which was removed?

PR #140, opened 2020-03-18
Validate scale from/to zero

enxebre:
Adds a validation for scaling from zero.
/hold
Needs https://github.com/openshift/kubernetes-autoscaler/pull/137

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -450,7 +450,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		}
 	})
 
-	It("It scales from zero", func() {
+	It("It scales from/to zero", func() {
 
 		// Only run in platforms which support autoscaling from/to zero.
 		clusterInfra, err := framework.GetInfrastructure(client)
@@ -515,6 +515,18 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		By("Waiting for the machineSet replicas to become nodes")
 		framework.WaitForMachineSet(client, machineSet.GetName())
 
-		//TODO: Remove workload, drain nodes and validate scale to zero.
+		expectedReplicas = 0
+		By("Deleting the workload")
+		Expect(deleteObject(workload.Name, cleanupObjects[workload.Name])).Should(Succeed())
+		delete(cleanupObjects, workload.Name)
+		Eventually(func() bool {
+			ms, err := framework.GetMachineSet(client, machineSet.GetName())
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("Waiting for machineSet replicas to scale in. Current replicas are %v, expected %v.",
+				*ms.Spec.Replicas, expectedReplicas))
+
+			return *ms.Spec.Replicas == expectedReplicas
+		}, framework.WaitLong, pollingInterval).Should(BeTrue())
 	})
 })
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/glog"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
 	mapiv1beta1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	caov1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1"
@@ -36,8 +37,8 @@ const (
 	workloadJobName                       = "e2e-autoscaler-workload"
 )
 
-func newWorkLoad(njobs int32, memoryRequest resource.Quantity) *batchv1.Job {
-	return &batchv1.Job{
+func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string) *batchv1.Job {
+	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      workloadJobName,
 			Namespace: "default",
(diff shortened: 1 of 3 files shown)


pkg/autoscaler/autoscaler.go
```diff
@@ -204,25 +205,48 @@ func remaining(t time.Time) time.Duration {
 }
 
 var _ = Describe("[Feature:Machines] Autoscaler should", func() {
+	defer GinkgoRecover()
```
JoelSpeed: Anything going on in goroutines to require this?  (/me starts hunting)

pkg/autoscaler/autoscaler.go
```diff
@@ -434,4 +446,72 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 			framework.WaitForMachineSetDelete(client, ms)
 		}
 	})
+
+	It("It scales from zero", func() {
+		defer GinkgoRecover()
```
JoelSpeed: Don't think there's any go routines calling gomega matchers to need a recover? Or is this needed for calling `Expect` within and `Eventually`?

pkg/autoscaler/autoscaler.go
```diff
@@ -434,4 +446,72 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+
+		machineSet, err := framework.CreateMachineSet(client, machineSetParams)
+		Expect(err).ToNot(HaveOccurred())
+		cleanupObjects[machineSet.GetName()] = runtime.Object(machineSet)
+
+		framework.WaitForMachineSet(client, machineSet.GetName())
+
+		By(fmt.Sprintf("Creating ClusterAutoscaler "))
+		clusterAutoscaler := clusterAutoscalerResource(100)
+		Expect(client.Create(context.TODO(), clusterAutoscaler)).Should(Succeed())
```
JoelSpeed: Might be worth setting a variable at the top of this describe `ctx = context.Background()` and using that instead of `TODO()` everywhere?

pkg/autoscaler/autoscaler.go
enxebre: I assumed this is needed for the `Eventually` async assertions

pkg/autoscaler/autoscaler.go
```diff
@@ -334,21 +360,9 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		By(fmt.Sprintf("Creating ClusterAutoscaler configured with maxNodesTotal:%v", maxNodesTotal))
 		clusterAutoscaler := clusterAutoscalerResource(maxNodesTotal)
 		Expect(client.Create(context.TODO(), clusterAutoscaler)).Should(Succeed())
+		cleanupObjects[clusterAutoscaler.GetName()] = runtime.Object(clusterAutoscaler)
```
JoelSpeed: Is this cast to a `runtime.Object` necessary? `clusterAutoscaler` is a pointer so should implement the interface already

There are a number of occurrences of this through this file

pkg/autoscaler/autoscaler.go
JoelSpeed: Only needed to recover test failures within go routines as in the example below, don't think it's needed in this test suite, does it pass if you remove it?
```
go func() {
  defer GinkgoRecover()
  Expect(doSomething()).To(Succeed())
}()
```

pkg/autoscaler/autoscaler.go
```diff
@@ -204,25 +208,47 @@ func remaining(t time.Time) time.Duration {
...
+		// 70% - enough that the existing and new nodes will
+		// be used, not enough to have more than 1 pod per
+		// node.
+		workloadMemRequest = resource.MustParse(fmt.Sprintf("%v", 0.7*float32(bytes)))
+	})
+
 	cascadeDelete := metav1.DeletePropagationForeground
+	deleteObject := func(name string, obj runtime.Object) error {
+		glog.Infof("[cleanup] %q (%T)", name, obj)
+		return client.Delete(context.TODO(), obj, &runtimeclient.DeleteOptions{
```
JoelSpeed: Can we either replace TODO with Background or declare `var ctx = client.Background()` on line 215 and use that everywhere? Nothing should be modifying the `ctx` variable so I think this is safe and this is a pattern I've seen used in a lot of test suites within kubernetes-sigs etc

pkg/autoscaler/autoscaler.go
```diff
@@ -434,4 +449,72 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+		By(fmt.Sprintf("Creating scale-out workload: jobs: %v, memory: %s", expectedReplicas, workloadMemRequest.String()))
+		workload := newWorkLoad(expectedReplicas, workloadMemRequest, targetedNodeLabel)
+		cleanupObjects[workload.GetName()] = workload
+		Expect(client.Create(ctx, workload)).Should(Succeed())
+
+		Eventually(func() bool {
+			ms, err := framework.GetMachineSet(client, machineSet.GetName())
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("Waiting for machineSet replicas to scale out. Current replicas are %v, expected %v.",
```
elmiko: to keep things consistent, could we say "scale up" here?

and also change the other occurrence to "scale down" instead of "scale in"

pkg/autoscaler/autoscaler.go
```diff
@@ -434,4 +449,72 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+				if err := deleteObject(name, obj); err != nil {
+					glog.Infof("[cleanup] error deleting object %q (%T): %v", name, obj, err)
+				}
+			}
+		}()
+
+		By("Creating a new MachineSet with 0 replicas")
+		machineSetParams := framework.BuildMachineSetParams(client, 0)
+		targetedNodeLabel := fmt.Sprintf("%v-scale-from-zero", autoscalerWorkerNodeRoleLabel)
+		machineSetParams.Labels[targetedNodeLabel] = ""
```
elmiko: do we need to propagate this label to machineSet.Spec.Template.Spec.Label as well?

enxebre: that's actually what this is doing, https://github.com/openshift/cluster-api-actuator-pkg/pull/140/files#diff-c7d1003b84db4b5057508a81ab203c46L66

elmiko: that's what i thought when i looked through the code. i was confused because i had _thought_ those labels would be copied by the autoscaler.

elmiko:
i am having issues pushing commits on top of this, as per my conversation with @enxebre i am going to open a new PR. unfortunately, it seems like i cannot close issues on this repo, otherwise i would close this one as well.

PR #141, opened 2020-03-18
Verify Expect(err).ToNot(HaveOccurred()) when scaling machineSets

enxebre:


pkg/infra/infra.go
```diff
@@ -280,8 +280,13 @@ var _ = Describe("[Feature:Machines] Managed cluster should", func() {
...
 		framework.WaitForMachineSet(client, machineSet2.GetName())
-		framework.ScaleMachineSet(machineSet.GetName(), 0)
-		framework.ScaleMachineSet(machineSet2.GetName(), 3)
+
+		err = framework.ScaleMachineSet(machineSet.GetName(), 0)
+		Expect(err).ToNot(HaveOccurred())
+
+		err = framework.ScaleMachineSet(machineSet2.GetName(), 3)
+		Expect(err).ToNot(HaveOccurred())
+
```
JoelSpeed: Might be better to inline these?
```suggestion
		Expect(framework.ScaleMachineSet(machineSet.GetName(), 0)).To(Succeed())
		Expect(framework.ScaleMachineSet(machineSet2.GetName(), 3)).To(Succeed())
```

PR #142, opened 2020-03-25
update readme with new run information

elmiko:
The change removes the build instructions and provides updated
instructions for running the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,32 +61,60 @@ $ make build
 $ ./bin/cluster-autoscaler-operator --kubeconfig=<PATH/TO/YOUR/CLUSTERS/KUBECONFIG> ...
 ```
 
-### Build the e2e tests
-
-```console
-$ make build-e2e 
-go test -c -o bin/e2e github.com/openshift/cluster-api-actuator-pkg/pkg/e2e
-```
-
 ### Run the autoscaler e2e tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,32 +61,60 @@ $ make build
...
-[Feature:Machines][Serial] Autoscaler should 
-  scale up and down
-  /home/aim/go-projects/cluster-api-actuator-pkg/src/github.com/openshift/cluster-api-actuator-pkg/pkg/e2e/autoscaler/autoscaler.go:229
-•SSSSSSSSSSSSSS
-Ran 1 of 15 Specs in 0.000 seconds
-SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 14 Skipped
---- PASS: TestE2E (0.00s)
-PASS
-ok      github.com/openshift/cluster-api-actuator-pkg/pkg/e2e   0.037s
+$ ./hack/ci-integration.sh -focus "Autoscaler should" -v -dryRun
```
JoelSpeed, line 75: Is this real output? Didn't think it ran parallel by default?

elmiko: it used to, but i took it out in #148, see [this commit](https://github.com/openshift/cluster-api-actuator-pkg/commit/ba0bf725211685eec27100d740da468cc7b0f2d7).

i probably created this before that change happened.

PR #144, opened 2020-03-30
[OCPCLOUD-789] Allow E2E suite to output JUnit files

JoelSpeed:
This enables packages using this test suite to output JUnit reports for consumption by Prow, this will make it easier to interpret test failures within these tests

Adding `JUNIT_DIR=$ARTIFACTS` when running the test suite should result in this being picked up by Prow


enxebre:
I you set the variable on the release repo already for this repository we can validate how the output will look like directly in this PR. Make sense?

JoelSpeed:
/test all

E2E suites should now output JUnit

PR #148, opened 2020-04-09
Bug 1827657: Validate scale from/to zero

elmiko:
This is a continuation of #140, it adds tests for the autoscaler to exercise scaling from and to zero nodes.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -450,7 +450,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		}
 	})
 
-	It("It scales from zero", func() {
+	It("It scales from/to zero", func() {
 
 		// Only run in platforms which support autoscaling from/to zero.
 		clusterInfra, err := framework.GetInfrastructure(client)
@@ -515,6 +515,18 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		By("Waiting for the machineSet replicas to become nodes")
 		framework.WaitForMachineSet(client, machineSet.GetName())
 
-		//TODO: Remove workload, drain nodes and validate scale to zero.
+		expectedReplicas = 0
+		By("Deleting the workload")
+		Expect(deleteObject(workload.Name, cleanupObjects[workload.Name])).Should(Succeed())
+		delete(cleanupObjects, workload.Name)
+		Eventually(func() bool {
+			ms, err := framework.GetMachineSet(client, machineSet.GetName())
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("Waiting for machineSet replicas to scale in. Current replicas are %v, expected %v.",
+				*ms.Spec.Replicas, expectedReplicas))
+
+			return *ms.Spec.Replicas == expectedReplicas
+		}, framework.WaitLong, pollingInterval).Should(BeTrue())
 	})
 })
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -212,6 +212,16 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 	var workloadMemRequest resource.Quantity
 	var client runtimeclient.Client
 	var err error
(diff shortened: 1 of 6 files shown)


elmiko:
i have added a commit that removes the parallel execution call for the autscaler tests. this is currently causing them to fail as both of the scaling tests need an autoscaler with different numbers of max nodes.

in the future these tests should be improved to allow the parallel flag by ensuring that the sensitive tests can either use the same autoscaler or they can coordinate their individual usages.

elmiko:
it appears that removing the parallelism is not having the same effect in the CI test suite as they do when running locally. i have added another commit to explicitly delete the ClusterAutoscaler objects.

pkg/autoscaler/autoscaler.go
```diff
@@ -204,25 +208,48 @@ func remaining(t time.Time) time.Duration {
...
+
+		// 70% - enough that the existing and new nodes will
+		// be used, not enough to have more than 1 pod per
+		// node.
+		workloadMemRequest = resource.MustParse(fmt.Sprintf("%v", 0.7*float32(bytes)))
+	})
+
+	ctx := context.Background()
 	cascadeDelete := metav1.DeletePropagationForeground
+	deleteObject := func(name string, obj runtime.Object) error {
```
enxebre, line 240: can't we add and `afterEach` clause which runs the logic in https://github.com/openshift/cluster-api-actuator-pkg/pull/148/files#diff-a6c4792dc41958d712d5dc9615145d28R484-R488 so we clean up consistently every validation?

enxebre:
Thanks! other than https://github.com/openshift/cluster-api-actuator-pkg/pull/148/files#r407429615 lgtm.
Since we are controlling scaling events with targeted machineSets per validation and node selectors I can't see any impediment for this to run in parallel along the whole suite. That better to enable in a follow up PR though.

pkg/autoscaler/autoscaler.go
elmiko: yeah, i think we definitely can. i'll add it

elmiko:
> Since we are controlling scaling events with targeted machineSets per validation and node selectors I can't see any impediment for this to run in parallel along the whole suite. That better to enable in a follow up PR though.

i agree that we could probably run the autoscaler tests in parallel with the other tests, but 2 of the autoscaler need to be run serially with each other. follow up pr makes perfect sense to me.

pkg/autoscaler/autoscaler.go
elmiko: i'm looking into this deeper and i think it will require some restructuring to have a common cleanup routine.

pkg/framework/clusterautoscalers.go
```diff
@@ -0,0 +1,21 @@
+package framework
+
+import (
+	"context"
+	"fmt"
+
+	caov1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1"
```
JoelSpeed, line 7: nit: I would have expected this to be imported as `autoscalingv1` (normally it's the groups shortname followed by a version), though that may clash with the upstream autoscaling APIs 🤔

elmiko: thanks, that makes sense to me. iirc i copied this import from another piece of code. but for future reference i will try to make these import more idiomatic.

elmiko:
@JoelSpeed @enxebre ptal, given our conversation today i am happy to merge this pr as-is. i have some ideas for cleaning up these tests but i think it's more important that we have a solid e2e for the autoscaler now.

JoelSpeed:
/lgtm

Defer to @enxebre to approve

JoelSpeed:
@elmiko Should we squash this up a bit before merge?

elmiko:
> @elmiko Should we squash this up a bit before merge?

i don't have a problem squashing if folks want, i figured since this is purely a test repo it wouldn't be so horrible if there are extra commits.

enxebre:
thanks @JoelSpeed @elmiko!
I agree commits could be squashed a bit, though history in the PR looks good enough to me to provide good context and tell the story for anyone looking at it.
Let's ship it and move on!
/hold cancel

PR #150, opened 2020-05-05
add a test for deletion annotations after scale down

elmiko:
This change adds a test to the autoscale "scale up and down" suite that
will check to see if there are any deletion annotations left on the
machine resources after a scale down. To accomplish this some logic has
been added to make the test wait until the scale down has completed so
that we can check the remaining machines.

This test is related to bugzilla case 1820410. reference:
https://bugzilla.redhat.com/show_bug.cgi?id=1820410

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -2,6 +2,7 @@ package autoscaler
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -35,6 +37,7 @@ const (
 	pollingInterval                       = 3 * time.Second
 	autoscalerWorkerNodeRoleLabel         = "machine.openshift.io/autoscaler-e2e-worker"
 	workloadJobName                       = "e2e-autoscaler-workload"
+	machineDeleteAnnotationKey            = "machine.openshift.io/cluster-api-delete-machine"
 )
 
 func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string) *batchv1.Job {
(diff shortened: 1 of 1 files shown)


elmiko:
/hold

i think this test needs a little further examination, i have seen some test failures that i am having trouble tracking down.

elmiko:
/hold cancel

i've fixed this test to take into account the time involved with resolving the deletion taints from the remaining machines. we expect the annotations will be removed but it is difficult to guarantee /when/ they will be removed. this test will now wait for a short period of time to give the machines extra time to reflect the annotation changes.

elmiko:
adjusted the wait time for the annotations from short to medium. although this test passes locally, i am seeing error in the ci jobs. i have a feeling relaxing the timing here will help.

elmiko:
just adding a little more info to the log statements

pkg/autoscaler/autoscaler.go
```diff
@@ -429,15 +436,62 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			for _, tms := range transientMachineSets {
+				machines, err := framework.GetMachinesFromMachineSet(client, &tms)
+				Expect(err).NotTo(HaveOccurred())
+				for _, m := range machines {
+					if m.ObjectMeta.Annotations == nil {
+						continue
+					}
+					if _, exists := m.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
+						glog.Infof("Found deletion annotation on machine %q", m.Name)
+						annotationsfound++
```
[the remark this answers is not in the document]
elmiko: i had originally coded it that way, but i wanted to add more information for debugging. basically in this loop it's possible to have multiple Machine resources with the annotation still attached. i wanted to record them all.

pkg/autoscaler/autoscaler.go
```diff
@@ -326,11 +332,12 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		var machineAutoscalers []*caov1beta1.MachineAutoscaler
 
 		By(fmt.Sprintf("Creating %v machineautoscalers", len(machineSets)))
+		transientMachineSetsMinSize := int32(1)
 		var clusterExpansionSize int
 		for i := range machineSets {
 			clusterExpansionSize++
-			glog.Infof("Create MachineAutoscaler backed by MachineSet %s/%s - min:%v, max:%v", machineSets[i].Namespace, machineSets[i].Name, 1, 2)
-			asr := machineAutoscalerResource(machineSets[i], 1, 2)
+			glog.Infof("Create MachineAutoscaler backed by MachineSet %s/%s - min:%v, max:%v", machineSets[i].Namespace, machineSets[i].Name, transientMachineSetsMinSize, 2)
```
JoelSpeed, line 339: If this were a `By`, it would auto be slurped by the GinkgoWriter, so it would only display for a particular test if it failed, WDYT?

pkg/autoscaler/autoscaler.go
```diff
@@ -429,15 +436,62 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 			return true
 		}, framework.WaitMedium, pollingInterval).Should(BeZero())
 
+		// Wait until the MachineSets have reach minimum size so that
```
JoelSpeed, line 439:
```suggestion
		// Wait until the MachineSets reach the minimum size so that
```

pkg/autoscaler/autoscaler.go
```diff
@@ -429,15 +436,62 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+		// we can test to ensure that the remaining Machines and Nodes
+		// do not contain deletion information.
+		By("Waiting for transient machinesets to reach minimum size")
+		transientMachineSetSelector := &metav1.LabelSelector{
+			MatchLabels: map[string]string{machinesetTestLabel: ""},
+		}
+		Eventually(func() bool {
+			transientMachineSets, err := framework.GetMachineSets(client, transientMachineSetSelector)
+			Expect(err).NotTo(HaveOccurred())
```
JoelSpeed, line 448: This would be better like this, so that we try the GetMachineSets over and over if it fails, until the timeout. Ginkgo will assert any parameters other than the first are nil when performing the `Should`, so this is safe to do
```suggestion
		Eventually(func() (bool, error) {
			transientMachineSets, err := framework.GetMachineSets(client, transientMachineSetSelector)
			if err != nil {
			    return false, err
			}
```

pkg/autoscaler/autoscaler.go
```diff
@@ -429,15 +436,62 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+		}, framework.WaitLong, pollingInterval).Should(BeTrue())
+
+		By("Checking to ensure that no deletion annotations exist on machines")
+		transientMachineSets, err := framework.GetMachineSets(client, transientMachineSetSelector)
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(func() bool {
+			annotationsfound := 0
+			for _, tms := range transientMachineSets {
+				machines, err := framework.GetMachinesFromMachineSet(client, &tms)
+				Expect(err).NotTo(HaveOccurred())
```
JoelSpeed, line 466: Similar to above, could return the error to get retries

pkg/autoscaler/autoscaler.go
elmiko: that makes sense to me. i copied this logic from other places in the code and i figured if we created an error on these calls something is seriously wrong, but i can see the wisdom in just having it retry. i'll update.

pkg/autoscaler/autoscaler.go
elmiko: ++

pkg/autoscaler/autoscaler.go
elmiko: let's talk about this, part of the reason i backed off on using `By` for these inner loop logs is that when they are printed in the logs it looks like the test is starting a new "step" everytime it does this. is there something similar to `By` that i can use more like a log emit?

elmiko:
updated based on comments. as discussed offline, we will push on the `By` changes in favor of improving our log usage(in a different PR).

JoelSpeed:
/lgtm

Needs a BZ if we want this to merge now

elmiko:
> Needs a BZ if we want this to merge now

we do have a bz for this, but i was under the impression that we don't need them for this repo.

JoelSpeed:
Tide status seems to me to suggest otherwise 🤔

elmiko:
i had _thought_ we didn't need them per our discussion offline, but looking at another recent pr i can see we do. i'll add it.

JoelSpeed:
@elmiko Changes look good to me, just wanted to double check you have tested covers the bug it was cloned from, as in, this fails if that fix is removed?

elmiko:
> @elmiko Changes look good to me, just wanted to double check you have tested covers the bug it was cloned from, as in, this fails if that fix is removed?

i tested this manually on the build up of this change(eg confirming the old code failed), but i don't think i ever went back and ran a pre-change version of the autoscaler against this test. we also had confirmation of the fix through qe.

elmiko:
i'm seeing some weird behavior running this test on a 4.4 cluster. i'm going to put a hold for now, i think i might need to adjust it slightly.

/hold

elmiko:
@JoelSpeed did a massive refactor and wrote a new test for deletion stuff. this one absolutely fails where it is supposed to, i still need to check that it works where it's supposed to but i wanted to share this to get your thoughts.

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +255,412 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			Eventually(func() (bool, error) {
+				sizeCount := 0
+				for _, machineSet := range transientMachineSets {
+					current, err := framework.GetMachineSet(client, machineSet.GetName())
+					if err != nil {
+						return false, err
+					}
+					if *current.Spec.Replicas == expectedReplicas {
+						sizeCount++
+					}
```
JoelSpeed, line 605: Could invert this equality and return early, then you don't need `sizeCount`
```suggestion
					if *current.Spec.Replicas != expectedReplicas {
						return false, nil
					}
```

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +255,412 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
```
JoelSpeed, line 631: Same comment as above

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +255,412 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+					if err != nil {
+						return false, err
+					}
+					for _, machine := range machines {
+						if machine.ObjectMeta.Annotations == nil {
+							continue
+						}
+						if _, exists := machine.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
+							annotationsfound++
+						}
```
JoelSpeed, line 656: Again, not sure `annotationsfound` is needed, we can just return early if we find one that doesn't meet the condition we are looking for 
```suggestion
						if _, exists := machine.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
							return false, nil
						}
```

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +255,412 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-			framework.WaitForMachineSetDelete(client, ms)
-		}
+			// explicitly delete the ClusterAutoscaler
+			// this is needed due to the autoscaler tests requiring singleton
+			// deployments of the ClusterAutoscaler.
+			caName := clusterAutoscaler.GetName()
+			Expect(deleteObject(caName, cleanupObjects[caName])).Should(Succeed())
+			delete(cleanupObjects, caName)
+			Eventually(func() bool {
+				By(fmt.Sprintf("Waiting for ClusterAutoscaler to delete"))
```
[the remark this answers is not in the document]
elmiko: oops, good catch. i saw those and wondered how they lived so long XD

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +255,412 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-			if ca, err := framework.GetClusterAutoscaler(client, caName); ca == nil && err != nil {
-				return true
+			caName := clusterAutoscaler.GetName()
+			Expect(deleteObject(caName, cleanupObjects[caName])).Should(Succeed())
+			delete(cleanupObjects, caName)
+			Eventually(func() bool {
+				if ca, err := framework.GetClusterAutoscaler(client, caName); ca == nil && err != nil {
+					return true
+				}
+				return false
```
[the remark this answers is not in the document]
elmiko: i had tried the `IsNotFound` in the past and gave me some unexpected results, i can certainly add it back here.

pkg/autoscaler/autoscaler.go
elmiko: good spot, this is a leftover from when i had the logs printing inside the loop. i'll remove it.

pkg/autoscaler/autoscaler.go
elmiko: yup, more leftovers from refactor

elmiko:
updated with requested changes

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			// The remainder of the logic in this test requires 3
+			// machinesets.
+			var machineSets [3]*mapiv1beta1.MachineSet
+
+			randomUUID := string(uuid.NewUUID())
 			for i := 0; i < len(machineSets); i++ {
-				nodes, err := framework.GetNodesFromMachineSet(client, machineSets[i])
-				if err != nil {
-					return false
+				targetMachineSet := existingMachineSets[i%len(existingMachineSets)]
```
JoelSpeed, line 291: What's the benefit of the module operation on the `existingMachineSets`? Would it make more sense to do `for i, targetMachineSet := range machineSets` to get `i` and `targetMachineSet`?

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			By("Getting nodes")
+			nodes, err := framework.GetNodes(client)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(nodes)).To(BeNumerically(">=", 1))
+
+			var machineAutoscalers []*caov1beta1.MachineAutoscaler
+
+			By(fmt.Sprintf("Creating %v machineautoscalers", len(machineSets)))
+			transientMachineSetsMinSize := int32(1)
+			var clusterExpansionSize int
```
JoelSpeed, line 334: This seems to only be used to check the length of `machineSets` indirectly, is that required? Could we not just assert `machineSets` has a length before doing the range operation?

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+				v := scaleDownCounter.get()
+				glog.Infof("[%s remaining] Expecting %v %q events; observed %v",
+					remaining(testDuration), clusterExpansionSize-1, clusterAutoscalerScaleDownEmpty, v)
+				return v == uint32(clusterExpansionSize-1)
+			}, framework.WaitLong, pollingInterval).Should(BeTrue())
+
+			Eventually(func() bool {
+				podList := corev1.PodList{}
+				err = client.List(ctx, &podList, runtimeclient.InNamespace(workload.Namespace))
+				Expect(err).NotTo(HaveOccurred())
```
JoelSpeed, line 427: Calling `Expect` within an `Eventually` will bail if the whole test if there's an error, is that intentional or would we want to retry?

pkg/autoscaler/autoscaler.go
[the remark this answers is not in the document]
JoelSpeed: Not resolved

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-		}
+			// explicitly delete the ClusterAutoscaler
+			// this is needed due to the autoscaler tests requiring singleton
+			// deployments of the ClusterAutoscaler.
+			caName := clusterAutoscaler.GetName()
+			Expect(deleteObject(caName, cleanupObjects[caName])).Should(Succeed())
+			delete(cleanupObjects, caName)
+			Eventually(func() bool {
+				By(fmt.Sprintf("Waiting for ClusterAutoscaler to delete"))
+				if ca, err := framework.GetClusterAutoscaler(client, caName); ca == nil && err != nil {
```
JoelSpeed, line 463: I'd prefer to use the `IsNotFound`, I think I commented that before but it seems to have gone 🤔 Did you say you had issues with it?

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			workload := newWorkLoad(expectedReplicas, workloadMemRequest, targetedNodeLabel)
+			cleanupObjects[workload.GetName()] = workload
+			Expect(client.Create(ctx, workload)).Should(Succeed())
 
-		machineSet, err := framework.CreateMachineSet(client, machineSetParams)
-		Expect(err).ToNot(HaveOccurred())
-		cleanupObjects[machineSet.GetName()] = machineSet
+			Eventually(func() bool {
+				ms, err := framework.GetMachineSet(client, machineSet.GetName())
+				Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 535: Same comment as above re `Expect`

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +256,406 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-		workload := newWorkLoad(expectedReplicas, workloadMemRequest, targetedNodeLabel)
-		cleanupObjects[workload.GetName()] = workload
-		Expect(client.Create(ctx, workload)).Should(Succeed())
+			expectedReplicas = 0
+			By("Deleting the workload")
+			Expect(deleteObject(workload.Name, cleanupObjects[workload.Name])).Should(Succeed())
+			delete(cleanupObjects, workload.Name)
+			Eventually(func() bool {
+				ms, err := framework.GetMachineSet(client, machineSet.GetName())
+				Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 552: Same comment as above re `Expect`

elmiko:
@JoelSpeed so, i apologize, this review got a little wonky. i totally agree with the comments you are making, but the areas you are commenting on are actually the original tests. i put them back the way they were originally and just added my new test at the bottom.

i /so/ want to go back and refactor those other 2 tests, as you point out there is room for improvement, but for this pr i wanted to not modify the older tests.

JoelSpeed:
> @JoelSpeed so, i apologize, this review got a little wonky. i totally agree with the comments you are making, but the areas you are commenting on are actually the original tests. i put them back the way they were originally and just added my new test at the bottom.
> 
> i /so/ want to go back and refactor those other 2 tests, as you point out there is room for improvement, but for this pr i wanted to not modify the older tests.

Ack, I just reviewed with whitespace hidden, this is fine when that's considered! 

/lgtm

elmiko:
updated with some changes to the cluster autoscaler detection, the watch for transient machine scale down, and the context messages.

@JoelSpeed i looked into the error situation by printing out `apierrors.ReasonForError(err)` in those faliure states. it looks like the client is returning an error that the errors package doesn't recognize. also, after looking at the docs i tried out `apierrors.IsGone` (which seemed more appropriate) but obviously errors doesn't understand what's coming back.

JoelSpeed:
I think I can see why it's not returning an error that is the right format. When we get the error back we `fmt.Errorf` it, but we are using `%v` instead of `%w`, so the original error is lost intead of being wrapped. If we swap the format characters over then the error coming out of `GetClusterAutoscaler` would be would with `IsNotFound` as it supports errors wrapped in this way
https://github.com/openshift/cluster-api-actuator-pkg/blob/0f66bbe659db881cc781b355a76c1c3e0ace809c/pkg/framework/clusterautoscalers.go#L17

https://github.com/kubernetes/apimachinery/blob/bdcc9f4ab675d837965d1bd5babf8299afead78a/pkg/api/errors/errors.go#L650-L652

Apologies for pushing on this one, but at the moment, the check (`ca ==nil && err != nil`) does not mean that we only return true on deletion, this could happen because the credentials are bad, or the API request timed out, or there was a conflict response. So the test at the moment doesn't check the condition we need and _could_ pass even when it's not supposed to.

elmiko:
> Apologies for pushing on this one, but at the moment, the check (`ca ==nil && err != nil`) does not mean that we only return true on deletion, this could happen because the credentials are bad, or the API request timed out, or there was a conflict response. So the test at the moment doesn't check the condition we need and _could_ pass even when it's not supposed to.

no apologies necessary. i'm all for fixing this as i've run into the issue twice without having a decent explanation for why it was happening. i'll revert the changes, although i think `IsGone` is probably what we want here instead of `IsNotFound`, but perhaps i'm not reading the comment correctly.

elmiko:
updated to include a more verbose error check for the cluster autoscaler delete, using the `IsNotFound` method.

thanks to @JoelSpeed for the solution.

elmiko:
it seems that all the pending PRs have merged and this issues here are fixed.
/hold cancel

enxebre:
@elmiko this ended up somehow pointing to a verified bz https://github.com/openshift/cluster-api-actuator-pkg/pull/150. Let’s remove it and merge it as feature 4.6 dev.

enxebre:
/retitle add a test for deletion annotations after scale down

elmiko:
> @elmiko this ended up somehow pointing to a verified bz #150. Let’s remove it and merge it as feature 4.6 dev.

makes sense to me. i created the [second bz](https://bugzilla.redhat.com/show_bug.cgi?id=1835873) because the first had be open so long it got verified/fixed on the qe side. so i made that second bug (1835873) to track the test issue.

i'm fine with pushing this to 4.6, should i close out the second bug i made or can we still link it to this merge request?

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +257,409 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			// machinesets.
+			var machineSets [3]*mapiv1beta1.MachineSet
+
+			randomUUID := string(uuid.NewUUID())
 			for i := 0; i < len(machineSets); i++ {
-				nodes, err := framework.GetNodesFromMachineSet(client, machineSets[i])
-				if err != nil {
-					return false
+				targetMachineSet := existingMachineSets[i%len(existingMachineSets)]
+				machineSetName := fmt.Sprintf("e2e-%s-w-%d", randomUUID[:5], i)
```
enxebre, line 293: As a follow up PR we should consolidate and:
reuse `BuildMachineSetParams` which will generate the name using the clusterID for you
reuse.
Use only `CreateMachineSet`. Drop `NewMachineSet`

enxebre:
/retest
lgtm but defering label as I would like to have another look before shipping  as change is quite big.

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +257,409 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-			framework.WaitForMachineSetDelete(client, ms)
-		}
+			// explicitly delete the ClusterAutoscaler
+			// this is needed due to the autoscaler tests requiring singleton
+			// deployments of the ClusterAutoscaler.
+			caName := clusterAutoscaler.GetName()
+			Expect(deleteObject(caName, cleanupObjects[caName])).Should(Succeed())
+			delete(cleanupObjects, caName)
+			Eventually(func() bool {
+				By(fmt.Sprintf("Waiting for ClusterAutoscaler to delete"))
```
JoelSpeed, line 463: Since this line is changed already, may as well remove the redundant fmt
```suggestion
				By("Waiting for ClusterAutoscaler to delete")
```

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +257,409 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+							continue
+						}
+						if _, exists := machine.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
+							return false, nil
+						}
+					}
+				}
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue())
+		})
```
JoelSpeed, line 663: This seems a bit odd to me, we iterate through the list once to print that we are checking for the annotation, then iterate throught the list again for actually checking, why not have the eventually immediately after the by and just have it check for the one machine having the annotation?

elmiko: it does look a little odd, i agree. my reasoning here was to print the names of the machines we are going to be watching for removal of annotations. but, i didn't want to print continuously through the `Eventually` clause (since it can take a few loops to finish).

i can either refactor this or add some comments to help explain, wdyt?

elmiko:
updated with fmt.Sprintf change suggested

pkg/autoscaler/autoscaler.go
JoelSpeed: Does this sort of thing not work?
```suggestion
			for _, machineSet := range transientMachineSets {
				machines, err := framework.GetMachinesFromMachineSet(client, machineSet)
				Expect(err).NotTo(HaveOccurred())
				for _, machine := range machines {
					By(fmt.Sprintf("Checking Machine %s for %s annotation", machine.Name, machineDeleteAnnotationKey))
					Eventually(func() (bool, error) {
					    m := &Machine{}
					    err := client.Get(ctx, client.ObjectKey{Namespace: machine.Namespace, Name: machine.Name}, m)
					    if err != nil {
					      return false, err
					    }
					    
					    if m.ObjectMeta.Annotations == nil {
					      return true, nil
					    }
					    if _, exists := machine.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
				              return false, nil
					    }
					    return true, nil
					})
				}
			}
```

elmiko: this makes sense to me, let me re-work this bit a little

elmiko:
refactored the last clause of the test per Joel's suggestion

elmiko:
improved the way the machineset scale down replicas are counted. it appeared that when the mechanism to count annotations was improved it exposed a timing issue between checking the MachineSet.Status.Replicas and iterating the actual Machines from an API call.

now the scale down will wait until the desired length is reached for the list of machines, then the annotations are checked.

elmiko:
added the cluster autoscaler event watcher into these tests. this is borrowed from the previous scale up/down test but it is added in a way to be automatic for tests in the same `Context` block.

i'm adding this because it has been invaluable in debugging issues with test failures.

elmiko:
rebase to account for other changes.

PR #151, opened 2020-05-05
add a test for deletion taints after scale down

elmiko:
This change adds a test to the autoscale "scale up and down" suite that
will check to see if there are any deletion taints left on the node
resources after a scale down.
    
This test is related to bugzilla case 1823667 . reference:
https://bugzilla.redhat.com/show_bug.cgi?id=1823667

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -38,6 +38,8 @@ const (
 	autoscalerWorkerNodeRoleLabel         = "machine.openshift.io/autoscaler-e2e-worker"
 	workloadJobName                       = "e2e-autoscaler-workload"
 	machineDeleteAnnotationKey            = "machine.openshift.io/cluster-api-delete-machine"
+	deletionCandidateTaintKey             = "DeletionCandidateOfClusterAutoscaler"
+	toBeDeletedTaintKey                   = "ToBeDeletedByClusterAutoscaler"
 )
 
 func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string) *batchv1.Job {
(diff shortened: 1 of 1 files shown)


elmiko:
/hold

this will require #150 to be merged, as well as https://github.com/openshift/kubernetes-autoscaler/pull/149

elmiko:
the dependent changes have merged, cancelling the hold
/hold cancel

elmiko:
scratch that, this needs #150 to merge first.
/hold

pkg/autoscaler/autoscaler.go
```diff
@@ -254,295 +259,444 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-			By(fmt.Sprintf("[%s remaining] Waiting for nodes to be Ready in %v transient machinesets",
-				remaining(testDuration), len(machineSets)))
-			var allNewNodes []*corev1.Node
+			By("Getting existing nodes")
+			existingNodes, err := framework.GetNodes(client)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(existingNodes)).To(BeNumerically(">=", 1))
+
+			klog.Infof("Have %v existing machinesets", len(existingMachineSets))
+			klog.Infof("Have %v existing machines", len(existingMachines))
```
enxebre, line 284: is this dropped by https://github.com/openshift/cluster-api-actuator-pkg/pull/156?
If so let's hold this one, get 156 first and the rebase here

elmiko: yes, they will be removed. i'm fine placing a hold on this until #156 merges

elmiko:
/hold
/retest

holding until #156 merges

JoelSpeed:
In this test, do we expect that there are machines that would have had these taints applied, that would not have been removed from the system? Do we have a test where a scale down fails and the CA needs to remove the taints afterwards?

elmiko:
> In this test, do we expect that there are machines that would have had these taints applied, that would not have been removed from the system? Do we have a test where a scale down fails and the CA needs to remove the taints afterwards?

yes, these tests assume that all the taints have been removed by the CA. we don't have a good path to test scaling failures (in any of these tests).

functionally, this should be exercising the same pathways that would occur during a failed scale up event. the CA has a few deferred, or catch style routines, that will remove these taints. one of the error states we saw came from a failed API update based on the stale replica information.

i think it would be interesting to create a series of tests that explicitly work for conditions that occur from failed scaling events. i think we would need to create those tests explicitly.

JoelSpeed:
Ack, SGTM

/lgtm

JoelSpeed:
Will this need a rebase now the first commit is merged? I don't think GitHub will auto notice that

elmiko:
i hope not, it is a fast forward from the previous commit

elmiko:
pushed a rebase

PR #155, opened 2020-05-13
Replace Glog with Klog

### LINKED ISSUE #154: Switch glog for klog
[Klog](https://github.com/kubernetes/klog) is a drop in replacement for glog which we use in various places throughout the test packages. However, klog has an advantage over glog. We can use `klog.SetOutput(GinkgoWrite)` to redirect log lines to the Ginkgo system. This means that if a particular test passes, the logs from that test are not shown, and if the test fails, only the logs from the failed test are shown.

This will help debugging failed tests as it will remove spurious logs from the test output.

JoelSpeed:
Fixes #154 

Basically a find and replace, plus setting `klog.SetOutput(GinkgoWriter)` in the suite init


elmiko:
i tried this out with some tests and everything looks good on my end.

PR #156, opened 2020-05-13
remove existingMachines and existingNodes variables

elmiko:
The Autoscaler "scale up and down" test used these variables to record
information but these usages do not have direct bearing on the test.
Additionally the initial check of "nodes == machines" does not test the
scaling behavior and it will cause issues when attempting to parallelize
these tests.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -262,21 +262,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		existingMachineSets, err := framework.GetMachineSets(client)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(len(existingMachineSets)).To(BeNumerically(">=", 1))
-
-		By("Getting existing machines")
-		existingMachines, err := framework.GetMachines(client)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(len(existingMachines)).To(BeNumerically(">=", 1))
-
-		By("Getting existing nodes")
-		existingNodes, err := framework.GetNodes(client)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(len(existingNodes)).To(BeNumerically(">=", 1))
-
 		klog.Infof("Have %v existing machinesets", len(existingMachineSets))
-		klog.Infof("Have %v existing machines", len(existingMachines))
-		klog.Infof("Have %v existing nodes", len(existingNodes))
-		Expect(len(existingNodes) == len(existingMachines)).To(BeTrue())
 
 		// The remainder of the logic in this test requires 3
 		// machinesets.
@@ -322,6 +308,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		nodes, err := framework.GetNodes(client)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(len(nodes)).To(BeNumerically(">=", 1))
+		klog.Infof("Found %q nodes", len(nodes))
 
 		var machineAutoscalers []*caov1beta1.MachineAutoscaler
 
@@ -338,7 +325,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		Expect(clusterExpansionSize).To(BeNumerically(">", 1))
 
 		// The total size of our cluster is
-		// len(existingMachineSets) + clusterExpansionSize. We
+		// len(nodes) + clusterExpansionSize. We
 		// cap that to $max-1 because we want to test that the
 		// maxNodesTotal flag is respected by the
 		// cluster-autoscaler
@@ -359,8 +346,6 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		clusterAutoscaler := clusterAutoscalerResource(maxNodesTotal)
 		Expect(client.Create(ctx, clusterAutoscaler)).Should(Succeed())
 		cleanupObjects[clusterAutoscaler.GetName()] = clusterAutoscaler
-
-		By(fmt.Sprintf("Deriving Memory capacity from machine %q", existingMachineSets[0].Name))
 
 		By(fmt.Sprintf("Creating scale-out workload: jobs: %v, memory: %s", maxNodesTotal+1, workloadMemRequest.String()))
 		scaledGroups := map[string]bool{}


elmiko:
looking over this test, these variables seem out of place and i don't think they add to the test. i would like to factor out `existingMachineSets` as well, but i think it will require further research. i think this is going to cause issues when we start to add parallel test execution as these checks could return incorrect info.

pkg/autoscaler/autoscaler.go
```diff
@@ -262,21 +262,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		existingMachineSets, err := framework.GetMachineSets(client)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(len(existingMachineSets)).To(BeNumerically(">=", 1))
-
-		By("Getting existing machines")
-		existingMachines, err := framework.GetMachines(client)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(len(existingMachines)).To(BeNumerically(">=", 1))
```
elmiko, line 269: this seems like a sanity check, just to make sure we have machines. but i don't think it is directly related to the scale up/down functionality. also, `existingMachines` doesn't seem to be used elsewhere.

pkg/autoscaler/autoscaler.go
```diff
@@ -262,21 +262,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-
-		By("Getting existing nodes")
-		existingNodes, err := framework.GetNodes(client)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(len(existingNodes)).To(BeNumerically(">=", 1))
```
elmiko, line 274: similar to above, this appears to be a sanity check but it is not directly about the scale up/down and the value is not used later.

pkg/autoscaler/autoscaler.go
```diff
@@ -262,21 +262,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
-
 		glog.Infof("Have %v existing machinesets", len(existingMachineSets))
-		glog.Infof("Have %v existing machines", len(existingMachines))
-		glog.Infof("Have %v existing nodes", len(existingNodes))
-		Expect(len(existingNodes) == len(existingMachines)).To(BeTrue())
```
elmiko, line 279: this appears to be another sanity check, additionally if we start to make the autoscaler tests parallel there could be flakes around the timing of queries for nodes and machines.

pkg/autoscaler/autoscaler.go
```diff
@@ -322,6 +308,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		nodes, err := framework.GetNodes(client)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(len(nodes)).To(BeNumerically(">=", 1))
+		glog.Infof("Found %q nodes", len(nodes))
```
elmiko, line 311: adding this comment to help additional context to the previous `By`.

pkg/autoscaler/autoscaler.go
```diff
@@ -338,7 +325,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		Expect(clusterExpansionSize).To(BeNumerically(">", 1))
 
 		// The total size of our cluster is
-		// len(existingMachineSets) + clusterExpansionSize. We
+		// len(nodes) + clusterExpansionSize. We
 		// cap that to $max-1 because we want to test that the
```
elmiko, line 329: cleaning up the comments to show the actual behavior. we aren't using `existingMachineSets` and instead query the number of nodes just before the following assignment.

pkg/autoscaler/autoscaler.go
```diff
@@ -360,8 +347,6 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		Expect(client.Create(ctx, clusterAutoscaler)).Should(Succeed())
 		cleanupObjects[clusterAutoscaler.GetName()] = clusterAutoscaler
 
-		By(fmt.Sprintf("Deriving Memory capacity from machine %q", existingMachineSets[0].Name))
-
```
elmiko, line 364: i think there used to be something here, but we are no longer using the `existingMachineSets` to derive the memory capacity.

elmiko:
adding some context on @michaelgugino 's suggestion.

i would like to make these changes to cleanup the code a little and remove things are directly related to the test. i think it will make code cleaner and easier for people to understand and hack on.

pkg/autoscaler/autoscaler.go
michaelgugino: Was this getting all machines from the cluster, or machines from a specific set?  It looks like it was all machines, in which case, we'd always have more than zero, so this can probably go.

pkg/autoscaler/autoscaler.go
michaelgugino: Same question.  Is this all nodes, or just a specific group of nodes?  Looks like it's probably ensuring we have at least one node, which I would assume is always true in our case.

pkg/autoscaler/autoscaler.go
michaelgugino: Having tests that are easier to reason about, especially if things go wrong, is a bigger priority for me than running tests in parallel.  Will removing these log lines make ascertaining what happened if the tests fail more difficult in the future in any way?

pkg/autoscaler/autoscaler.go
elmiko: yeah, all the machinesets from openshift-machine-api project

pkg/autoscaler/autoscaler.go
elmiko: yeah, all nodes. my thought was similar, this is just checking to see that we have at least 1 node, but i don't think it adds to this specific test. if we want to ensure that we have a single node before beginning then i would move it to the out context.

pkg/autoscaler/autoscaler.go
elmiko: i don't think removing these will make it more difficult. this test specifically creates 3 machinesets and then reports on the sizes of those machinesets as it operates. i don't think knowing the initial number of machines and nodes adds much, especially when we start to consider that other tests could affect these numbers which would make them even more difficult to interpret, imo.

elmiko:
rebased

PR #158, opened 2020-05-27
OCPCLOUD-867 - Add MAO e2e for validating the deployment is resynced

Danil-Grigorev:
Adds check for `deployment.spec` update after deletion.

### DIFF
--- a/pkg/framework/deployment.go
+++ b/pkg/framework/deployment.go
@@ -3,7 +3,7 @@ package framework
 import (
 	"context"
 	"fmt"
-	"time"
+	"reflect"
 
 	kappsapi "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -20,7 +20,7 @@ func GetDeployment(c client.Client, name, namespace string) (*kappsapi.Deploymen
 	}
 	d := &kappsapi.Deployment{}
 
-	if err := wait.PollImmediate(1*time.Second, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
 		if err := c.Get(context.TODO(), key, d); err != nil {
 			klog.Errorf("Error querying api for Deployment object %q: %v, retrying...", name, err)
 			return false, nil
@@ -34,7 +34,7 @@ func GetDeployment(c client.Client, name, namespace string) (*kappsapi.Deploymen
 
 // DeleteDeployment deletes the specified deployment
 func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
-	return wait.PollImmediate(1*time.Second, WaitShort, func() (bool, error) {
+	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
 		if err := c.Delete(context.TODO(), deployment); err != nil {
 			klog.Errorf("error querying api for deployment object %q: %v, retrying...", deployment.Name, err)
 			return false, nil
@@ -43,9 +43,25 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
 	})
 }
 
+// UpdateDeployment updates the specified deployment
+func UpdateDeployment(c client.Client, name, namespace string, updated *kappsapi.Deployment) error {
+	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace)
+		if err != nil {
+			klog.Errorf("Error getting deployment: %v", err)
+			return false, nil
+		}
+		if err := c.Patch(context.TODO(), d, client.MergeFrom(updated)); err != nil {
+			klog.Errorf("error patching deployment object %q: %v, retrying...", name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
 // IsDeploymentAvailable returns true if the deployment has one or more availabe replicas
 func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
-	if err := wait.PollImmediate(1*time.Second, WaitLong, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
 		d, err := GetDeployment(c, name, namespace)
 		if err != nil {
 			klog.Errorf("Error getting deployment: %v", err)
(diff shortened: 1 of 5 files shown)


pkg/operators/machine-api-operator.go
```diff
@@ -39,6 +39,9 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
 
 		By(fmt.Sprintf("checking deployment %q is available again", maoManagedDeployment))
 		Expect(framework.IsDeploymentAvailable(client, maoManagedDeployment, framework.MachineAPINamespace)).To(BeTrue())
+
```
enxebre, line 42: we should manipulate it, e.g update replica number then wait for it to be reconciled with the enforced spec e.g replica number=1

pkg/framework/deployment.go
```diff
@@ -66,6 +78,27 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
 	return true
 }
 
+// IsDeploymentSynced returns true if the deployment has one or more availabe replicas
```
elmiko, line 81: i think the comment is a little off here, the function looks like it's checking equality between the deployment specs not just replicas.

pkg/framework/deployment.go
```diff
@@ -66,6 +78,27 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
...
+func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment, name, namespace string) bool {
+	if err := wait.PollImmediate(1*time.Second, WaitLong, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace)
+		if err != nil {
+			glog.Errorf("Error getting deployment: %v", err)
+			return false, nil
+		}
+		if !reflect.DeepEqual(d.Spec, dep.Spec) {
+			glog.Infof("Deployment %q is updated. Spec is not equal to: %v",
+				d.Name, dep.Spec)
```
elmiko, line 91: should this be returning false here?

Danil-Grigorev: Good catch!

pkg/framework/deployment.go
```diff
@@ -43,6 +44,17 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
 	})
 }
 
+// UpdateDeployment updates the specified deployment
+func UpdateDeployment(c client.Client, deployment *kappsapi.Deployment, updated *kappsapi.Deployment) error {
+	return wait.PollImmediate(1*time.Second, WaitMedium, func() (bool, error) {
```
JoelSpeed, line 49: Perhaps we should add a constant for `retryShort` in framework.go? We only have `retryMedium` right now which is 5 seconds

pkg/framework/deployment.go
```diff
@@ -43,6 +44,17 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
...
+		if err := c.Patch(context.TODO(), deployment.DeepCopy(), client.MergeFrom(updated)); err != nil {
+			glog.Errorf("error querying api for deployment object %q: %v, retrying...", deployment.Name, err)
```
JoelSpeed, line 51: Can you switch this over to klog please?

pkg/framework/deployment.go
```diff
@@ -43,6 +44,17 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
 	})
 }
 
+// UpdateDeployment updates the specified deployment
+func UpdateDeployment(c client.Client, deployment *kappsapi.Deployment, updated *kappsapi.Deployment) error {
```
JoelSpeed, line 48: Does this work? I assume the only reason a patch would fail is because there is a conflict, in which case you'd just be retrying the same operation over and over and constantly get conflicts no?

pkg/framework/deployment.go
JoelSpeed, line 51: We aren't querying here, we are patching, perhaps this is better?
```suggestion
			klog.Errorf("error patching deployment object %q: %v, retrying...", deployment.Name, err)
```

pkg/framework/deployment.go
```diff
@@ -66,6 +78,27 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
 	return true
 }
 
+// IsDeploymentSynced returns true if provided deployment spec matched one found on cluster
+func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment, name, namespace string) bool {
+	if err := wait.PollImmediate(1*time.Second, WaitLong, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace)
+		if err != nil {
+			glog.Errorf("Error getting deployment: %v", err)
```
JoelSpeed, line 86:
```suggestion
			klog.Errorf("Error getting deployment: %v", err)
```

pkg/framework/deployment.go
```diff
@@ -66,6 +78,27 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
...
+			return false, nil
+		}
+		if !reflect.DeepEqual(d.Spec, dep.Spec) {
+			return true, fmt.Errorf("Deployment %q is updated. Spec is not equal to: %v",
```
JoelSpeed, line 90: It's not updated right?
```suggestion
			return true, fmt.Errorf("Deployment %q is not updated. Spec is not equal to: %v",
```

pkg/operators/machine-api-operator.go
```diff
@@ -39,6 +40,35 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
...
+		By(fmt.Sprintf("checking deployment %q spec matches", maoManagedDeployment))
+		Expect(framework.IsDeploymentSynced(client, initialDeployment, maoManagedDeployment, framework.MachineAPINamespace)).To(BeTrue())
+	})
+
+	It("maintains deployment spec", func() {
+		var err error
```
JoelSpeed, line 49: This line is unneeded

pkg/framework/deployment.go
```diff
@@ -66,6 +82,27 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
...
+// IsDeploymentSynced returns true if provided deployment spec matched one found on cluster
+func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment, name, namespace string) bool {
+	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace)
+		if err != nil {
+			klog.Errorf("Error getting deployment: %v", err)
+			return false, nil
+		}
+		if !reflect.DeepEqual(d.Spec, dep.Spec) {
+			return true, fmt.Errorf("Deployment %q is not updated. Spec is not equal to: %v",
```
elmiko, line 94: shouldn't this return false and the error?

Danil-Grigorev: More like polling here is needless :) Thanks, just returning `false` now. But in case of error it will short circuit polling and return error anyways

pkg/framework/framework.go
```diff
@@ -34,6 +34,7 @@ const (
 	WaitMedium              = 3 * time.Minute
 	WaitLong                = 15 * time.Minute
 	RetryMedium             = 5 * time.Second
+	RetryShort              = 1 * time.Second
```
JoelSpeed, line 37: Small one, I would reorder these for consistency with the above ones for waiting
```suggestion
	RetryShort              = 1 * time.Second
	RetryMedium             = 5 * time.Second
```

Danil-Grigorev: Fixed

PR #159, opened 2020-05-27
OCPCLOUD-439 - Make "recover from deleted worker machines" test more disruptive

Danil-Grigorev:
Ensure e2e test check recovery from all machines deletion

### DIFF
--- a/pkg/framework/machines.go
+++ b/pkg/framework/machines.go
@@ -103,26 +103,27 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
 	return machine, nil
 }
 
-// DeleteMachine deletes a specific machine and returns an error otherwise
-func DeleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) error {
+// DeleteMachines deletes the specified machines and returns an error on failure.
+func DeleteMachines(client runtimeclient.Client, machines ...*mapiv1beta1.Machine) error {
 	return wait.PollImmediate(1*time.Second, time.Minute, func() (bool, error) {
-		if err := client.Delete(context.TODO(), machine); err != nil {
-			klog.Errorf("Error querying api for machine object %q: %v, retrying...", machine.Name, err)
-			return false, err
+		for _, machine := range machines {
+			if err := client.Delete(context.TODO(), machine); err != nil {
+				klog.Errorf("Error querying api for machine object %q: %v, retrying...", machine.Name, err)
+				return false, err
+			}
 		}
 		return true, nil
 	})
 }
 
-// WaitForMachineDelete polls until the given Machines are not found.
-func WaitForMachineDelete(c client.Client, machines ...*mapiv1beta1.Machine) {
+// WaitForMachinesDeleted polls until the given Machines are not found.
+func WaitForMachinesDeleted(c client.Client, machines ...*mapiv1beta1.Machine) {
 	Eventually(func() bool {
 		for _, m := range machines {
 			err := c.Get(context.Background(), client.ObjectKey{
 				Name:      m.GetName(),
 				Namespace: m.GetNamespace(),
 			}, &mapiv1beta1.Machine{})
-
 			if !apierrors.IsNotFound(err) {
 				return false // Not deleted, or other error.
 			}
--- a/pkg/infra/infra.go
+++ b/pkg/infra/infra.go
@@ -229,12 +229,10 @@ var _ = Describe("[Feature:Machines] Managed cluster should", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(machines).ToNot(BeEmpty())
 
-		machine := machines[0]
-
-		By(fmt.Sprintf("deleting machine object %q", machine.Name))
-		err = framework.DeleteMachine(client, machine)
+		By(fmt.Sprint("deleting all machines"))
+		err = framework.DeleteMachines(client, machines...)
 		Expect(err).NotTo(HaveOccurred())
-		framework.WaitForMachineDelete(client, machine)
+		framework.WaitForMachinesDeleted(client, machines...)
 
 		framework.WaitForMachineSet(client, machineSet.GetName())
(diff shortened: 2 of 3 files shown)


pkg/framework/machines.go
```diff
@@ -104,11 +104,13 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
 }
 
 // DeleteMachine deletes a specific machine and returns an error otherwise
-func DeleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) error {
+func DeleteMachine(client runtimeclient.Client, machines ...*mapiv1beta1.Machine) error {
```
alexander-demichev, line 107: should this be `DeleteMachines` ?

Danil-Grigorev: No

arapov: Out of curiosity, how is it not? It gets a number of arguments?

pkg/framework/machines.go
```diff
@@ -104,11 +104,13 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
 	return wait.PollImmediate(1*time.Second, time.Minute, func() (bool, error) {
-		if err := client.Delete(context.TODO(), machine); err != nil {
-			glog.Errorf("Error querying api for machine object %q: %v, retrying...", machine.Name, err)
-			return false, err
+		for _, machine := range machines {
```
arapov, line 109: Iterates here.

pkg/framework/machines.go
```diff
@@ -104,11 +104,13 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
+			if err := client.Delete(context.TODO(), machine); err != nil {
```
arapov, line 110: And delete one by one here....

pkg/framework/machines.go
Danil-Grigorev: It revolves out of the fact you could pass a single instance. There is no reason for referencing it with DeleteMachine`s`

alexander-demichev: If you are able to pass multiple machines, then it should be called `DeleteMachines`. If you want very much to have a function called `DeleteMachine` make a wrapper.

arapov: I don't know... This also means that the comment to the function should be clarified. I would be stuck seeing one thing in a comment and other in a function.

pkg/framework/machines.go
```diff
@@ -118,13 +120,16 @@ func DeleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) er
 func WaitForMachineDelete(c client.Client, machines ...*mapiv1beta1.Machine) {
```
JoelSpeed, line 120: If this is now iterating, perhaps we should call it `WaitForMachinesDeleted`?

enxebre: same for `DeleteMachine`

pkg/framework/machines.go
```diff
@@ -103,28 +103,34 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
 	return machine, nil
 }
 
-// DeleteMachine deletes a specific machine and returns an error otherwise
-func DeleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) error {
+// DeleteMachines deletes a specific machine and returns an error otherwise
```
alexander-demichev, line 106:
```suggestion
// DeleteMachines deletes a specific machines and returns an error otherwise
```

pkg/framework/machines.go
```diff
@@ -103,28 +103,34 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
 	return machine, nil
 }
 
-// DeleteMachine deletes a specific machine and returns an error otherwise
-func DeleteMachine(client runtimeclient.Client, machine *mapiv1beta1.Machine) error {
+// DeleteMachines deletes a specific machines and returns an error otherwise
```
elmiko, line 106: just a nit about the comment,

```suggestion
// DeleteMachines deletes the specified machines and returns an error on failure.
```

pkg/framework/machines.go
```diff
@@ -103,28 +103,34 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
 	})
 }
 
-// WaitForMachineDelete polls until the given Machines are not found.
-func WaitForMachineDelete(c client.Client, machines ...*mapiv1beta1.Machine) {
+// WaitForMachinesDeleted polls until the given Machines are not found.
+func WaitForMachinesDeleted(c client.Client, machines ...*mapiv1beta1.Machine) {
 	Eventually(func() bool {
 		for _, m := range machines {
+			currentMachine := &mapiv1beta1.Machine{}
```
enxebre, line 123: why this change?

pkg/framework/machines.go
```diff
@@ -103,28 +103,34 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
 			err := c.Get(context.Background(), client.ObjectKey{
 				Name:      m.GetName(),
 				Namespace: m.GetNamespace(),
-			}, &mapiv1beta1.Machine{})
-
-			if !apierrors.IsNotFound(err) {
-				return false // Not deleted, or other error.
+			}, currentMachine)
+			if apierrors.IsNotFound(err) {
```
enxebre, line 128: why this change?

Danil-Grigorev: For no reason. Somehow forgot that a name for the machine is randomly generated by a `machineSet`. Will revert the change.

Danil-Grigorev:
@elmiko @JoelSpeed @enxebre @alexander-demichev  Are you guys good with the changes?

enxebre:
thanks!
/retest
/lgtm

PR #160, opened 2020-05-28
MHC check to respect maxUnhealthy value

Danil-Grigorev:
OCPCLOUD-874 - Implement a validation for MHC maxUnhealthy

### DIFF
--- a/pkg/framework/framework.go
+++ b/pkg/framework/framework.go
@@ -11,10 +11,10 @@ import (
 	cov1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
@@ -119,13 +119,6 @@ func LoadClientset() (*kubernetes.Clientset, error) {
 	return kubernetes.NewForConfig(cfg)
 }
 
-// RandomString returns a random 6 character string.
-func RandomString(clusterName string) string {
-	randID := string(uuid.NewUUID())
-
-	return fmt.Sprintf("%s-%s", clusterName, randID[:6])
-}
-
 func IsStatusAvailable(client runtimeclient.Client, name string) bool {
 	key := types.NamespacedName{
 		Namespace: MachineAPINamespace,
@@ -175,4 +168,27 @@ func WaitForValidatingWebhook(client runtimeclient.Client, name string) bool {
 	}
 
 	return true
+}
+
+// WaitForEvent expects to find the given event
+func WaitForEvent(c runtimeclient.Client, kind, name, reason string) error {
+	return wait.PollImmediate(RetryMedium, WaitMedium, func() (bool, error) {
+		eventList := corev1.EventList{}
+		if err := c.List(context.Background(), &eventList); err != nil {
+			klog.Errorf("error querying api for eventList object: %v, retrying...", err)
+			return false, nil
+		}
+
+		for _, event := range eventList.Items {
+			if event.Reason != reason ||
+				event.InvolvedObject.Kind != kind ||
+				event.InvolvedObject.Name != name {
+				continue
+			}
+
+			return true, nil
+		}
+
+		return false, nil
+	})
 }
--- a/pkg/framework/machines.go
+++ b/pkg/framework/machines.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
@@ -42,6 +43,24 @@ func GetMachine(c client.Client, name string) (*mapiv1beta1.Machine, error) {
 	}
 
 	return machine, nil
+}
+
+// MachinesPresent search for each provided machine in `machines` argument in the predefined `existingMachines` list
+// and returns true when all of them were found
+func MachinesPresent(existingMachines []*mapiv1beta1.Machine, machines ...*mapiv1beta1.Machine) bool {
+	if len(existingMachines) < len(machines) {
+		return false
+	}
+	existingMachinesMap := map[types.UID]struct{}{}
+	for _, existing := range existingMachines {
+		existingMachinesMap[existing.UID] = struct{}{}
+	}
+	for _, machine := range machines {
+		if _, found := existingMachinesMap[machine.UID]; !found {
+			return false
+		}
+	}
+	return true
 }
 
 // GetMachines gets a list of machinesets from the default machine API namespace.
--- a/pkg/framework/machinesets.go
+++ b/pkg/framework/machinesets.go
@@ -3,8 +3,8 @@ package framework
 import (
 	"context"
 	"fmt"
-	"time"
 
+	"github.com/google/uuid"
 	. "github.com/onsi/gomega"
 
 	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
@@ -47,28 +47,22 @@ func BuildMachineSetParams(client runtimeclient.Client, replicas int) MachineSet
 	Expect(err).NotTo(HaveOccurred())
 	Expect(clusterInfra.Status.InfrastructureName).ShouldNot(BeEmpty())
 
+	uid, err := uuid.NewUUID()
+	Expect(err).NotTo(HaveOccurred())
+
 	return MachineSetParams{
 		Name:         clusterInfra.Status.InfrastructureName,
 		Replicas:     int32(replicas),
 		ProviderSpec: providerSpec,
 		Labels: map[string]string{
-			"mhc.framework.openshift.io": clusterInfra.Status.InfrastructureName,
-			ClusterKey:                   clusterName,
+			"e2e.openshift.io": uid.String(),
+			ClusterKey:         clusterName,
 		},
 	}
 }
 
 // CreateMachineSet creates a new MachineSet resource.
 func CreateMachineSet(c client.Client, params MachineSetParams) (*mapiv1beta1.MachineSet, error) {
-	if params.Labels == nil {
-		params.Labels = make(map[string]string)
-	}
-
-	// TODO(bison): It would be nice to automatically set the Cluster ID / name
-	// in the labels here, but I'm not sure how to easily find it.
-	params.Labels["e2e.openshift.io"] =
-		fmt.Sprintf("%s-%d", params.Name, time.Now().Unix())
-
 	ms := &mapiv1beta1.MachineSet{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "MachineSet",
@@ -306,7 +300,7 @@ func WaitForMachineSet(c client.Client, name string) {
 
 		if len(machines) != int(replicas) {
 			return fmt.Errorf("found %d Machines, but MachineSet has %d replicas",
-				len(machines), machineSet.Spec.Replicas)
+				len(machines), int(replicas))
 		}
 
 		running := FilterRunningMachines(machines)
--- a/pkg/framework/machinesets.go
+++ b/pkg/framework/machinesets.go
@@ -47,14 +47,12 @@ func BuildMachineSetParams(client runtimeclient.Client, replicas int) MachineSet
 	Expect(err).NotTo(HaveOccurred())
 	Expect(clusterInfra.Status.InfrastructureName).ShouldNot(BeEmpty())
 
-	msName := RandomString(clusterInfra.Status.InfrastructureName)
-
 	return MachineSetParams{
-		Name:         msName,
+		Name:         clusterInfra.Status.InfrastructureName,
 		Replicas:     int32(replicas),
 		ProviderSpec: providerSpec,
 		Labels: map[string]string{
-			"mhc.framework.openshift.io": msName,
+			"mhc.framework.openshift.io": clusterInfra.Status.InfrastructureName,
 			ClusterKey:                   clusterName,
 		},
 	}
@@ -77,9 +75,9 @@ func CreateMachineSet(c client.Client, params MachineSetParams) (*mapiv1beta1.Ma
 			APIVersion: "machine.openshift.io/v1beta1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      params.Name,
-			Namespace: MachineAPINamespace,
-			Labels:    params.Labels,
+			GenerateName: params.Name,
+			Namespace:    MachineAPINamespace,
+			Labels:       params.Labels,
 		},
 		Spec: mapiv1beta1.MachineSetSpec{
 			Selector: metav1.LabelSelector{
--- a/pkg/machinehealthcheck/machinehealthcheck.go
+++ b/pkg/machinehealthcheck/machinehealthcheck.go
@@ -2,13 +2,12 @@ package machinehealthcheck
 
 import (
 	"context"
-	"fmt"
-	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
+	"github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -19,7 +18,9 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
 	var client client.Client
 
 	var machineSet *mapiv1beta1.MachineSet
-	var machineSetParams framework.MachineSetParams
+	var machinehealthcheck *v1beta1.MachineHealthCheck
+	var maxUnhealthy = 3
+	const expectedReplicas = 5
 
 	const E2EConditionType = "MachineHealthCheckE2E"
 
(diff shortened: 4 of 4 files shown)


pkg/framework/machines.go
```diff
@@ -103,28 +125,34 @@ func GetMachineFromNode(client runtimeclient.Client, node *corev1.Node) (*mapiv1
...
+			}
 		}
 		return true, nil
 	})
 }
 
-// WaitForMachineDelete polls until the given Machines are not found.
-func WaitForMachineDelete(c client.Client, machines ...*mapiv1beta1.Machine) {
+// WaitForMachinesDeleted polls until the given Machines are not found.
+func WaitForMachinesDeleted(c client.Client, machines ...*mapiv1beta1.Machine) {
```
Danil-Grigorev, line 142: Waits for https://github.com/openshift/cluster-api-actuator-pkg/pull/159 functionality to be merged.

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -101,24 +88,95 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+			Name:         msName,
+			Replicas:     3,
+			ProviderSpec: providerSpec,
+			Labels: map[string]string{
+				"mhc.framework.openshift.io": msName,
+				framework.ClusterKey:         clusterName,
+			},
+		}
+
+		By("Creating a new MachineSet")
```
alexander-demichev, line 126: Creating and deleting machineset is done in `BeforeEach()` and `AfterEach()`

Danil-Grigorev: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/160/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/195 What else could cause this, but a double create call in a parallel execution of `BeforeEach` with a create call. There is definitely a race condition somewhere. Something like `BeforeAll`/`AfterAll` would be perfect

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -46,10 +46,6 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
 		providerSpec := workers[0].Spec.Template.Spec.ProviderSpec.DeepCopy()
 		clusterName := workers[0].Spec.Template.Labels[framework.ClusterKey]
 
-		// TODO(bison): This should probably be appended with
-		// something other than a timestamp, e.g. a random string.
-		msName := fmt.Sprintf("e2e-mhc-%d", time.Now().Unix())
```
elmiko, line 51: i see that msName has moved to the outer scope, but i am wondering about this TODO note. do we need to have some sort random string on the name?

pkg/infra/infra.go
```diff
@@ -229,12 +229,10 @@ var _ = Describe("[Feature:Machines] Managed cluster should", func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(machines).ToNot(BeEmpty())
 
-		machine := machines[0]
-
-		By(fmt.Sprintf("deleting machine object %q", machine.Name))
-		err = framework.DeleteMachine(client, machine)
+		By(fmt.Sprint("deleting all machines"))
+		err = framework.DeleteMachines(client, machines...)
```
enxebre, line 233: how is this change related to this PR?

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -118,7 +115,51 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+		Expect(machines).ToNot(BeEmpty())
+
+		for _, machine := range machines {
+			node, err := framework.GetNodeForMachine(client, machine)
+			Expect(err).ToNot(HaveOccurred())
+			err = framework.AddNodeCondition(client, node, nodeCondition)
+			Expect(err).ToNot(HaveOccurred())
+		}
+
+		By("Creating a MachineHealthCheck resource with low maxUnhealthy threshold")
```
enxebre, line 133: what is low? can we be explicit?

enxebre: can we do something more disruptive with a higher value for maxUnhealthy? wait for a few machines to be remediated and see the MHC short circuit for the others?

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -20,7 +18,9 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
 
 	var machineSet *mapiv1beta1.MachineSet
 	var machineSetParams framework.MachineSetParams
+	var maxAllowedUnhealthy = 1
```
enxebre, line 21: can we name this consistently with api field, i.e `MaxUnhealthy`

enxebre: also should it be rather a constant?

pkg/framework/machines.go
```diff
@@ -44,6 +44,28 @@ func GetMachine(c client.Client, name string) (*mapiv1beta1.Machine, error) {
 	return machine, nil
 }
 
+// MachinesPresent looks for matching machine in a predefined list
+func MachinesPresent(existingMachines []*mapiv1beta1.Machine, machines ...*mapiv1beta1.Machine) bool {
+	if len(existingMachines) < len(machines) {
+		return false
+	}
+	for _, machine := range machines {
```
enxebre, line 52: this nested loops would be cheaper in terms of complexity if you have two independent loops over a map and hash objects by UID.

pkg/framework/machines.go
```diff
@@ -44,6 +44,28 @@ func GetMachine(c client.Client, name string) (*mapiv1beta1.Machine, error) {
 	return machine, nil
 }
 
+// MachinesPresent looks for matching machine in a predefined list
```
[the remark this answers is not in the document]
Danil-Grigorev: You are right, the comment message is completely misleading. It is supposed to check that each machine in `machines` list is present in provided `existingMachines` list.

pkg/infra/infra.go
Danil-Grigorev: It is the same thing, as in comment https://github.com/openshift/cluster-api-actuator-pkg/pull/160/files#r431705540, will completely exclude this and rebase after merging. Letft it here in hope to see CI jobs passing.

pkg/machinehealthcheck/machinehealthcheck.go
Danil-Grigorev: I decided to move onto `GenerateName` which will have the same effect here, and will give a unique name for the machine - https://github.com/openshift/cluster-api-actuator-pkg/pull/160/files#diff-c7d1003b84db4b5057508a81ab203c46R80

pkg/machinehealthcheck/machinehealthcheck.go
Danil-Grigorev: It should, but I can't pass a pointer to a constant.

pkg/machinehealthcheck/machinehealthcheck.go
Danil-Grigorev: I extended the case and made it more visible, so it will ensures the machines will be remediated when the number of unhealthy entities does not reach the threshold, and won't be past that. Also added checks for all events being emitted during that procedure.

pkg/framework/framework.go
```diff
@@ -176,3 +177,26 @@ func WaitForValidatingWebhook(client runtimeclient.Client, name string) bool {
...
 	return true
 }
+
+// WaitForEvent expects to find the given event in a reasonable timeframe
+func WaitForEvent(c runtimeclient.Client, kind, name, reason string) error {
+	return wait.PollImmediate(RetryMedium, WaitShort, func() (bool, error) {
+		eventList := corev1.EventList{}
+		if err := c.List(context.TODO(), &eventList); err != nil {
+			glog.Errorf("error querying api for eventList object: %v, retrying...", err)
+			return false, nil
```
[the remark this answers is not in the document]
elmiko: @Danil-Grigorev you marked this as resolved, but i don't see any change or discussion. what's the story?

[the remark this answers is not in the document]
Danil-Grigorev: Sorry, I picked the change, but returning the error here was causing CI to fail, so I reverted it. Especially in a long running task like this, returning an error which aborts the loop is dangerous...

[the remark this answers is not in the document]
elmiko: weird that it was causing failure since you still return false as well. but, no worries, thanks for the update!

pkg/framework/framework.go
```diff
@@ -176,3 +177,26 @@ func WaitForValidatingWebhook(client runtimeclient.Client, name string) bool {
 
 	return true
 }
+
+// WaitForEvent expects to find the given event
+func WaitForEvent(c runtimeclient.Client, kind, name, reason string) error {
+	return wait.PollImmediate(RetryMedium, WaitMedium, func() (bool, error) {
+		eventList := corev1.EventList{}
+		if err := c.List(context.TODO(), &eventList); err != nil {
```
JoelSpeed, line 185: Use background for tests where we don't think we will replace it at a later date
```suggestion
		if err := c.List(context.Background(), &eventList); err != nil {
```

pkg/framework/machines.go
```diff
@@ -44,6 +45,24 @@ func GetMachine(c client.Client, name string) (*mapiv1beta1.Machine, error) {
...
+// MachinesPresent search for each provided machine in `machines` argument in the predefined `existingMachines` list
+// and returns true when all of them were found
+func MachinesPresent(existingMachines []*mapiv1beta1.Machine, machines ...*mapiv1beta1.Machine) bool {
+	if len(existingMachines) < len(machines) {
+		return false
+	}
+	existingMachinesMap := map[types.UID]bool{}
+	for _, existing := range existingMachines {
+		existingMachinesMap[existing.UID] = true
+	}
```
JoelSpeed, line 57: Technically this is more efficient, but what you have is fine, just as an education point, change if you want. A bool takes up more memory than an empty struct
```suggestion
	existingMachinesMap := map[types.UID]struct{}{}
	for _, existing := range existingMachines {
		existingMachinesMap[existing.UID] = struct{}
	}
```

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
 	})
 
 	AfterEach(func() {
+		By("Deleting the MachineHealthCheck resource")
+		err := client.Delete(context.Background(), machinehealthcheck)
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 54: I would have inlined this, but as it matches the existing, can leave it (or update both) `Expect(client.Delete(...)).To(Succeed()`

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
 
-		By("Verifying the matching Machine is deleted")
-		framework.WaitForMachineDelete(client, machine)
+		By("Waiting for each unhealthy machine to be deleted")
+		framework.WaitForMachinesDeleted(client, unhealthyMachines...)
+
+		By("Waiting for MachineDeleted event from MachineHealthCheck for each unhealthy machine")
+		for _, machine := range unhealthyMachines {
+			err = framework.WaitForEvent(client, "Machine", machine.Name, "MachineDeleted")
+			Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 102: Could inline this? That way we don't shadow the err variable
```suggestion
			Expect(framework.WaitForEvent(client, "Machine", machine.Name, "MachineDeleted")).To(Succeed())
```

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(machines).ToNot(BeEmpty())
+
+		By("Setting unhealthy conditions on machine nodes, but exceding maxUnhealthy threshold")
+		unhealthyMachines := machines[:maxUnhealthy+1]
+		for _, machine := range unhealthyMachines {
+			node, err := framework.GetNodeForMachine(client, machine)
+			Expect(err).ToNot(HaveOccurred())
+			err = framework.AddNodeCondition(client, node, nodeCondition)
+			Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 121: Could inline this

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+			},
+			MaxUnhealthy: &maxUnhealthy,
+		}
+
+		machinehealthcheck, err = framework.CreateMHC(client, mhcParams)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(machinehealthcheck).ToNot(BeNil())
+
+		By("Waiting for RemediationRestricted event from MachineHealthCheck")
+		err = framework.WaitForEvent(client, "MachineHealthCheck", machineSet.Name, "RemediationRestricted")
```
JoelSpeed, line 143: Could inline this

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
 		Expect(err).ToNot(HaveOccurred())
-	})
 
-	// It("should respect maxUnhealthy", func() {
-	// 	// TODO(bison): This.
-	// })
+		By("Ensuring none of the unhealthy machines are deleted")
```
JoelSpeed, line 146: Should this be healthy rather than unhealthy?

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+		untouchedMachines, err := framework.GetMachines(client, &selector)
```
JoelSpeed, line 147: Perhaps healthy rather than untouched, untouched seems a bit of a weird description to me 😅

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,105 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
-
-		err = framework.AddNodeCondition(client, node, nodeCondition)
-		Expect(err).ToNot(HaveOccurred())
+		By("Setting unhealthy conditions on machine nodes, but not exceding maxUnhealthy threshold")
+		unhealthyMachines, healthyMachines := machines[:maxUnhealthy-1], machines[maxUnhealthy-1:]
+		for _, machine := range unhealthyMachines {
+			node, err := framework.GetNodeForMachine(client, machine)
+			Expect(err).ToNot(HaveOccurred())
+			err = framework.AddNodeCondition(client, node, nodeCondition)
+			Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 73: Potentially inline this

pkg/machinehealthcheck/machinehealthcheck.go
Danil-Grigorev: Realized that it is no longer needed.

pkg/framework/machinesets.go
```diff
@@ -47,39 +47,31 @@ func BuildMachineSetParams(client runtimeclient.Client, replicas int) MachineSet
...
 // CreateMachineSet creates a new MachineSet resource.
 func CreateMachineSet(c client.Client, params MachineSetParams) (*mapiv1beta1.MachineSet, error) {
-	if params.Labels == nil {
-		params.Labels = make(map[string]string)
-	}
-
-	// TODO(bison): It would be nice to automatically set the Cluster ID / name
-	// in the labels here, but I'm not sure how to easily find it.
-	params.Labels["e2e.openshift.io"] =
-		fmt.Sprintf("%s-%d", params.Name, time.Now().Unix())
```
JoelSpeed, line 72: Will this effect other tests? How many of the tests use `CreateMachineSet`?

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,104 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
-		Expect(mhc).ToNot(BeNil())
+		Expect(machinehealthcheck).ToNot(BeNil())
+
+		By("Waiting for each unhealthy machine to be deleted")
+		framework.WaitForMachinesDeleted(client, unhealthyMachines...)
+
+		By("Waiting for MachineDeleted event from MachineHealthCheck for each unhealthy machine")
+		for _, machine := range unhealthyMachines {
+			Expect(framework.WaitForEvent(client, "Machine", machine.Name, "MachineDeleted")).ToNot(HaveOccurred())
+		}
```
JoelSpeed, line 99: Should these be the other way around, since the events will happen before the machines are deleted right?

Danil-Grigorev: This happens roughly at the same time, and the events are persisted for some time, so I left it there. Also, if the machines are not deleted, we'd rather see the timeout on machine deletion, then on some event search in my opinion. It is better for debugging imo

pkg/framework/machinesets.go
Danil-Grigorev: Yes, it will ensure we will never see a race condition, when the CreateMachineSet was called at the same time, so the labels on two machines would be matching, and there will be no good way for tests to find out how many machines the machine et actually spawned

JoelSpeed: Sounds like a good improvement 👍

pkg/machinehealthcheck/machinehealthcheck.go
JoelSpeed: That's a reasonable argument, sounds good to me :)

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -68,57 +49,104 @@ var _ = Describe("[Feature:MachineHealthCheck] MachineHealthCheck", func() {
...
+		Expect(allMachines).ToNot(BeEmpty())
+		Expect(framework.MachinesPresent(allMachines, healthyMachines...)).To(BeTrue())
 
 		By("Verifying the MachineSet recovers")
 		framework.WaitForMachineSet(client, machineSet.GetName())
+	})
 
-		By("Deleting the MachineHealthCheck resource")
-		err = client.Delete(context.Background(), mhc)
+	It("should not remediate larger number of unhealthy machines then maxUnhealthy", func() {
```
elmiko, line 111: this should be "than" rather then "then", but it's not a blocker for me.

PR #162, opened 2020-06-16
Test for balanced nodes

elmiko:
This change brings in a test for the --balance-similar-nodes option of the cluster autoscaler. It also includes a change which refactors the event watcher to be more common among the autoscaler tests.

The new test will create 2 MachineSets and then a workload which should demonstrate a balanced growth of those sets without reaching their maximum node sizes.

This change is dependent on #151

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


elmiko:
/hold

need to wait for #151 , also seeing some odd behavior on this test it might need further tweaking.

pkg/autoscaler/autoscaler.go
```diff
@@ -249,11 +250,26 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 		// node.
 		workloadMemRequest = resource.MustParse(fmt.Sprintf("%v", 0.7*float32(bytes)))
 
+		By("Starting Cluster Autoscaler event watcher")
```
enxebre, line 253: Why does this commit needs to go in this PR?

enxebre: Can we have a different PR discussion for that? I'd be in favour of not relying on core autoscaler events for e2e testing which might break e2e on each rebase.

elmiko: sure, i can break that out into a different PR.

my reasoning for adding it in at the top level is that i've found it invaluable in debugging failures. it is not used for any testing purposes outside of the "scale up and down" test, it will print the autoscaler events as they are happening during the test.

elmiko: created #163

elmiko:
removing the hold as this has been rebased on master.
/hold cancel

pkg/autoscaler/autoscaler.go
```diff
@@ -674,4 +674,90 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+		})
+
+		It("places nodes evenly across node groups [Slow]", func() {
+			By("Creating 2 MachineSets each with 1 replica")
+			var transientMachineSets [2]*mapiv1beta1.MachineSet
+			targetedNodeLabel := fmt.Sprintf("%v-delete-cleanup", autoscalerWorkerNodeRoleLabel)
+			for i, machineSet := range transientMachineSets {
+				machineSetParams := framework.BuildMachineSetParams(client, 1)
+				machineSetParams.Labels[targetedNodeLabel] = ""
+				delete(machineSetParams.Labels, "e2e.openshift.io")
```
JoelSpeed, line 715: Is this required so that the labels match? If so, could do with a comment IMO

pkg/autoscaler/autoscaler.go
```diff
@@ -674,4 +674,90 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+				machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+				Expect(err).ToNot(HaveOccurred())
+				cleanupObjects[machineSet.GetName()] = machineSet
+				transientMachineSets[i] = machineSet
+			}
```
JoelSpeed, line 720: Perhaps somewhere in the test, before creating the machine autoscalers, we could verify that the labels match?

pkg/autoscaler/autoscaler.go
```diff
@@ -674,4 +674,90 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+				By(fmt.Sprintf("Waiting for MachineSet %s replicas to scale out", machineSet.GetName()))
+			}
+			expectedReplicas := int32(2)
+			Eventually(func() (bool, error) {
+				for _, machineSet := range transientMachineSets {
+					current, err := framework.GetMachineSet(client, machineSet.GetName())
+					if err != nil {
+						return false, err
+					}
+					if *current.Spec.Replicas != expectedReplicas {
```
JoelSpeed, line 755: Any chance this dereference could fail and cause the test to panic?

pkg/autoscaler/autoscaler.go
```diff
@@ -674,4 +674,90 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			workload := newWorkLoad(jobReplicas, workloadMemRequest, targetedNodeLabel)
+			cleanupObjects[workload.GetName()] = workload
+			Expect(client.Create(ctx, workload)).Should(Succeed())
+
+			for _, machineSet := range transientMachineSets {
+				By(fmt.Sprintf("Waiting for MachineSet %s replicas to scale out", machineSet.GetName()))
+			}
+			expectedReplicas := int32(2)
+			Eventually(func() (bool, error) {
+				for _, machineSet := range transientMachineSets {
```
JoelSpeed, line 750: Maybe combine these loops? So the By is immediately before the check?

pkg/autoscaler/autoscaler.go
elmiko: yes, i'll add that

pkg/autoscaler/autoscaler.go
elmiko: that's a good idea! i can certainly incorporate it

JoelSpeed: It's a good early failure to indicate our test case is bad in case something changes at a later date, so probably worth a comment to that effect too

pkg/autoscaler/autoscaler.go
elmiko: yeah, i can make it look like the other loops

pkg/autoscaler/autoscaler.go
elmiko: it _shouldn't_ but checking is easy, i'll add something

pkg/autoscaler/autoscaler.go
elmiko: agreed. i was thinking about adding a comment section here with a reference to the autoscaler faq about balance nodes. just to be clear on why we need it.

elmiko:
updated with requested changes

pkg/autoscaler/autoscaler.go
```diff
@@ -674,4 +675,96 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+				// remove this label to make the MachineSets similar, see below for more details
+				delete(machineSetParams.Labels, "e2e.openshift.io")
+				machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+				Expect(err).ToNot(HaveOccurred())
+				cleanupObjects[machineSet.GetName()] = machineSet
+				transientMachineSets[i] = machineSet
+			}
+
+			// balance similar nodes requires that all the participating MachineSets have the same
+			// instance types and labels. while the instance types should be the same, we want to
```
michaelgugino, line 725: Balance similar nodes should work off of CPU/RAM/GPU provided primarily, then instance types as a fall back if that info is not available.

Why do the labels need to be the same?  Not saying your tests are wrong, but that seems like a broken implementation.  Or is this magic all just needed for the tests?

elmiko: this is a limitation coming from the autoscaler core, [referenced here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#im-running-cluster-with-nodes-in-multiple-zones-for-ha-purposes-is-that-supported-by-cluster-autoscaler)

> If you set the flag to true, CA will automatically identify node groups with the same instance type and the same set of labels (except for automatically added zone label) and try to keep the sizes of those node groups balanced.

iirc, MachineSets basically equate to node groups internally for the autoscaler, so it's just asking for all the labels and then checking them (at the autoscaler). this might be inefficient behavior from the autoscaler, but thinking about normalizing the labels before returning to the autoscaler seems like it could cause more problems.

elmiko:
> Looks good, though I have one suggestion. It might be better to explicitly delete any machinesets that we are creating and wait for them to go away before finishing the test. This prevents this test from (potentitally) interfering with others. The alternative is to have other tests wait for their desired state to happen before starting their tests, which could also be a valid way of achieving this

we do have the AfterEach clause which does an expect delete to have succeeded for each object created, but are you suggesting we modify that to wait for the objects to finish deleting as well?

JoelSpeed:
> we do have the AfterEach clause which does an expect delete to have succeeded for each object created, but are you suggesting we modify that to wait for the objects to finish deleting as well?

For machinesets in particular, we have a method in the framework that waits for the machines to be gone, would be good to call that when we delete the MachineSets so that we know that they, and their child resources are all gone before we start the next test. Otherwise they could interfere

elmiko:
> For machinesets in particular, we have a method in the framework that waits for the machines to be gone, would be good to call that when we delete the MachineSets so that we know that they, and their child resources are all gone before we start the next test. Otherwise they could interfere

i don't have an objection to that, but i do wonder if i should create a new pr to add waits for all the machinesets we create? it just seems like if i add it for this test, i should add for the others as well.

JoelSpeed:
> i don't have an objection to that, but i do wonder if i should create a new pr to add waits for all the machinesets we create? it just seems like if i add it for this test, i should add for the others as well.

Happy for it to be another PR, just keep this in mind as it may cause the test suites to flake (Eg see fixes in #169)
/lgtm

PR #163, opened 2020-06-17
Refactor event watcher for autoscaler tests

elmiko:
This change moves the common functionality of the event watcher into the
global Describe scope for the autoscaler tests. It is being moved to
make it more widely available without duplicating code. As more tests
are added this event watcher proves more useful, it should be common to
all tests for the autoscaler.


elmiko:
/hold
this needs #151 first

elmiko:
removing the hold as this has been rebased off master.
/hold cancel

elmiko:
dropping this as i think it will change in an upcoming patch to the scale up/down tests.

PR #166, opened 2020-06-23
Add GCP support for spot instances tests

alexander-demichev:
Add GCP support for spot instances tests.
This should cover the following PRs: 
https://github.com/openshift/machine-api-operator/pull/628
https://github.com/openshift/cluster-api-provider-gcp/pull/97
https://github.com/openshift/cluster-api-provider-gcp/pull/100
https://github.com/openshift/cluster-api-provider-gcp/pull/99


alexander-demichev:
/hold

This won't pass CI until https://github.com/openshift/cluster-api-provider-gcp/pull/99 gets merged

pkg/infra/spot.go
```diff
@@ -37,6 +38,7 @@ var _ = Describe("[Feature:Machines] Running on Spot", func() {
 		platform = clusterInfra.Status.PlatformStatus.Type
 		switch platform {
 		case configv1.AWSPlatformType:
+		case configv1.GCPPlatformType:
```
JoelSpeed, line 41: Nit, I'd keep these as a list
```suggestion
		case configv1.AWSPlatformType, configv1.GCPPlatformType:
```

JoelSpeed:
@alexander-demichev This doesn't appear to be working for some reason, can you spin up a nightly build and manually verify that the preempible instances are being created as expected?

JoelSpeed:
This is blocked on https://github.com/openshift/cluster-api-provider-gcp/pull/100

enxebre:
Can you please link in the desc and the commit the counter PRs that we are testing here?

enxebre:
/hold
https://github.com/openshift/cluster-api-actuator-pkg/pull/166#issuecomment-656090364

pkg/infra/spot.go
```diff
@@ -66,7 +66,7 @@ var _ = Describe("[Feature:Machines] Running on Spot", func() {
 		Expect(machines).To(HaveLen(3))
 
 		for _, machine := range machines {
-			Expect(machine.Spec.GetLabels()).To(HaveKeyWithValue(machinecontroller.MachineInterruptibleInstanceLabelName, ""))
+			Expect(machine.Spec.Labels).To(HaveKeyWithValue(machinecontroller.MachineInterruptibleInstanceLabelName, ""))
```
enxebre, line 69: why? does this make any difference?

pkg/framework/machinesets.go
```diff
@@ -153,7 +153,11 @@ func GetWorkerMachineSets(client runtimeclient.Client) ([]*mapiv1beta1.MachineSe
 	// but the Machines themselves are labled as such via the template., so we
 	// can reach into the template and check the lables there.
 	for i, ms := range machineSets.Items {
-		labels := ms.Spec.Template.GetLabels()
+		labels := ms.Spec.Template.Labels
```
enxebre, line 156: why drop GetLabels()? does this make any difference?

pkg/infra/spot.go
enxebre: I realised this is because we use our own `ObjectMeta`.
@JoelSpeed do we still need that?
anyway can tackle separately so
/lgtm

JoelSpeed: If we rely on webhooks for validation, we may be able to remove it now, will add a card to investigate switching back for a future sprint. It won't protect us completely because we have the failure policy open, so we could regress on the bug we fixed with it.

JoelSpeed:
/hold

This has merge conflicts, no point restesting while those are present

PR #168, opened 2020-07-02
[OCPCLOUD-895] Ensure Webhooks are tested in E2E suite

JoelSpeed:
This PR adds tests that check that the defaulting and validating webhooks for Machines and MachineSets are available and working within the cluster by checking that a minimal providerspec still creates machines and that updating a providerspec to remove required fields results in a rejection by the validating webhook


go.mod
```diff
@@ -16,8 +16,14 @@ require (
...
 	sigs.k8s.io/controller-runtime v0.6.0
 )
 
 // Use openshift forks
-replace sigs.k8s.io/cluster-api-provider-aws => github.com/openshift/cluster-api-provider-aws v0.2.1-0.20200429204912-67edca9c65de
+replace sigs.k8s.io/cluster-api-provider-aws => github.com/openshift/cluster-api-provider-aws v0.2.1-0.20200618031251-e16dd65fdd85
+
+replace sigs.k8s.io/cluster-api-provider-azure => github.com/openshift/cluster-api-provider-azure v0.1.0-alpha.3.0.20200620092221-ff90663025f1
+
+replace sigs.k8s.io/cluster-api-provider-gcp => github.com/openshift/cluster-api-provider-gcp v0.0.1-0.20200618110630-09cef98ae787
```
enxebre, line 37: gcp is not a forked repo

JoelSpeed: Good catch, will fix

pkg/infra/webhooks.go
```diff
@@ -0,0 +1,179 @@
...
+	})
+
+	AfterEach(func() {
+		Expect(deleteObjects(client, delObjects)).To(Succeed())
+	})
+
+	It("should be able to create a machine from a minimal providerSpec", func() {
+		machine := &mapiv1.Machine{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: fmt.Sprintf("%s-webhook-", machineSetParams.Name),
```
enxebre, line 73: can we prefix with `clusterInfra.Status.InfrastructureName`. This is used by installer/CI to garbage collect on destroy.

JoelSpeed: It's using the name from the `machineSetParams` which was built by `BuildMachineSetParams`, so it should have that already, it should be `<clusterInfra.Status.InfrastructureName>-webhook-<rand>` when it's created

enxebre:
thanks! overall looks great.
/approve

pkg/framework/machinesets.go
```diff
@@ -123,7 +124,13 @@ func GetMachineSets(client runtimeclient.Client, selectors ...*metav1.LabelSelec
 		return nil, fmt.Errorf("error querying api for machineSetList object: %w", err)
 	}
 
-	return machineSetList.Items, nil
+	machineSets := []*mapiv1beta1.MachineSet{}
+	for _, ms := range machineSetList.Items {
+		machineSet := ms
```
Danil-Grigorev, line 129: Shouldn't this be a `.DeepCopy()`? This is just assigning the same pointer. Am I missing something here?

pkg/framework/machinesets.go
```diff
@@ -153,7 +160,7 @@ func GetWorkerMachineSets(client runtimeclient.Client) ([]*mapiv1beta1.MachineSe
 	// but the Machines themselves are labled as such via the template., so we
 	// can reach into the template and check the lables there.
 	for i, ms := range machineSets.Items {
-		labels := ms.Spec.Template.GetLabels()
```
Danil-Grigorev, line 156: Aren't those two lines functionally equal?..

JoelSpeed: They are normally, except we changed the type of the `ObjectMeta` to our own version which doesn't have the accesor methods defined, so we have to do this now

PR #169, opened 2020-07-03
Reduce flakiness of autoscaler tests

JoelSpeed:
I've update two of the test cases that appeared to be flaky while I've been looking into the Azure E2E failures.

From what I can tell, it can sometimes take longer than expected to shut down machines based on which provider we are using.

The first test that I updated expected that the machine count and node count should be equal at the beginning of the test, while this may be true, I don't see any reason why we shouldn't give the test a moment to stabilise before starting, so now we give it a `WaitMedium` to allow the nodes/machines to fully shut down from previous tests before starting this one.

Eg:
- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/167/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1277971520013996032

The second test expected that once a termination had been requested for a machine, that it would be completely shutdown within `WaitMedium`, this does not appear to be the case for Azure and it was consistently failing to shutdown machines during within this timeout

Eg:
- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/168/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1278735958425997312
- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/166/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1278680115689033728
- https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/167/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1278668987076448256

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -653,7 +653,7 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 					}
 				}
 				return true, nil
-			}, framework.WaitMedium, pollingInterval).Should(BeTrue())
+			}, framework.WaitLong, pollingInterval).Should(BeTrue())
 
 			for _, machineSet := range transientMachineSets {
 				machines, err := framework.GetMachinesFromMachineSet(client, machineSet)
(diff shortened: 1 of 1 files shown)


Danil-Grigorev:
I think this change worth being merged quickly, as many pending PRs are suffering on azure CI, and such fix would speed the process up.
/lgtm

PR #170, opened 2020-07-06
[OCPCLOUD-865] Add Azure support for spot instances tests

JoelSpeed:
Allow Azure spot instances to be tested as part of the end to end suite


JoelSpeed:
This needs https://github.com/openshift/machine-api-operator/pull/631 to merge before the test suite will pass

JoelSpeed:
/hold

I've been doing some manual testing on our dev account and noticed that running the spot tests in parallel results in some quote limits being hit, I suspect this is the problem here as well. Will see if I can get the limits increased and unhold this once that is done.

PR #171, opened 2020-07-14
[OCPCLOUD-901] Test spot termination handlers react to termination events

JoelSpeed:
This PR adds a new test to the spot tests that deploys a mock metadata service to the openshift cluster, and then intercepts traffic from one of the spot nodes to reroute traffic from the metadata service to the mock metadata service to trigger a termination event for the Machine. This should exercise that a termination handler is capable of shutting down a Machine if it receives a termination event.

### DIFF
--- a/pkg/infra/mock/metadata_mock.go
+++ b/pkg/infra/mock/metadata_mock.go
@@ -1,0 +1,168 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"log"
+	"net/http"
+	"strings"
+)
+
+const (
+	awsTerminationPattern      = "/latest/meta-data/spot/termination-time"
+	gcpTerminationPattern      = "/computeMetadata/v1/instance/preempted"
+	azureTerminationPattern    = "/metadata/scheduledevents"
+	azureTerminationAPIVersion = "2019-08-01"
+)
+
+func main() {
+	provider := flag.String("provider", "", "Cloud Provider metadata service to mock (One of AWS, Azure or GCP)")
+	listenAddr := flag.String("listen-addr", "0.0.0.0:80", "Address on which metadata mock service should listen")
+	flag.Parse()
+
+	var handler http.Handler
+	switch {
+	case strings.EqualFold(*provider, "aws"):
+		handler = awsMetadataMockHandler()
+	case strings.EqualFold(*provider, "azure"):
+		handler = azureMetadataMockHandler()
+	case strings.EqualFold(*provider, "gcp"):
+		handler = gcpMetadataMockHandler()
+	default:
+		log.Fatal("--provider must be one of: aws, azure or gcp")
+	}
+
+	log.Printf("Starting mock metadata service for provider: %s", *provider)
+	if err := http.ListenAndServe(*listenAddr, logRequests(handler)); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// AWS instances expect an OK response to indicate that the instance has been scheduled for termination
+func awsMetadataMockHandler() http.Handler {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc(awsTerminationPattern, func(rw http.ResponseWriter, req *http.Request) {
+		rw.WriteHeader(http.StatusOK)
+	})
+
+	return mux
+}
+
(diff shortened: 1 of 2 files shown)


elmiko:
this looks cool Joel, gonna try to take a look this afternoon.

pkg/infra/mock/metadata_mock.go
```diff
@@ -0,0 +1,167 @@
...
+
+	mux.HandleFunc(azureTerminationPattern, func(rw http.ResponseWriter, req *http.Request) {
+		if req.Header.Get("Metadata") != "true" {
+			// Require the "Metadata" header to be correct
+			rw.WriteHeader(http.StatusBadRequest)
+			rw.Write([]byte("400 Bad Request"))
+			return
+		}
+
+		if req.URL.Query().Get("api-version") != "2019-08-01" {
```
elmiko, line 66: given that this api version value will change at some point in the future, i think it would be nice for visibility to pull this out as a top level var or const. just a suggestion though, not a blocker for me.

pkg/infra/spot.go
```diff
@@ -173,3 +233,244 @@ func setSpotOnGCPProviderSpec(params framework.MachineSetParams) error {
...
+	terminationSimulatorName = "termination-simulator"
+)
+
+func getTerminationSimulatorPod(nodeName string) *corev1.Pod {
+	script := fmt.Sprintf(`apk update && apk add iptables bind-tools;
+export SERVICE_IP=$(dig +short %s.${NAMESPACE}.svc.cluster.local);
+if [ -z ${SERVICE_IP} ]; then echo "No service IP"; exit 1; fi;
+iptables-nft -t nat -A OUTPUT -p tcp -d 169.254.169.254 -j DNAT --to-destination ${SERVICE_IP}:%d;
+iptables-nft -t nat -A POSTROUTING -j MASQUERADE;
+echo "Redirected metadata service to ${SERVICE_IP}:%d";`, metadataServiceMockName+"-service", metadataServiceMockPort, metadataServiceMockPort)
```
elmiko, line 362: this part looks a little crazy here, but i don't have a suggestion to make it cleaner.

elmiko:
looks like there are still a few bugs to workout in the ci, i'm seeing some failures to find the termination-simulator, but i didn't look too deep.

JoelSpeed:
This problem comes from the fact that, if the pod that we deploy is successful in removing the machine, it gets deleted, so the cleanup code can't delete it. My plan is to switch to a job that runs the pod, that way we can get around this issue as the job won't be deleted until we delete it

elmiko:
ahh, gotcha lol. makes perfect sense

pkg/infra/spot.go
alexander-demichev: iptables stuff can be clean? :D

pkg/infra/mock/metadata_mock.go
JoelSpeed: Yep, seems good. We may want to transition it at some point in the future as well, might need to make it one of a set of supported values, but I'll leave that for when we need to do it

pkg/infra/spot.go
JoelSpeed: I could switch the fmt string to just being a static string and use environment variables for the substitutions, that might make it a little cleaner

elmiko: i don't know, it seems like we are just pushing things around if we move. i'm fine to leave this, it's not a blocker for me just something i noticed.

JoelSpeed: I moved it in a separate commit, so can undo easily, wdyt, this way or the old way? Not sure how else to make it cleaner

elmiko: i think this is an improvement for readability. the env variables do make it easier for me to understand what is happening. big +1 here, thanks!

JoelSpeed:
The GCP tests won't pass until https://github.com/openshift/cluster-api-provider-gcp/pull/103 is merged, though I have manually verified that the tests pass once that PR is included.

Likewise, the Azure stuff isn't merged yet, but once all of the outstanding PRs are merged, this does work against Azure.

elmiko:
this looks great to me Joel, i'd prefer to let someone else add the lgtm label but i can add if we don't get another review here.

PR #174, opened 2020-07-16
rewrite scale up and down test for autoscaler

elmiko:
This change replaces the old test with a new version. This is being done
to move away from the event based testing of the previous version
towards an observation based test. What this means is that instead of
watching for scaling events the test will now observe the sizes of
MachineSet resources to determine scaling activity.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"path"
 	"reflect"
 	"strings"
 	"time"
@@ -22,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,96 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			// To exercise the MaxNodesTotal mechanism we want to craft this
+			// MachineAutoscaler so that its maximum size would cause the cluster
+			// to grow beyond the MaxNodesTotal. This is calculated as the
+			// MaxNodesTotal - current number of nodes
+			// + 1 (for the initial size of the MachineSet)
+			// + 1 (to exceed the maximum)
+			By("Calculating maximum size for MachineAutoscaler")
+			nodes, err := framework.GetNodes(client)
+			Expect(err).NotTo(HaveOccurred())
+			maxMachineSetReplicas := int32(caMaxNodesTotal - len(nodes) + 2)
```
JoelSpeed, line 543: Why not just have the `maxMachineSetReplicas` as `caMaxNodesTotal + 1`, then the maximum size of this machineset is definitely bigger than the max size of the cluster and we don't need to worry about this sum? We are never going to hit the max right? (Assuming everything works 😅 )

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,96 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			Consistently(func() (bool, error) {
+				current, err := framework.GetMachineSet(client, transientMachineSet.GetName())
+				if err != nil {
+					return false, err
+				}
+				if pointer.Int32PtrDerefOr(current.Spec.Replicas, 0) != expectedReplicas {
+					return false, nil
+				}
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue())
```
JoelSpeed, line 589: Should we not just watch the size of the cluster to make sure that it doesn't exceed that maximum count? Unless I've misunderstood, we are currently not actually checking that, just a derivative of that that implies the max size is observed, assuming our logic is correct?

I would have expected this to look something like

```
<create workload>
Eventually(<cluster hits max node count>)
Expect(<all machineset nodes are ready>) // To check there's no more coming, if there are, logic is borked
Consistently(<cluster stays at max node count>)
<delete workload>
``

File: pkg/autoscaler/autoscaler.go [hunk 3]
```diff
@@ -732,6 +518,96 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			framework.WaitForMachineSet(client, transientMachineSet.GetName())
+
+			// To exercise the MaxNodesTotal mechanism we want to craft this
+			// MachineAutoscaler so that its maximum size would cause the cluster
+			// to grow beyond the MaxNodesTotal. This is calculated as the
+			// MaxNodesTotal - current number of nodes
+			// + 1 (for the initial size of the MachineSet)
+			// + 1 (to exceed the maximum)
+			By("Calculating maximum size for MachineAutoscaler")
+			nodes, err := framework.GetNodes(client)
```
JoelSpeed, line 541: What happens if some of these nodes disappear shortly after you list them, means the later numbers won't line up right? (eg previous test has nodes going away?)

pkg/autoscaler/autoscaler.go
elmiko: i thought about that, i think i went with this because if everything works as we expect it these sizes should hold accurate. i can certainly switch to just use the max cluster size though.

pkg/autoscaler/autoscaler.go
elmiko: good point, i can switch this logic to focus around max cluster size instead.

pkg/autoscaler/autoscaler.go
elmiko: it would definitely get messed up, i marked this test as "[serial]". i think when we are testing if the autoscaler will stop at the max cluster size we need to run that test by itself.

this kinda ties into something we had talked about before, namely the tests cleaning up after themselves. currently, we shouldn't have an issue with this test breaking in the middle like this, but it could. i can see that switching to checking the cluster size will probably make this interaction less harmful, especially if i just make the machineset be able to grow as large as needed.

elmiko:
updated with the suggested changes. the test now incorporates more logic to watch total node count against maximum size.

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			Consistently(func() (bool, error) {
+				nodes, err := framework.GetNodes(client)
+				if err != nil {
+					return false, err
+				}
+				if len(nodes) != caMaxNodesTotal {
+					return false, nil
+				}
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue())
```
JoelSpeed, line 589: `WaitMedium` is three minutes, I wonder if that's quite long here, do you think `WaitShort` (1 minute) would be enough?

elmiko: that's a good question, i wondered something similar while putting it together. given what i have seen, i would imagine `WaitShort` would work just fine, i think i went with medium just to make sure it sat long enough.

i can certainly adjust this to the short wait.

elmiko:
i changed the `Consistently` wait period down to `WaitShort`, i tested this against a running cluster several times and it passed everytime. i think we can probably just switch to the shorter wait, i'm not sure _soaking_ for 3 minutes adds a lot here.

my initial thought was that we would want to wait for a couple minutes to make sure nothing else was going to be added by the autoscaler, but i'm not sure there is much difference between 1 and 3 minutes given that we wait for all machines to become running.

JoelSpeed:
Sounds good to me! Thanks for doing the manual testing as well, really helpful to have that analysis! 

/approve

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+
+			By("Deleting the workload")
+			Expect(deleteObject(workload.Name, cleanupObjects[workload.Name])).Should(Succeed())
+			delete(cleanupObjects, workload.Name)
+
+			// With the workload gone, the MachineSet should scale back down to
+			// its minimum size of 1.
+			By(fmt.Sprintf("Waiting for MachineSet %s replicas to scale down", transientMachineSet.GetName()))
+			Eventually(func() (bool, error) {
+				current, err := framework.GetMachineSet(client, transientMachineSet.GetName())
```
Danil-Grigorev, line 599: Could we use `transientMachineSet` as a selector here, and compare number of Nodes in existence? By using `GetNodesFromMachineSet`

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			By(fmt.Sprintf("Waiting for cluster to scale up to %d nodes", caMaxNodesTotal))
+			Eventually(func() (bool, error) {
+				nodes, err := framework.GetNodes(client)
+				if err != nil {
+					return false, err
+				}
+				if len(nodes) != caMaxNodesTotal {
+					return false, nil
+				}
+				return true, nil
```
Danil-Grigorev, line 568: You could just do this here:
```golang
nodes, err := framework.GetNodes(client)
return len(nodes) == caMaxNodesTotal, err
``` 
It will assert the first value is true, and all the following values are `nil`:
https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L193-L195

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			By("Watching Cluster node count to ensure it remains consistent")
+			Consistently(func() (bool, error) {
+				nodes, err := framework.GetNodes(client)
+				if err != nil {
+					return false, err
+				}
+				if len(nodes) != caMaxNodesTotal {
+					return false, nil
+				}
+				return true, nil
```
Danil-Grigorev, line 588: You could just do this here:
```golang
nodes, err := framework.GetNodes(client)
return len(nodes) == caMaxNodesTotal, err
``` 
It will assert the first value is true, and all the following values are `nil`:
https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L193-L195

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			By(fmt.Sprintf("Waiting for MachineSet %s replicas to scale down", transientMachineSet.GetName()))
+			Eventually(func() (bool, error) {
+				current, err := framework.GetMachineSet(client, transientMachineSet.GetName())
+				if err != nil {
+					return false, err
+				}
+				if pointer.Int32PtrDerefOr(current.Spec.Replicas, 0) != 1 {
+					return false, nil
+				}
+				return true, nil
```
Danil-Grigorev, line 606: You could just do this here:
```golang
current, err := framework.GetMachineSet(client, transientMachineSet.GetName())
return pointer.Int32PtrDerefOr(current.Spec.Replicas, 0) == 1, err
``` 
It will assert the first value is true, and all the following values are `nil`:
https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L193-L195

pkg/autoscaler/autoscaler.go
elmiko: i can change to this

pkg/autoscaler/autoscaler.go
```diff
@@ -732,6 +518,95 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
...
+			// that the cluster has reached a steady state and no more machines
+			// are in the process of being added.
+			By("Waiting for all Machines in MachineSet to enter Running phase")
+			framework.WaitForMachineSet(client, transientMachineSet.GetName())
+
+			// Now that the cluster has reached maximum size, we want to ensure
+			// that it doesn't try to grow larger.
+			By("Watching Cluster node count to ensure it remains consistent")
+			Consistently(func() (bool, error) {
+				nodes, err := framework.GetNodes(client)
```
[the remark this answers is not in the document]
elmiko: i marked this test as serial, it should never be run in parallel because one of the mechanisms we are testing is the max total nodes value.

pkg/autoscaler/autoscaler.go
elmiko: in this case we don't actually need to see the node disappear, just testing that the autoscaler has adjusted the replicas on the machineset is sufficient to detect that the scale down has happened to the desired level.

in the next stage of the test we will wait for the nodes to appear in the proper places, so it doesn't seem like we need to wait here.

elmiko:
updated with requested changes from Danil

Danil-Grigorev:
/hold 
Feel free to unhold once https://github.com/openshift/cluster-api-actuator-pkg/pull/183 will fix test flakes related to webhooks provisioning timeouts.

JoelSpeed:
/retest

@Danil-Grigorev The test failures here aren't related to webhook provisioning timeouts? They were related to other problems that Alberto and I have already fixed. Not sure why we need to hold this, can I unhold?

elmiko:
@Danil-Grigorev seems like this is good to go, can we remove the hold?

PR #176, opened 2020-07-17
OCPCLOUD-913 - Tests for webhook configurations provisioning by MAO

Danil-Grigorev:
Currently, we are managing webhooks from MAO, as CVO is not expected to handle resources with changing spec by another operator at the upgrade time. The implementation is covered with unit tests, but we need to ensure `cluster-actuator-pkg` contains e2e tests for such implementation too.

This PR is satisfying this requirement.


pkg/framework/webhooks.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// DeleteValidatingWebhookConfiguration deletes the specified ValidatingWebhookConfiguration object
+func DeleteValidatingWebhookConfiguration(c client.Client, webhookConfiguraiton *admissionregistrationv1.ValidatingWebhookConfiguration) error {
+	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+		if err := c.Delete(context.TODO(), webhookConfiguraiton); err != nil {
+			klog.Errorf("error querying api for ValidatingWebhookConfiguration object %q: %v, retrying...", webhookConfiguraiton.Name, err)
+			return false, nil
+		}
+		return true, nil
```
JoelSpeed, line 62: What about trying to delete a resource that doesn't exist? Should we account for that?

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// DeleteMutatingWebhookConfiguration deletes the specified MutatingWebhookConfiguration object
+func DeleteMutatingWebhookConfiguration(c client.Client, webhookConfiguraiton *admissionregistrationv1.MutatingWebhookConfiguration) error {
+	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+		if err := c.Delete(context.TODO(), webhookConfiguraiton); err != nil {
+			klog.Errorf("error querying api for MutatingWebhookConfiguration object %q: %v, retrying...", webhookConfiguraiton.Name, err)
+			return false, nil
+		}
+		return true, nil
```
JoelSpeed, line 73: What about trying to delete a resource that doesn't exist? Should we account for that?

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,149 @@
...
+		if err := c.Patch(context.TODO(), existing, client.MergeFrom(updated)); err != nil {
+			klog.Errorf("error patching ValidatingWebhookConfiguration object %q: %v, retrying...", updated.Name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
+// IsMutatingWebhookConfigurationMatched expects a matching MutatingWebhookConfiguration to be present in the cluster
+func IsMutatingWebhookConfigurationMatched(c client.Client) bool {
```
JoelSpeed, line 110: Maybe this should be called `Synced` rather than `Matched`, since we are checking if MAO has synced it to what we expect right?

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,149 @@
...
+		return equal, nil
+	}); err != nil {
+		klog.Errorf("Error waiting for match with expected MutatingWebhookConfigurationMatched: %v", err)
+		return false
+	}
+	return true
+}
+
+// IsValidatingWebhookConfigurationMatched expects a matching MutatingWebhookConfiguration to be present in the cluster
+func IsValidatingWebhookConfigurationMatched(c client.Client) bool {
```
JoelSpeed, line 131: Maybe this should be called `Synced` rather than `Matched`, since we are checking if MAO has synced it to what we expect right?

pkg/operators/machine-api-operator.go
```diff
@@ -65,6 +66,93 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
...
+
+		Expect(framework.IsMutatingWebhookConfigurationMatched(client)).To(BeTrue())
+	})
+
+	It("reconcile validating webhook configuration", func() {
+		client, err := framework.LoadClient()
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(framework.IsValidatingWebhookConfigurationMatched(client)).To(BeTrue())
+	})
```
JoelSpeed, line 82: One of these specifies the type, the other doesn't, should we make them both specify?

pkg/operators/machine-api-operator.go
```diff
@@ -65,6 +66,93 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
...
+
+		toUpdate := initial.DeepCopy()
+		for _, webhook := range toUpdate.Webhooks {
+			webhook.ClientConfig.CABundle = []byte("test")
+			webhook.AdmissionReviewVersions = append(webhook.AdmissionReviewVersions, "v1")
+		}
+
+		Expect(framework.UpdateMutatingWebhookConfiguration(client, toUpdate)).To(Succeed())
+
+		Expect(framework.IsMutatingWebhookConfigurationMatched(client)).To(BeTrue())
```
JoelSpeed, line 118: Surely after updating it can never match the default again, because there will be no CA Bundle in there?

pkg/operators/machine-api-operator.go
```diff
@@ -65,6 +66,93 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
...
+
+		toUpdate := initial.DeepCopy()
+		for _, webhook := range toUpdate.Webhooks {
+			webhook.ClientConfig.CABundle = []byte("test")
+			webhook.AdmissionReviewVersions = append(webhook.AdmissionReviewVersions, "v1")
+		}
+
+		Expect(framework.UpdateValidatingWebhookConfiguration(client, toUpdate)).To(Succeed())
+
+		Expect(framework.IsValidatingWebhookConfigurationMatched(client)).To(BeTrue())
```
JoelSpeed, line 145: Surely after updating it can never match the default again, because there will be no CA Bundle in there?

Danil-Grigorev: It's the place where `DeepDerivative` would still fit, as it will ignore differences with empty field on the default resource. So it will still match :)

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,154 @@
...
+// IsMutatingWebhookConfigurationSynced expects a matching MutatingWebhookConfiguration to be present in the cluster
+func IsMutatingWebhookConfigurationSynced(c client.Client) bool {
+	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
+		existing, err := GetMutatingWebhookConfiguration(c, DefaultMutatingWebhookConfiguration.Name)
+		if err != nil {
+			klog.Errorf("Error getting MutatingWebhookConfiguration: %v", err)
+			return false, nil
+		}
+
+		equal := equality.Semantic.DeepDerivative(DefaultMutatingWebhookConfiguration.Webhooks, existing.Webhooks)
```
JoelSpeed, line 123: Might be worth a note that this will should the CABundle field that has been set by the injector and that's why we are using `DeepDerivative`

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,154 @@
...
+// IsValidatingWebhookConfigurationSynced expects a matching MutatingWebhookConfiguration to be present in the cluster
+func IsValidatingWebhookConfigurationSynced(c client.Client) bool {
+	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
+		existing, err := GetValidatingWebhookConfiguration(c, DefaultValidatingWebhookConfiguration.Name)
+		if err != nil {
+			klog.Errorf("Error getting MutatingWebhookConfiguration: %v", err)
+			return false, nil
+		}
+
+		equal := equality.Semantic.DeepDerivative(DefaultValidatingWebhookConfiguration.Webhooks, existing.Webhooks)
```
JoelSpeed, line 144: Might be worth a note that this will should the CABundle field that has been set by the injector and that's why we are using `DeepDerivative`

Danil-Grigorev:
@JoelSpeed I added a small comment on the `DeepDerivative` usage.

JoelSpeed:
/approve

Thanks!

pkg/framework/webhooks.go
```diff
@@ -0,0 +1,160 @@
...
+			klog.Errorf("error querying api for MutatingWebhookConfiguration object %q: %v, retrying...", webhookConfiguraiton.Name, err)
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
+// UpdateMutatingWebhookConfiguration updates the specified mutating webhook configuration
+func UpdateMutatingWebhookConfiguration(c client.Client, updated *admissionregistrationv1.MutatingWebhookConfiguration) error {
+	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
```
enxebre, line 84: why `WaitMedium` here?
this is `GetMutatingWebhookConfiguration` and so nesting `wait.PollImmediate` calls.

I think the only purpose of using `wait.PollImmediate` in this library funcs is to protect and don't fail the test on a transient network issue? to that end we could may be just wait a few seconds? wdyt?

Danil-Grigorev: Sounds good to me. Webhooks are ready to react as soon as they are updated in the cluster. I could see potential to use `apply` method from library-ho, but short wait should be enough in any case.

Danil-Grigorev: @enxebre I decreased both wait periods to be at `1 min` at max, in case of the client throttling appears, and `3 min` for waiting on webhook configuration to be synced, PTAL

enxebre:
thanks, not a blocker for this PR but my point is that I think even WaitShort is too much here. This are not eventually consistent functions. This are atomic actions, the only reason to wait.poll here, if any, is protection against network transient issues. For that a shorter e.g 5s wait period would be reasonable. WDYT?
/lgtm

PR #178, opened 2020-07-20
Add webhook test for machineSet/machine clusterID label machine.openshift.io/cluster-api-cluster

enxebre:
With #608 we dropped the burden from the user to set the clusterID label on machines.
As elaborated in #608 (comment) the motivation is that this is an implementation detail that users shouldn't care about.

However as the labels are used by machineSet to determine ownership, the change introduced above might result in edge scenarios where the machineSet and machine label has a different value. This would result in machines going orphan and the machineSet recreating new instances. Bad.

Therefore we choose now to remove the burden from users by enforcing the label value via webhhooks and keeping the old behaviour in the backend to avoid any chance of breaking existing environments where bad input might have been set as in https://bugzilla.redhat.com/show_bug.cgi?id=1857175. This should be fixed by https://github.com/openshift/machine-api-operator/pull/644 and this PR validates that scenario.

### DIFF
--- a/pkg/infra/webhooks.go
+++ b/pkg/infra/webhooks.go
@@ -15,6 +15,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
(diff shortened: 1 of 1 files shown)


enxebre:
>Also, are we validating that people can't change the labels later via webhooks? I think we should probably be doing so otherwise users would be able to update the labels to make them invalid after the creation

They can do that. Same that they can do it today with out this. Default Webhooks are only enable on creation. Enabling on update will have multiple other implications that will need to be analysed separately.

PR #180, opened 2020-07-22
Degraded validation

enxebre:
The first commit introduces a validation for the machine-api clusterOperator going degraded.
The second commit renames IsStatusAvailable -> WaitForStatusAvailable and increase the waiting timeout. This is needed as we are testing status transitions now. Names for funcs which run actions in loop should be prefixed with wait.
So if the specific action need to be called within any other "eventually" func it can be name appriopriately e.g IsStatusAvailable.
The timeout is increased so this can succeed for status transitions. E.g the mao will wait for a pod to be running for at least 3min before considering it available.
https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L28

This PR is to test https://github.com/openshift/machine-api-operator/pull/651

### DIFF
--- a/pkg/framework/framework.go
+++ b/pkg/framework/framework.go
@@ -120,14 +120,14 @@ func LoadClientset() (*kubernetes.Clientset, error) {
 	return kubernetes.NewForConfig(cfg)
 }
 
-func IsStatusAvailable(client runtimeclient.Client, name string) bool {
+func WaitForStatusAvailable(client runtimeclient.Client, name string) bool {
 	key := types.NamespacedName{
 		Namespace: MachineAPINamespace,
 		Name:      name,
 	}
 	clusterOperator := &configv1.ClusterOperator{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, 10*time.Minute, func() (bool, error) {
 		if err := client.Get(context.TODO(), key, clusterOperator); err != nil {
 			klog.Errorf("error querying api for OperatorStatus object: %v, retrying...", err)
 			return false, nil
--- a/pkg/framework/framework.go
+++ b/pkg/framework/framework.go
@@ -152,6 +152,27 @@ func IsStatusAvailable(client runtimeclient.Client, name string) bool {
 	return true
 }
 
+func IsStatusDegraded(client runtimeclient.Client, name string) (bool, error) {
+	key := types.NamespacedName{
+		Namespace: MachineAPINamespace,
+		Name:      name,
+	}
+	clusterOperator := &configv1.ClusterOperator{}
+	if err := client.Get(context.TODO(), key, clusterOperator); err != nil {
+		return false, fmt.Errorf("error querying api for OperatorStatus object: %v, retrying...", err)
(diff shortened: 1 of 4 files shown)


pkg/operators/machine-api-operator.go
```diff
@@ -75,4 +79,45 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
...
+			// get machine API controllers pods
+			podList := &v1.PodList{}
+			if err := c.List(context.TODO(), podList, client.MatchingLabels(mAPIControllersLabels)); err != nil {
+				klog.Errorf("failed to list pods: %v", err)
+				return false, nil
+			}
+			if len(podList.Items) < 1 {
+				klog.Errorf("list of pods is empty")
+				return false, nil
+			}
```
JoelSpeed, line 103: Does this matter? If the list of pods is empty, status should be degraded right?

pkg/framework/framework.go
```diff
@@ -120,14 +120,14 @@ func LoadClientset() (*kubernetes.Clientset, error) {
...
-func IsStatusAvailable(client runtimeclient.Client, name string) bool {
+func WaitForStatusAvailable(client runtimeclient.Client, name string) bool {
 	key := types.NamespacedName{
 		Namespace: MachineAPINamespace,
 		Name:      name,
 	}
 	clusterOperator := &configv1.ClusterOperator{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, 10*time.Minute, func() (bool, error) {
```
JoelSpeed, line 130: Commit description says waits for 3 minutes, which is `WaitMedium`, but it's actually waiting for 10 minutes, should we update one of these? I think 3 minutes to become available should be sufficient shouldn't it?
```suggestion
	if err := wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
```

pkg/operators/machine-api-operator.go
alexander-demichev: I think it doesn't matter, but do we expect pods not to be running when at this part of the test?

JoelSpeed: Well my thought was we go through this cycle, delete a pod, status isn't degraded yet, so start again, pod has not been recreated yet, could the status be degraded by this point? I am not sure, I would assume they are independent since they're different controllers controlling these actions

pkg/framework/framework.go
enxebre: 3 min is the minimum time the mao expects the pod to have been available https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L28

5 min is the total the mao waits for the pod to rollout out https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L30

10 min is where we set the bar in our tests to consider this a failure. I set 10 to account for scenarios where this is running in parallel with tests that are disrupting the status.

I'm ok to put any lower (>5m) if you prefer.

JoelSpeed: Nope, that makes sense, happy to leave as 10 mins now commit description is updated

PR #182, opened 2020-08-02
Fail fast if machines enter Failed phase

Danil-Grigorev:
According to MAO specifics, machines entering `Failed` phase will no longer be provisioned or processed. This commit ensures that `e2e` tests requiring machines to enter `Running` phase, or waiting for a full `MachineSet` rollout, would fail faster and give more readable failure output, instead of a generic timeout on condition.

This will help improve CI readability.

### DIFF
--- a/pkg/framework/machines.go
+++ b/pkg/framework/machines.go
@@ -20,17 +20,23 @@ import (
 )
 
 // FilterRunningMachines returns a slice of only those Machines in the input
-// that are in the "Running" phase.
-func FilterRunningMachines(machines []*mapiv1beta1.Machine) []*mapiv1beta1.Machine {
+// that are in the "Running" phase. If some of the machine enters Failed phase,
+// it will indicate this wil an error.
+func FilterRunningMachines(machines []*mapiv1beta1.Machine) ([]*mapiv1beta1.Machine, error) {
 	var result []*mapiv1beta1.Machine
 
 	for i, m := range machines {
-		if m.Status.Phase != nil && *m.Status.Phase == MachinePhaseRunning {
-			result = append(result, machines[i])
+		if m.Status.Phase != nil {
+			switch *m.Status.Phase {
+			case MachinePhaseRunning:
+				result = append(result, machines[i])
+			case MachinePhaseFailed:
+				return nil, fmt.Errorf("Machine entered the Failed phase: %q, reason: %v", m.GetName(), m.Status.ErrorMessage)
+			}
 		}
 	}
 
-	return result
+	return result, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/framework/machines.go
```diff
@@ -139,13 +145,19 @@ func DeleteMachines(client runtimeclient.Client, machines ...*mapiv1beta1.Machin
...
 				Name:      m.GetName(),
 				Namespace: m.GetNamespace(),
-			}, &mapiv1beta1.Machine{})
+			}, machine)
 			if !apierrors.IsNotFound(err) {
 				return false // Not deleted, or other error.
 			}
+			// Fail fast if machine entered Failed phase
+			if machine.Status.Phase != nil {
+				Expect(*machine.Status.Phase).NotTo(Equal(MachinePhaseFailed),
```
enxebre, line 158: would this break the eventually loop? we still want to WaitForMachinesDeleted

Danil-Grigorev: The failed machine will never be deleted, as it is not going to reconcile. It should break the loop and report an error, unlike returning an error inside the block.

JoelSpeed: A failed machine still gets reconciled by the deletion logic does it not? The machine failed check is after the deletion timestamp check. As far as I can tell a failed machine can still be deleted, it will likely just be quicker.

https://github.com/openshift/machine-api-operator/blob/487298d74beeadfe564c5a7db6b9e37296b641b9/pkg/controller/machine/controller.go#L265-L268

pkg/framework/machines.go
```diff
@@ -20,17 +20,23 @@ import (
...
 	for i, m := range machines {
-		if m.Status.Phase != nil && *m.Status.Phase == MachinePhaseRunning {
-			result = append(result, machines[i])
+		if m.Status.Phase != nil {
+			switch *m.Status.Phase {
+			case MachinePhaseRunning:
+				result = append(result, machines[i])
+			case MachinePhaseFailed:
+				return nil, fmt.Errorf("Machine entered the Failed phase: %q, reason: %v", m.GetName(), m.Status.ErrorMessage)
+			}
```
JoelSpeed, line 35: One question that comes to mind is whether any of the tests rely on a Machine going into a failed phase, we need to make sure we aren't breaking those by doing this

Danil-Grigorev: We are not testing that at the moment, so it is not a breaking change. Why wait for machines running if you are expecting them not to?

JoelSpeed: Good question 😅  I just wondered if we were using this to check that a machine didn't become running. I think it's ok

PR #183, opened 2020-08-02
Set 10 minute timeout on webhook and deployment operations

Danil-Grigorev:
Webhooks may take up to 10 minutes to sync. When changed we first have to wait for a full `Deployment`
rollout (5 minutes) and `DaemonSet` rollout (another 5 minutes) to occur before they are synced. Then the
process starts all over again. This applies to all webhook operations. For simplicity, the same applies
onto `machine-api-controllers` `Deployment`, which may take up to 5 minutes to successfully reconcile, or
even be re-provisioned after deletion.

This PR is a response to multiple CI failures:
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1290284947541594112  
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/1290284947608702976
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/180/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1286272804777365504
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/180/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/1286272804827697152
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/174/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1291923580057030656

### DIFF
--- a/pkg/framework/deployment.go
+++ b/pkg/framework/deployment.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	kappsapi "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -12,15 +13,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetDeployment gets deployment object by name and namespace.
-func GetDeployment(c client.Client, name, namespace string) (*kappsapi.Deployment, error) {
+// GetDeployment gets deployment object by name and namespace with custom timeout
+func GetDeployment(c client.Client, name, namespace string, timeout time.Duration) (*kappsapi.Deployment, error) {
 	key := types.NamespacedName{
 		Namespace: namespace,
 		Name:      name,
 	}
 	d := &kappsapi.Deployment{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, timeout, func() (bool, error) {
 		if err := c.Get(context.TODO(), key, d); err != nil {
 			klog.Errorf("Error querying api for Deployment object %q: %v, retrying...", name, err)
 			return false, nil
@@ -32,9 +33,9 @@ func GetDeployment(c client.Client, name, namespace string) (*kappsapi.Deploymen
 	return d, nil
 }
 
-// DeleteDeployment deletes the specified deployment
-func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
-	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+// DeleteDeployment deletes the specified deployment with custom timeout
+func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment, timeout time.Duration) error {
+	return wait.PollImmediate(RetryShort, timeout, func() (bool, error) {
 		if err := c.Delete(context.TODO(), deployment); err != nil {
 			klog.Errorf("error querying api for deployment object %q: %v, retrying...", deployment.Name, err)
 			return false, nil
@@ -43,10 +44,11 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
 	})
 }
 
-// UpdateDeployment updates the specified deployment
-func UpdateDeployment(c client.Client, name, namespace string, updated *kappsapi.Deployment) error {
-	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
-		d, err := GetDeployment(c, name, namespace)
+// UpdateDeployment updates the specified deployment with custom timeout
+func UpdateDeployment(c client.Client, name, namespace string,
+	updated *kappsapi.Deployment, timeout time.Duration) error {
+	return wait.PollImmediate(RetryShort, timeout, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace, WaitShort)
 		if err != nil {
 			klog.Errorf("Error getting deployment: %v", err)
 			return false, nil
@@ -59,10 +61,10 @@ func UpdateDeployment(c client.Client, name, namespace string, updated *kappsapi
 	})
 }
 
-// IsDeploymentAvailable returns true if the deployment has one or more availabe replicas
-func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
-	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
-		d, err := GetDeployment(c, name, namespace)
+// IsDeploymentAvailable returns true if the deployment has one or more availabe replicas with custom timeout
+func IsDeploymentAvailable(c client.Client, name, namespace string, timeout time.Duration) bool {
+	if err := wait.PollImmediate(RetryShort, timeout, func() (bool, error) {
+		d, err := GetDeployment(c, name, namespace, WaitShort)
 		if err != nil {
 			klog.Errorf("Error getting deployment: %v", err)
 			return false, nil
@@ -82,9 +84,11 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
 	return true
 }
 
-// IsDeploymentSynced returns true if provided deployment spec matched one found on cluster
-func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment, name, namespace string) bool {
-	d, err := GetDeployment(c, name, namespace)
+// IsDeploymentSynced returns true if provided deployment spec
+// matched one found on cluster with custom timeout
+func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment,
+	name, namespace string, timeout time.Duration) bool {
+	d, err := GetDeployment(c, name, namespace, timeout)
 	if err != nil {
 		klog.Errorf("Error getting deployment: %v", err)
 		return false
--- a/pkg/framework/webhooks.go
+++ b/pkg/framework/webhooks.go
@@ -3,6 +3,7 @@ package framework
 import (
 	"context"
 	"fmt"
+	"time"
 
 	mapiv1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
@@ -19,12 +20,18 @@ var DefaultValidatingWebhookConfiguration = mapiv1.NewValidatingWebhookConfigura
 // DefaultMutatingWebhookConfiguration is a default mutating webhook configuration resource provided by MAO
 var DefaultMutatingWebhookConfiguration = mapiv1.NewMutatingWebhookConfiguration()
 
+// Webhooks may take up to 10 minutes to sync. When changed we first have to wait for a full Deployment
+// rollout (5 minutes) and DaemonSet rollout (another 5 minutes) to occur before they are synced. Then the
+// process starts all over again. This applies to all webhook operations.
+// TODO: Make MAO resource state management unblocking
+var waitWebhooks = 10 * time.Minute
+
 // GetMutatingWebhookConfiguration gets MutatingWebhookConfiguration object by name
 func GetMutatingWebhookConfiguration(c client.Client, name string) (*admissionregistrationv1.MutatingWebhookConfiguration, error) {
 	key := client.ObjectKey{Name: name}
 	existing := &admissionregistrationv1.MutatingWebhookConfiguration{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, waitWebhooks, func() (bool, error) {
 		if err := c.Get(context.TODO(), key, existing); err != nil {
 			klog.Errorf("Error querying api for MutatingWebhookConfiguration object %q: %v, retrying...", name, err)
 			return false, nil
@@ -41,7 +48,7 @@ func GetValidatingWebhookConfiguration(c client.Client, name string) (*admission
 	key := client.ObjectKey{Name: name}
 	existing := &admissionregistrationv1.ValidatingWebhookConfiguration{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryShort, waitWebhooks, func() (bool, error) {
 		if err := c.Get(context.TODO(), key, existing); err != nil {
 			klog.Errorf("Error querying api for ValidatingWebhookConfiguration object %q: %v, retrying...", name, err)
 			return false, nil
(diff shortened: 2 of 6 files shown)


pkg/framework/webhooks.go
```diff
@@ -81,7 +81,7 @@ func DeleteMutatingWebhookConfiguration(c client.Client, webhookConfiguraiton *a
 
 // UpdateMutatingWebhookConfiguration updates the specified mutating webhook configuration
 func UpdateMutatingWebhookConfiguration(c client.Client, updated *admissionregistrationv1.MutatingWebhookConfiguration) error {
-	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
```
enxebre, line 84: can you elaborate what's motivating this change? can you share CI failures links that support that motivation and that this change would mitigate?

pkg/framework/webhooks.go
```diff
@@ -97,7 +97,7 @@ func UpdateMutatingWebhookConfiguration(c client.Client, updated *admissionregis
 
 // UpdateValidatingWebhookConfiguration updates the specified mutating webhook configuration
 func UpdateValidatingWebhookConfiguration(c client.Client, updated *admissionregistrationv1.ValidatingWebhookConfiguration) error {
-	return wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
```
enxebre, line 100: can you elaborate what's motivating this change? can you share CI failures links that support that motivation and that this change would mitigate?

pkg/operators/machine-api-operator.go
```diff
@@ -14,7 +14,7 @@ var (
 	maoManagedDeployment = "machine-api-controllers"
 )
 
-var _ = Describe("[Feature:Operators] Machine API operator deployment should", func() {
+var _ = Describe("[Feature:Operators] Machine API operator deployment should [Serial]", func() {
```
JoelSpeed, line 17: Does naming these as `[serial]` make them run as serial?

enxebre, line 17: is this something ginkgo just understand? can you reference the docs?

Danil-Grigorev: @enxebre @JoelSpeed You are right, not at the moment. Here is a separate PR for serial support: https://github.com/openshift/cluster-api-actuator-pkg/pull/186

pkg/framework/webhooks.go
Danil-Grigorev: I can't but shortly the workflow I imagine:webhook is deleted,

pkg/framework/webhooks.go
Danil-Grigorev: https://github.com/openshift/cluster-api-actuator-pkg/pull/183#discussion_r465690265

pkg/operators/machine-api-operator.go
elmiko: not, specifically. these are labels inspired from the kubernetes community[0]

[0] https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md#kinds-of-tests

elmiko: these labels are inspired by the kubernetes community docs[0]. ginkgo doesn't know them, but i have started to clean them up in the autoscaler tests as i think it helps people to understand how to operate the tests.

additionally, as demonstrated in Danil's pr, they can be used to filter the tests.

JoelSpeed: Out of interest, do you have any estimate how much time this might add to each test run by making these serial? Do you think there are any tests within here that can still run in parallel? We could maybe divide them into parallel and serial sets

pkg/framework/webhooks.go
enxebre: >https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/183/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1290937797384867840

mm that link happened running against this PR... And the failure is for `API operator deployment should [Serial] maintains deployment spec`, so I'm not really sure how it's related to this change at all.

Those links are timeouts, if they are match the something went wrong with the operator, none test will pass. Some time to run the pod + the 3 min to check permanent availability should be the norm.

```UpdateValidatingWebhookConfiguration``` are atomic functions, the only reason they have a PollImmediate atm is to mitigate transient errors and should probably top doing that, assume there won't be any and if there's we fail the test. If something needs to wait for a good reason please create a `WaitBlah` function.

Comment (Danil-Grigorev) [reply pkg/framework/webhooks.go:100]: That's why enabling `Serial` keyword is necessary here. It failed on this line, and that's the case I described:

https://github.com/openshift/cluster-api-actuator-pkg/blob/e943d397dea02e9c3f09cb1dc6543cf4a0303da0/pkg/operators/machine-api-operator.go#L60

> But shortly the workflow I imagine is happening:
> 
>     * tests are running in parallel
> 
>     * webhook is deleted
> 
>     * MAO is busy on other things, and does not yet know (1-3 minutest), with the inclusion of [openshift/machine-api-operator#651](https://github.com/openshift/machine-api-operator/pull/651) it is now up to 10 minutes (https://github.com/openshift/machine-api-operator/blob/6e867bf92c3ec296c4c93d73abd93b47af9426a3/pkg/operator/sync.go#L31 + https://github.com/openshift/machine-api-operator/blob/6e867bf92c3ec296c4c93d73abd93b47af9426a3/pkg/operator/sync.go#L28)

Comment (Danil-Grigorev) [reply pkg/framework/webhooks.go:100]: In order to enable this, we need to merge https://github.com/openshift/cluster-api-actuator-pkg/pull/186 first.

File: pkg/operators/machine-api-operator.go [hunk 4]
```diff
@@ -66,6 +66,10 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
 		Expect(framework.IsDeploymentAvailable(client, maoManagedDeployment, framework.MachineAPINamespace)).To(BeTrue())
 
 	})
+})
+
```
enxebre, line 70: can you help me understand what this is solving? All the IT validations within this describe still run in parallel right?

enxebre:
>In a typical environment MAO resource provisioning
occurs only during upgrades and serves as a one-time procedure.
Frequent manipulation with provisioned resources by the user is not expected.

Anything could manipulate mao owned resources and mao still should be able to react to that appropriately. I think this tests should run in parallel and we use `waitForXtoSync` funcs.

pkg/operators/machine-api-operator.go
enxebre: Can you share CI failures/flakes this PR would solve?

JoelSpeed:
The tests have passed once for the latest commit, since the primary purpose of this PR is to improve stability, we need to verify that it will repeatedly pass

/test e2e-aws-operator
/test e2e-azure-operator
/test e2e-gcp-operator

Danil-Grigorev:
Those PRs are failing CI without this change:
- https://github.com/openshift/cluster-api-actuator-pkg/pull/184
- https://github.com/openshift/cluster-api-actuator-pkg/pull/182
- https://github.com/openshift/cluster-api-actuator-pkg/pull/180

pkg/operators/machine-api-operator.go
Danil-Grigorev: - https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1290284947541594112  
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/1290284947608702976
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/180/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1286272804777365504
- Guess this one as well: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/180/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-gcp-operator/1286272804827697152

enxebre:
>Those PRs are failing CI without this change:

It'd be better to link the failure jobs rather than PR, otherwise the links will became meaningless in time.

Can you please answer this https://github.com/openshift/cluster-api-actuator-pkg/pull/183#discussion_r466275006

For https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1290284947541594112 have you dug into the details wouldn't we just to increase the timeout for `waitToSync` as per in https://github.com/openshift/cluster-api-actuator-pkg/pull/183#issuecomment-669818711? That's current mao behaviour and so we want to test it, we don't want to relax tests.

Danil-Grigorev:
> > Those PRs are failing CI without this change:
> 
> It'd be better to link the failure jobs rather than PR, otherwise the links will became meaningless in time.
> 
> Can you please answer this [#183 (comment)](https://github.com/openshift/cluster-api-actuator-pkg/pull/183#discussion_r466275006)
> 
> For https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/184/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1290284947541594112 have you dug into the details wouldn't we just to increase the timeout for `waitToSync` as per in [#183 (comment)](https://github.com/openshift/cluster-api-actuator-pkg/pull/183#issuecomment-669818711)? That's current mao behaviour and so we want to test it, we don't want to relax tests.

Here are some CI jobs: https://github.com/openshift/cluster-api-actuator-pkg/pull/183#discussion_r466384937
Here is the reason: https://github.com/openshift/cluster-api-actuator-pkg/pull/183#discussion_r465690265 The places where the `waitToSync` may still fail, are the short lasting atomic operations like `updateWebhooks` as the timeout there should be ~ `>=5 minutes` to work all the time.

https://github.com/openshift/cluster-api-actuator-pkg/pull/183#issuecomment-669818711 - worth a BZ to work out. Reaction time is sometimes really slow.

Danil-Grigorev:
Reason why those tests are still failing:

pkg/framework/webhooks.go
```diff
@@ -19,12 +20,18 @@ var DefaultValidatingWebhookConfiguration = mapiv1.NewValidatingWebhookConfigura
 // DefaultMutatingWebhookConfiguration is a default mutating webhook configuration resource provided by MAO
 var DefaultMutatingWebhookConfiguration = mapiv1.NewMutatingWebhookConfiguration()
 
+// Webhooks may take up to 10 minutes to reconcile, as they could be just changed, but before we know about that,
+// a full Deployment (5 minutes) rollout and DaemonSet (another 5 minutes) rollout should occure. Then the
```
JoelSpeed, line 24: I think this will read better like this, wdyt?
```suggestion
// Webhooks may take up to 10 minutes to sync. When changed we first have to wait for a full Deployment
// rollout (5 minutes) and DaemonSet rollout (another 5 minutes) to occur before they are synced. Then the
```

pkg/framework/deployment.go
```diff
@@ -34,7 +40,12 @@ func GetDeployment(c client.Client, name, namespace string) (*kappsapi.Deploymen
 
 // DeleteDeployment deletes the specified deployment
 func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
```
enxebre, line 42: why do we need to wrap this? now `DeleteDeployment` is unused. Can't we just have a single `waitForDerploymentDeletion`

pkg/framework/deployment.go
```diff
@@ -45,7 +56,13 @@ func DeleteDeployment(c client.Client, deployment *kappsapi.Deployment) error {
 
 // UpdateDeployment updates the specified deployment
 func UpdateDeployment(c client.Client, name, namespace string, updated *kappsapi.Deployment) error {
-	return wait.PollImmediate(RetryShort, WaitMedium, func() (bool, error) {
```
enxebre, line 48: why do we need to wrap this? now `UpdateDeployment ` is unused. Can't we just have a single `waitForDeploymentUpdate`?

pkg/framework/deployment.go
```diff
@@ -61,7 +78,12 @@ func UpdateDeployment(c client.Client, name, namespace string, updated *kappsapi
 
 // IsDeploymentAvailable returns true if the deployment has one or more availabe replicas
 func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
-	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
+	return IsDeploymentAvailableIn(c, name, namespace, WaitLong)
```
enxebre, line 81: why do we need to wrap this and have two different functions? is there any valid case where we want to use different wait timeout here?

pkg/framework/deployment.go
```diff
@@ -84,7 +106,14 @@ func IsDeploymentAvailable(c client.Client, name, namespace string) bool {
 
 // IsDeploymentSynced returns true if provided deployment spec matched one found on cluster
 func IsDeploymentSynced(c client.Client, dep *kappsapi.Deployment, name, namespace string) bool {
```
enxebre, line 108: why this wrap? is IsDeploymentSynced is unused now.

Danil-Grigorev:
> I have no idea what this PR is trying to do. It either needs to link to a BZ with great detail, or the first comment and commit message describe exactly the conditions we're trying to resolve. Making some tests run in serial, I have no idea why we're doing that, I'm not going to try to determine the context from all the comments and code changes. We need clear commit messages describing the purpose of things.

Sorry, the PR description was outdated. Right now the PR is simply increasing timeouts on MAO operations, as multiple CI jobs failed recently due to that. I added the links, and an explanation, why it will take 10 minutes for each to complete.

JoelSpeed:
@Danil-Grigorev Did you see https://github.com/openshift/cluster-api-actuator-pkg/pull/183#pullrequestreview-463171879? Don't think I got a response but I think it's still relevant

Thanks for updating the PR description and adding those links, will help us work out why we did this 6 months down the line when we are doing some archeology I'm sure! Nit, the 4th link seems to be an unrelated failure, PTAL

I wonder if we should try, as an alternative (or maybe compliment) to this, refactoring the MAO so that it applies the Deployment, Daemonset and webhooks, and then performs the waiting process. This would reduce the time waiting since the deployment and daemonset would be rolling out in parallel. If we try this idea first, we may not even need this PR right? The tests may consistently pass in the allowed time and we have done some optimisation to the product as well :D

michaelgugino:
> to this, refactoring the MAO so that it applies the Deployment, Daemonset and webhooks, and then performs the waiting process. This would reduce the time waiting since the deployment and daemonset would be rolling out in parallel. If we try this idea first, we may not even need this PR right? The tests may consistently pass in the allowed time and we have done some optimisation to the product as well :D

I endorse this suggestion.

Danil-Grigorev:
/close

This is no longer relevant, as the issue with webhooks was resolved by https://github.com/openshift/machine-api-operator/pull/707 and later substituted with upstream implementation - https://github.com/openshift/machine-api-operator/pull/642

PR #184, opened 2020-08-03
Disable spot-termination mock metadata test

JoelSpeed:
This test seems to be broken, need to unblock other work so going to  disable, make sure everything else is good, work out why it is broken and re-enable when fixed.


JoelSpeed:
We may not need to do this if https://github.com/openshift/cluster-api-actuator-pkg/pull/185 works (which I believe it does)

enxebre:
@JoelSpeed do you think we can close this one?

JoelSpeed:
Yeah this was fixed in #185

PR #186, opened 2020-08-05
Enable serial keyword to run tests serially

Danil-Grigorev:
Adding `[Serial]` keyword to the test description will mark the tests for serial execution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,10 +63,11 @@ test-e2e: ## Run openshift specific e2e test
 	# failures and flakes.
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -focus="Feature:Operators" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -focus="Feature:Operators" -skip="Serial" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler|Serial" || (hack/junitmerge.sh && exit 1)
 	# TODO: parallelise autoscaler
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" || (hack/junitmerge.sh && exit 1)
+	# Run serial tests only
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,10 +63,11 @@ test-e2e: ## Run openshift specific e2e test
...
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -focus="Feature:Operators" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -focus="Feature:Operators" -skip="Serial" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler|Serial" || (hack/junitmerge.sh && exit 1)
 	# TODO: parallelise autoscaler
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" || (hack/junitmerge.sh && exit 1)
+	# Run serial tests only
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler|Serial" || (hack/junitmerge.sh && exit 1)
```
elmiko, line 70: i think we should leave the autoscaler test line the way it was, those tests need more cleanup before we can start running them in parallel.

Danil-Grigorev: This could gradually increase overall CI running time, I left the `TODO` as a reminder to do it later.

Danil-Grigorev: Oh, they are both serial... Ok, I'll change.

JoelSpeed:
/test e2e-aws-operator
/test e2e-azure-operator
/test e2e-gcp-operator

As far as I can tell, at the moment this PR is actually a no-op, it's not changing the behaviour of running the tests because we don't have any labelled as `serial`. All of the operator tests passed against this PR already, which makes me hesitant, do we actually need this? Let's try and run the tests here a few times to see if they are flaking on this

Makefile
JoelSpeed, line 70: One concern is that we could have a test named with `Serial` in it that may not be intended to be serial in this sense, eg `Serialize content of foo`, at the moment, this would match that, should we instead update these to be more explicit around matching with `[]` as well? eg. `--skip="[Serial]"`, this would leave no ambiguity about the intention

Danil-Grigorev: Was thinking about that, let me check how the regex will react on the change.

enxebre:
this PR has no impact by its own. We can't validate it does what it intends and It's difficult to understand the motivation.
Can we please put this as commit here https://github.com/openshift/cluster-api-actuator-pkg/pull/183 and review there.
Then I'd suggest we just use “Machine API operator deployment should” waitForXtoSync functions.

PR #189, opened 2020-09-22
Bug 1876680: Temporarily disable webhook validation

enxebre:
We will enable this after openshift/machine-api-operator#697 get merged.
Alternative for github.com/openshift/cluster-api-actuator-pkg/pull/187 to investigate CI failures.

PR #190, opened 2020-10-02
Re-enable webhook validation

JoelSpeed:
Reverts openshift/cluster-api-actuator-pkg#189

Webhooks changes have been merged, we should re-enable the validation to prove the webhooks are performing as expected


alexander-demichev (approved):
/approve

PR #192, opened 2020-10-09
Bug 1906732: Add E2E to ensure Cluster Wide Proxies are honoured in Machine API components

rvanderp3:


pkg/operators/machine-api-operator.go
```diff
@@ -164,3 +166,72 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
...
+		client, err := framework.LoadClient()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(framework.IsProxyInjectedForDeployment(client, maoManagedDeployment, framework.MachineAPINamespace, false)).To(BeTrue())
+	})
+
+	It("tear down cluster proxy", func() {
+		client, err := framework.LoadClient()
+		Expect(err).NotTo(HaveOccurred())
+		framework.DestroyClusterProxy(client)
+	})
```
JoelSpeed, line 235: I think there's some confusion here over how the test framework ginkgo works. Each `It` constitutes a different test case, these look like steps in a single test case?

If so, make each of these `It`s a `By` and wrap them all in an `It`

eg

```
It("should be able to create a machineSet when a proxy is configured", func() {
  By("deploy proxy", func() {
    ... 
  })

  By("configure proxy", func() {
    ... 
  })
})

File: pkg/operators/machine-api-operator.go [hunk 2]
```diff
@@ -164,3 +166,72 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
 		Expect(framework.IsStatusAvailable(client, "machine-api")).To(BeTrue())
 	})
 })
+
+var _ = Describe("[Feature:Operators] When cluster-wide proxy is configured, Machine API cluster operator should ", func() {
+	defer GinkgoRecover()
```
JoelSpeed, line 171: Shouldn't need GinkgoRecover unless you have an go routines in your tests, which I can't see

pkg/operators/machine-api-operator.go
```diff
@@ -1,11 +1,13 @@
 package operators
 
 import (
+	"context"
 	"fmt"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	v1 "github.com/openshift/api/config/v1"
```
JoelSpeed, line 8: We normally import this as `configv1` throughout this codebase IIRC

pkg/framework/proxies.go
```diff
@@ -0,0 +1,196 @@
...
+	DeleteDeployment(c, &kappsapi.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "squid-proxy",
+			Namespace: "default",
+		},
+	})
+}
+
+// IsProxyInjectedForDeployment checks to see if containers in the deployment define proxy variables
+func IsProxyInjectedForDeployment(c runtimeclient.Client, name, namespace string, shouldBePresent bool) bool {
```
JoelSpeed, line 123: Since this is a wait function, it should probably include Wait in the name, eg `WaitForProxyInjectedInDeployment`?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,196 @@
...
+			Namespace: "default",
+		},
+		Data: map[string]string{
+			"squid.conf": "acl localnet src 192.168.0.0/16\t# RFC1918 possible internal network\nacl crcnet src 10.88.0.0/16\t# RFC1918 possible internal network\nacl SSL_ports port 443\nacl Safe_ports port 80\t\t# http\nacl Safe_ports port 21\t\t# ftp\nacl Safe_ports port 443\t\t# https\nacl Safe_ports port 70\t\t# gopher\nacl Safe_ports port 210\t\t# wais\nacl Safe_ports port 1025-65535\t# unregistered ports\nacl Safe_ports port 280\t\t# http-mgmt\nacl Safe_ports port 488\t\t# gss-http\nacl Safe_ports port 591\t\t# filemaker\nacl Safe_ports port 777\t\t# multiling http\nacl CONNECT method CONNECT\nhttp_access deny !Safe_ports\nhttp_access deny CONNECT !SSL_ports\nhttp_access allow localhost manager\nhttp_access deny manager\nacl mysubnet src 192.168.0.0/16\nhttp_access allow localhost localnet crcnet\nhttp_access allow all\nhttp_port 3128\ncoredump_dir /var/spool/squid\nrefresh_pattern ^ftp:\t\t1440\t20%\t10080\nrefresh_pattern ^gopher:\t1440\t0%\t1440\nrefresh_pattern -i (/cgi-bin/|\\?) 0\t0%\t0\nrefresh_pattern (Release|Packages(.gz)*)$      0       20%     2880\nrefresh_pattern .\t\t0\t20%\t4320",
```
JoelSpeed, line 23: This config should be moved to a constant, ideally a multiline constant so that we don't have to have all of the escape characters in it, can you update please?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,196 @@
+package framework
+
+import (
+	"context"
+	"errors"
+	configv1 "github.com/openshift/api/config/v1"
+	kappsapi "k8s.io/api/apps/v1"
```
JoelSpeed, line 7: I notice this isn't imported as `appsv1` in this package, but it is in the rest of the codebases and this is the convention, would you mind updating to `appsv1`?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,196 @@
...
+	v1 "k8s.io/api/core/v1"
```
JoelSpeed, line 8: This should be imported as `corev1`

pkg/framework/proxies.go
```diff
@@ -0,0 +1,196 @@
+package framework
+
+import (
+	"context"
+	"errors"
```
JoelSpeed, line 5: The linter will probably complain at you for not having a blank line between standard library packages and non-standard library packages

pkg/framework/pods.go
```diff
@@ -0,0 +1,13 @@
...
+	"context"
+	corev1 "k8s.io/api/core/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetPods(client runtimeclient.Client, selector map[string]string) (corev1.PodList, error) {
+	pods := corev1.PodList{}
+	err := client.List(context.TODO(), &pods, runtimeclient.MatchingLabels(selector))
+	return pods, err
+}
```
JoelSpeed, line 13: Normally you would return a pointer to an object so that it implements the right interfaces (runtime.Object, metav1.Object), makes it easier to use at the call site
```suggestion
func GetPods(client runtimeclient.Client, selector map[string]string) (*corev1.PodList, error) {
	pods := &corev1.PodList{}
	err := client.List(context.TODO(), pods, runtimeclient.MatchingLabels(selector))
	return pods, err
}
```

JoelSpeed (commented):
This is great, left a few comments which are pretty much all stylistic nit picks, please update as appropriate

pkg/framework/pods.go
rvanderp3: Thanks for reviewing @JoelSpeed .  This is my first interaction with this specific test frame.  I really do appreciate the guidance here!!  I'll make the appropriate edits.

pkg/framework/proxies.go
rvanderp3: updated

pkg/operators/machine-api-operator.go
rvanderp3: good know, I pulled that out.

pkg/framework/proxies.go
rvanderp3: sure thing

pkg/operators/machine-api-operator.go
rvanderp3: makes total sense. I restructured in to a three test cases which test specific aspects.

pkg/framework/proxies.go
rvanderp3: makes sense.  i refactored it.

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
```
JoelSpeed, line 13: I think we should be using `klog/v2` now
```suggestion
	"k8s.io/klog/v2"
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+acl mysubnet src 192.168.0.0/16
+http_access allow localhost localnet crcnet
+http_access allow all
+http_port 80
+coredump_dir /var/spool/squid
+refresh_pattern ^ftp:		1440	20%	10080
+refresh_pattern ^gopher:	1440	0%	1440
+refresh_pattern -i (/cgi-bin/|\?) 0	0%	0
+refresh_pattern (Release|Packages(.gz)*)$      0       20%     2880
+refresh_pattern .		0	20%	4320`
```
JoelSpeed, line 44: This is much better! 😄

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+
+// CreateProxyMachineSet creates a machineset that results in a single node being spun up. This node will not be in the worker MCP.
+func CreateProxyMachineSet(client runtimeclient.Client) (*mapiv1beta1.MachineSet, error) {
+	var machineSet *mapiv1beta1.MachineSet
+	var machineSetParams MachineSetParams
+
+	machineSetParams = BuildMachineSetParams(client, 1)
```
JoelSpeed, line 51: Don't need to declare these vars up here, they can be created inline
```suggestion
	machineSetParams := BuildMachineSetParams(client, 1)
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	WaitForMachineSet(client, machineSet.GetName())
+
+	nodes, err := GetNodesFromMachineSet(client, machineSet)
+	if len(nodes) == 0 {
+		return nil, errors.New("No nodes found in machineset")
+	}
+	node := nodes[0]
+	delete(node.Labels, "node-role.kubernetes.io/worker")
+	client.Update(context.Background(), node)
```
JoelSpeed, line 66: Should we not check this error?
```suggestion
	if err := client.Update(context.Background(), node); err != nil {
		return nil, err
	}
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+	return machineSet, nil
+}
+
+// DeployClusterProxy Deploys an HTTP proxy to the proxy node
+func DeployClusterProxy(c runtimeclient.Client) (*appsv1.Deployment, error) {
+	configMap := corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{},
```
JoelSpeed, line 73: We can get rid of this one
```suggestion
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "squid-conf",
+			Namespace: "default",
+		},
+		Data: map[string]string{
+			"squid.conf": squidConf,
+		},
+	}
+
+	c.Create(context.Background(), &configMap)
```
JoelSpeed, line 83: Check error?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+
+	deployment := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{},
```
JoelSpeed, line 86: Can remove this
```suggestion
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "squid-proxy",
+			Namespace: "default",
+			Labels: map[string]string{
+				"app": "squid",
+			},
```
JoelSpeed, line 92: Could make this label set a var above so we can reuse it rather than redeclaring thought this config

```
labels := map[string]string{
	"app": "squid",
}
```

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+							},
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "conf",
+									ReadOnly:  false,
+									MountPath: "/etc/squid/squid.conf",
+									SubPath:   "squid.conf",
+								},
+							},
+							SecurityContext: &corev1.SecurityContext{},
```
JoelSpeed, line 138: Odd to have this empty? Shall we drop it?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+					NodeSelector: map[string]string{
+						"proxy-host": "",
+					},
+					HostNetwork: true,
+				},
+			},
+			Strategy: appsv1.DeploymentStrategy{
+				Type: "Recreate",
+			},
+			Paused: false,
```
JoelSpeed, line 150: This would be a default, don't need it right?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+		},
+	}
+	c.Create(context.Background(), deployment)
```
JoelSpeed, line 153: Need to check error

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+	IsDeploymentAvailable(c, deployment.Name, "default")
```
JoelSpeed, line 154: Need to check return value

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+// DestroyClusterProxy destroys the HTTP proxy and associated configmap
+func DestroyClusterProxy(c runtimeclient.Client) error {
+	configMap := &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "squid-conf",
+			Namespace: "default",
+		},
+	}
+	c.Delete(context.Background(), configMap)
```
JoelSpeed, line 167: Check error

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "squid-proxy",
+			Namespace: "default",
+		},
+	})
+}
+
+// WaitForProxyInjectionSync waits for the deployment to sync with the state of the cluster-proxy
+func WaitForProxyInjectionSync(c runtimeclient.Client, name, namespace string, shouldBePresent bool) bool {
+	if err := wait.PollImmediate(RetryShort, WaitLong, func() (bool, error) {
```
JoelSpeed, line 179: On a `WaitLong`, might want to reduce the retry, perhaps `RetryMedium` would be more appropriate here?

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+				case "HTTP_PROXY":
+					hasHttpProxy = true
+				}
+			}
+		}
+		return (hasHttpProxy &&
+			hasHttpsProxy &&
+			hasNoProxy) == shouldBePresent, nil
+	}); err != nil {
+		klog.Errorf("Error checking isDeploymentAvailable: %v", err)
```
JoelSpeed, line 203: This error message needs an update

pkg/framework/proxies.go
```diff
@@ -0,0 +1,249 @@
...
+	if updatedProxy.Spec.HTTPSProxy != "" {
+		proxy.Spec.HTTPSProxy = updatedProxy.Spec.HTTPSProxy
+	} else {
+		proxy.Spec.HTTPSProxy = ""
+	}
+	if updatedProxy.Spec.NoProxy != "" {
+		proxy.Spec.NoProxy = updatedProxy.Spec.NoProxy
+	} else {
+		proxy.Spec.NoProxy = ""
+	}
```
JoelSpeed, line 243: All of these can be condensed to 
```suggestion
	proxy.Spec.HTTPProxy = updatedProxy.Spec.HTTPProxy
	proxy.Spec.HTTPSProxy = updatedProxy.Spec.HTTPSProxy
	proxy.Spec.NoProxy = updatedProxy.Spec.NoProxy
```
But do we really need to do this? Would we not want to just `c.Update(context.Background(), updatedProxy)`? Do we even need this wrapper function?

pkg/operators/machine-api-operator.go
```diff
@@ -164,3 +167,71 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
...
+		framework.SetClusterProxy(client, proxy)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("waiting for machine-api-controller deployment to reflect unconfigured cluster-wide proxy")
+		Expect(framework.WaitForProxyInjectionSync(client, maoManagedDeployment, framework.MachineAPINamespace, false)).To(BeTrue())
+		framework.DestroyClusterProxy(client)
+
+		Expect(proxyMachineSet != nil).To(BeTrue())
+		framework.DeleteMachineSets(client, proxyMachineSet)
+	})
```
JoelSpeed, line 235: This either needs to be one `It` statement as below, or it needs to be a `BeforeEach` which sets up the cluster-wide proxy and an `AfterEach` which cleans this up again, with the `It` statement being the one that creates and destroys a MachineSet
```suggestion
	It("create machines when configured behind a proxy", func() {
		By("creating host for proxy")
		client, err := framework.LoadClient()
		Expect(err).NotTo(HaveOccurred())
		proxyMachineSet, err := framework.CreateProxyMachineSet(client)
		Expect(err).NotTo(HaveOccurred())

		By("deploying an HTTP proxy")
		deployment, err := framework.DeployClusterProxy(client)
		Expect(err).NotTo(HaveOccurred())
		Expect(framework.IsDeploymentAvailable(client, deployment.Name, deployment.Namespace)).To(BeTrue())

		By("configuring cluster-wide proxy")
		pods, err := framework.GetPods(client, map[string]string{"app": "squid"})
		Expect(err).NotTo(HaveOccurred())
		Expect(len(pods.Items) > 0).To(BeTrue())
		proxy := configv1.Proxy{}
		proxy.Spec.HTTPProxy = "http://" + pods.Items[0].Status.HostIP + ":80"
		proxy.Spec.HTTPSProxy = "http://" + pods.Items[0].Status.HostIP + ":80"
		framework.SetClusterProxy(client, proxy)
		Expect(err).NotTo(HaveOccurred())

		By("waiting for machine-api-controller deployment to reflect configured cluster-wide proxy")
		Expect(framework.WaitForProxyInjectionSync(client, maoManagedDeployment, framework.MachineAPINamespace, true)).To(BeTrue())

		By("creating a machineset")
		machineSetParams = framework.BuildMachineSetParams(client, 3)
		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
		Expect(err).ToNot(HaveOccurred())
		framework.WaitForMachineSet(client, machineSet.GetName())

		By("destroying a machineset")
		Expect(client.Delete(context.Background(), machineSet)).To(Succeed())
		framework.WaitForMachineSetDelete(client, machineSet)

		By("unconfiguring cluster-wide proxy")
		proxy := configv1.Proxy{}
		proxy.Spec.HTTPProxy = ""
		proxy.Spec.HTTPSProxy = ""
		framework.SetClusterProxy(client, proxy)
		Expect(err).NotTo(HaveOccurred())

		By("waiting for machine-api-controller deployment to reflect unconfigured cluster-wide proxy")
		Expect(framework.WaitForProxyInjectionSync(client, maoManagedDeployment, framework.MachineAPINamespace, false)).To(BeTrue())
		framework.DestroyClusterProxy(client)

		Expect(proxyMachineSet != nil).To(BeTrue())
		framework.DeleteMachineSets(client, proxyMachineSet)
	})
```

pkg/operators/machine-api-operator.go
```diff
@@ -164,3 +167,71 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
 		Expect(framework.IsStatusAvailable(client, "machine-api")).To(BeTrue())
 	})
 })
+
+var _ = Describe("[Feature:Operators] When cluster-wide proxy is configured, Machine API cluster operator should ", func() {
```
JoelSpeed, line 171: I don't think this can run in parallel with other tests, I think we need to mark as serial as suggested below, but I think @elmiko can confirm this
```suggestion
var _ = Describe("[Serial][Feature:Operators] When cluster-wide proxy is configured, Machine API cluster operator should ", func() {
```

pkg/framework/proxies.go
rvanderp3: very good call.  when I wrote that I thought those variables inited to nil when undefined, which they don't.  I'll make that change.

pkg/framework/proxies.go
```diff
@@ -0,0 +1,222 @@
...
+										Name: "squid-conf",
+									},
+								},
+							},
+						},
+					},
+					Containers: []corev1.Container{
+						{
+							Name:  "squid",
+							Image: "sameersbn/squid:3.5.27-2",
```
elmiko, line 119: i have one suggestion and one question here,

i think we should use the full pull spec for the image, just to ensure there are no questions of provenance.

i looked at this image and repo it is generated from, is this a common community image for squid or is there something with more recent maintenance we could use? (i'm just thinking about this test in the long term, how can we maintain this implementation)

pkg/operators/machine-api-operator.go
```diff
@@ -203,3 +203,58 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
...
+		Expect(framework.IsDeploymentAvailable(client, deployment.Name, deployment.Namespace)).To(BeTrue())
+
+		By("configuring cluster-wide proxy")
+		pods, err := framework.GetPods(client, map[string]string{"app": "squid"})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(pods.Items) > 0).To(BeTrue())
+		proxy, err := framework.GetClusterProxy(client)
+		Expect(err).NotTo(HaveOccurred())
+		proxy.Spec.HTTPProxy = "http://" + pods.Items[0].Status.HostIP + ":80"
+		proxy.Spec.HTTPSProxy = "http://" + pods.Items[0].Status.HostIP + ":80"
```
elmiko, line 227: should this be `443` for the https proxy?

elmiko (commented):
this looks mostly good to me, i had a suggestion and a couple questions inline.

also, we will need to adjust the makefile is this needs to be run as a serial job. we currently run most of these tests in parallel from the make target[0]. i think we'll need to add a skip for this job in the parallel call and either add it to the serial test instruction or add a new instruction for this test.

[0] https://github.com/openshift/cluster-api-actuator-pkg/blob/master/Makefile#L67

pkg/operators/machine-api-operator.go
rvanderp3: It is non-intuitive but both HTTP and HTTPS proxy requests are proxied through HTTP.  Most HTTP proxies don't accept HTTPS connections directly to the proxy.  Additionally, the proxy I deployed for this is configured to meet this spec.

pkg/framework/proxies.go
rvanderp3: ah, that is a great point.  I'll need to think on that.

pkg/operators/machine-api-operator.go
elmiko: ack, thanks!

JoelSpeed:
@rvanderp3 Looks like either the proxying isn't working or the test isn't for some reason, getting proxy connection errors https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/192/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1328671453553364992/artifacts/e2e-aws-operator/gather-extra/pods/openshift-machine-api_machine-api-controllers-6c69996fd8-7q5wq_machine-controller.log

rvanderp3:
> @rvanderp3 Looks like either the proxying isn't working or the test isn't for some reason, getting proxy connection errors https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/192/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-operator/1328671453553364992/artifacts/e2e-aws-operator/gather-extra/pods/openshift-machine-api_machine-api-controllers-6c69996fd8-7q5wq_machine-controller.log

I'm looking in to it now.

JoelSpeed:
Is there any update on getting this to work on AWS?

rvanderp3:
> Is there any update on getting this to work on AWS?

Yep, I am performing some unit testing now.  I've been having issues getting access to an AWS cluster for test purposes.  As soon as I get the latest changes tested, I will get them in to the PR.

pkg/framework/proxies.go
rvanderp3: I needed to restructure the test due to a restriction in host networking between nodes which I thought I had taken in to account.  During the process of addressing that, I went ahead and changed the proxy image to be built on ubi-8 and pull from registry.redhat.io.

JoelSpeed:
/retest

Test failures look unrelated, infrastructure issues seemingly

pkg/framework/proxies.go
alexander-demichev: this still needs to be adressed

alexander-demichev (changes requested):

pkg/framework/proxies.go
rvanderp3: thanks.  i'll take care of that today.

JoelSpeed:
There's a few nits I could make about consistency with imports and naming, but I think at this point it's better to get this merged than nitpick. Will need to create a BZ to get this merged now though. I'll get that done shortly.

/approve

Thanks for all your work on this @rvanderp3!

JoelSpeed:
/retitle Bug 1906732: Add E2E to ensure Cluster Wide Proxies are honoured in Machine API components

alexander-demichev (approved):
Thanks!
/lgtm

PR #193, opened 2020-10-13
Merge spot tests to a single test

JoelSpeed:
This means that the test suite only needs to spin up a single machineset for spot instances.

This is an easy alternative to having a `BeforeAll` which will hopefully come in Ginkgo V2, once it does, we can revert this and change the `BeforeEach` to a `BeforeAll`.

My hope is that by reducing the number of spot instances we need to spin up, this will improve the reliability of this test on Azure which seems to have some issues bringing up spot instances.

I'd suggest reviewing this with white space changes ignored, it should make a lot more sense that way 🤞 

CC @kwoodson

### DIFF
--- a/pkg/infra/spot.go
+++ b/pkg/infra/spot.go
@@ -71,119 +71,121 @@ var _ = Describe("[Feature:Machines] Running on Spot", func() {
 		Expect(deleteObjects(client, delObjects)).To(Succeed())
 	})
 
-	It("should label the Machine specs as interruptible", func() {
-		selector := machineSet.Spec.Selector
-		machines, err := framework.GetMachines(client, &selector)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(machines).To(HaveLen(3))
+	It("should handle the spot instances", func() {
+		By("should label the Machine specs as interruptible", func() {
+			selector := machineSet.Spec.Selector
+			machines, err := framework.GetMachines(client, &selector)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(machines).To(HaveLen(3))
 
-		for _, machine := range machines {
-			Expect(machine.Spec.ObjectMeta.Labels).To(HaveKeyWithValue(machinecontroller.MachineInterruptibleInstanceLabelName, ""))
-		}
-	})
+			for _, machine := range machines {
+				Expect(machine.Spec.ObjectMeta.Labels).To(HaveKeyWithValue(machinecontroller.MachineInterruptibleInstanceLabelName, ""))
+			}
+		})
 
-	It("should deploy a termination handler pod to each instance", func() {
-		nodes, err := framework.GetNodesFromMachineSet(client, machineSet)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(nodes).To(HaveLen(3))
+		By("should deploy a termination handler pod to each instance", func() {
+			nodes, err := framework.GetNodesFromMachineSet(client, machineSet)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(nodes).To(HaveLen(3))
 
-		terminationLabels := map[string]string{
-			"api":     "clusterapi",
-			"k8s-app": "termination-handler",
-		}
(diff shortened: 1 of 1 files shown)


elmiko (approved):
this seems reasonable to me, do we have any data yet that would demonstrate the stability?

JoelSpeed:
/test e2e-azure-operator
/test e2e-aws-operator
/test e2e-gcp-operator

Retesting to check stability

> this seems reasonable to me, do we have any data yet that would demonstrate the stability?

I was hoping that the metric added in https://github.com/openshift/machine-api-operator/pull/640 would provide this but it's not merged yet.

This has come from a repeated observation that the spot test is failing and some analysis from @kwoodson who suspects that bringing up spot instances is causing much of the Azure CI flakiness

JoelSpeed:
Different tests failed for azure and gcp

/test e2e-gcp-operator
/test e2e-aws-operator

JoelSpeed:
Last Azure test failed for CPU limit reasons, this has now been fixed!
/retest

kwoodson:
@JoelSpeed Looks like the test failed when the autoscaler was scaling back down from 6.
```
• Failure [1416.664 seconds]
[Feature:Machines] Autoscaler should
/go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:214
  use a ClusterAutoscaler that has 12 maximum total nodes count and balance similar nodes enabled
  /go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:490
    scales up and down while respecting MaxNodesTotal [Slow][Serial] [It]
    /go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:521

    Timed out after 360.000s.
    Error: Unexpected non-nil/non-zero extra argument at index 1:
    	<*fmt.wrapError>: &fmt.wrapError{msg:"error getting node from machine \"ci-op-9i19wc9q-d57e2-kxndq4vmr6-556fl\": nodes \"ci-op-9i19wc9q-d57e2-kxndq4vmr6-556fl\" not found", err:(*errors.StatusError)(0xc000603040)}

    /go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:589
```

JoelSpeed:
/retest

@kwoodson hoping to resolve that flakiness with #195

JoelSpeed:
@elmiko In the last 24 hours this test suite has run 4 times for this PR, none of the failures related to spot. On the MAO repo it ran 11 times yesterday and saw 2 failures from spot on azure, possibly not statistically significant but on the other hand, I can't really see how this change would make things worse, it reduces the number of VMs we are spinning up and therefore should make the suite faster and less prone to slow VM creation when we create many VMs at once.

JoelSpeed:
/retest

That said, it did just fail on spot, but for a different reason, which I believe is unrelated to the issue we are trying to fix here

elmiko:
makes sense to me @JoelSpeed , i'm good with this change
/approve

kwoodson:
@JoelSpeed I'm unfamiliar with this test failure.  Looks like a timeout but I'm not sure why.
```
Unexpected error:
    <*errors.errorString | 0xc00053b990>: {
        s: "error getting ValidatingWebhookConfiguration \"machine-api\": timed out waiting for the condition",
    }
    error getting ValidatingWebhookConfiguration "machine-api": timed out waiting for the condition
occurred
/go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/operators/machine-api-operator.go:173
```

JoelSpeed:
/retest

@kwoodson This test failure is really odd, but I think it's because we are running multiple disruptive tests in parallel

There are a couple of tests that affect the VWC, one that deletes it, one that modifies it, both test the controller puts it back. These shouldn't be running in parallel, they will clash with one another. Will take a look at a quick fix by merging the disruptive tests to a single test.

https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/operators/machine-api-operator.go

PR #194, opened 2020-10-15
Reduce failures in webhook tests

JoelSpeed:
I keep seeing the following in test failures, looks like we are checking the object before the error. Would be better to check the error first so that the error message is displayed in the JUnit output
```
/go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/operators/machine-api-operator.go:129
Expected
    <*v1.ValidatingWebhookConfiguration | 0x0>: nil
not to be nil
/go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/operators/machine-api-operator.go:134
```

Secondly, based on some analysis of the test results, it looks like we are not actually waiting for the object to be deleted and recreated in the recover after deletion tests.

These are the only tests that should be deleting the MWC/VWC yet looking at the must-gather for https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-provider-azure/158/pull-ci-openshift-cluster-api-provider-azure-master-e2e-azure-operator/1316423821439275008 shows that the MWC/VWC are recreated after the test suite ends, ie, we didn't actually wait for them to get recreated in the test.

I've added some extra checks to the tests that should check that either the MWC/VWC is gone, or that the UID has changed. Both should signify that the resource was deleted successfully. Then, and only then, is it valid to wait for the new object to be created and to be in sync with the expectations.

### DIFF
--- a/pkg/operators/machine-api-operator.go
+++ b/pkg/operators/machine-api-operator.go
@@ -1,12 +1,17 @@
 package operators
 
 import (
+	"context"
 	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apitypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/operators/machine-api-operator.go
```diff
@@ -85,17 +90,51 @@ var _ = Describe("[Feature:Operators] Machine API operator deployment should", f
...
-		Expect(framework.DeleteValidatingWebhookConfiguration(client, framework.DefaultValidatingWebhookConfiguration)).To(Succeed())
+		// Record the UID of the current ValidatingWebhookConfiguration
+		initial, err := framework.GetValidatingWebhookConfiguration(client, framework.DefaultValidatingWebhookConfiguration.Name)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(initial).ToNot(BeNil())
+		initialUID := initial.GetUID()
+
+		Expect(framework.DeleteValidatingWebhookConfiguration(client, initial)).To(Succeed())
 
+		// Ensure that either UID changes (to show a new object) or that the existing object is gone
```
elmiko, line 101: i just want to understand this a little better, shouldn't this always check that the old object was deleted?

elmiko (commented):
this generally makes sense to me, i just have a question about the deletion detection.

pkg/operators/machine-api-operator.go
JoelSpeed: If the old object was deleted, and the new one created in a very quick succession, we wouldn't get a 404, so I've written this so that it can detect that the object has changed to minimise flakes if at all possible

michaelgugino (approved):
/lgtm

PR #196, opened 2020-10-21
Run operator tests in serial rather than parallel

JoelSpeed:
We have disruptive tests in the operator suite that do not tolerate running in parallel. Disable the parallelism until we come up with a better way to run the tests in parallel. (This may not actually be possible given the types of objects we are working with)


michaelgugino (approved):
/lgtm

PR #197, opened 2020-10-28
Temporarily skip tests to ease termination handler conditions migration

JoelSpeed:
To enable https://github.com/openshift/machine-api-operator/pull/627 to merge and allow the providers to migrate, we are going to need to temporarily disable the tests.

Merging the MAO changes first breaks the permissions for the service account and hence breaks the test. Merging the implementation first then doesn't have the appropriate permissions to set conditions on nodes, and breaks the tests.
Only with both merged can we re-enable the tests.


alexander-demichev (approved):
/approve

PR #198, opened 2020-10-29
Add e2e tests for AWS dedicated instances

alexander-demichev:
Add e2e tests for AWS dedicated instances


JoelSpeed (commented):
Looks good to me! I assume we will need to request some dedicated capacity for the CI environment? Have you reached out to them to mention you're adding this?

/approve

JoelSpeed:
/retest

The AWS failure was for the new test, but not obviously an issue with the test that I could see, only errors logged are capacity which doesn't seem to imply any issue with the test as far as I can tell. Let's try running again

pkg/infra/dedicated.go
```diff
@@ -0,0 +1,97 @@
...
+		By("Creating a Dedicated backed MachineSet", func() {
+			machineSetParams = framework.BuildMachineSetParams(client, 3)
+			Expect(setDedicatedOnProviderSpec(platform, machineSetParams, "")).To(Succeed())
+
+			machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+			Expect(err).ToNot(HaveOccurred())
+			delObjects[machineSet.Name] = machineSet
+
+			framework.WaitForMachineSet(client, machineSet.GetName())
+		})
```
JoelSpeed, line 51: This is duplicated in the IT statement below, we probably don't need it here

PR #201, opened 2020-12-03
Improve resilience of autoscaler 12 nodes total test

JoelSpeed:
This is an attempt to improve the resilience of the following test. In this case, the node had gone away but the Machine had not. We can check those two things separately. The `GetNodesFromMachineSet` helper in the framework should be ignoring the not found error, it is up to the caller to check the number is correct.

```
Machine Suite: [Feature:Machines] Autoscaler should use a ClusterAutoscaler that has 12 maximum total nodes count and balance similar nodes enabled scales up and down while respecting MaxNodesTotal [Slow][Serial] expand_less | 23m20s
-- | --
/go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:521 Timed out after 360.001s. Error: Unexpected non-nil/non-zero extra argument at index 1: 	<*fmt.wrapError>: &fmt.wrapError{msg:"error getting node from machine \"ci-op-cr8lz3b0-d5662-pr6wdmng5s-27ltb\": nodes \"ci-op-cr8lz3b0-d5662-pr6wdmng5s-27ltb\" not found", err:(*errors.StatusError)(0xc00065c320)} /go/src/github.com/openshift/cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:589
```

### DIFF
--- a/pkg/framework/nodes.go
+++ b/pkg/framework/nodes.go
@@ -9,6 +9,7 @@ import (
 
 	mapiv1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


elmiko (approved):
thanks Joel!
/approve

pkg/framework/nodes.go
```diff
@@ -71,7 +72,11 @@ func GetNodesFromMachineSet(client runtimeclient.Client, machineSet *mapiv1beta1
 	var nodes []*corev1.Node
 	for key := range machines {
 		node, err := GetNodeForMachine(client, machines[key])
-		if err != nil {
+		if err != nil && apierrors.IsNotFound(err) {
```
Danil-Grigorev, line 75: Just `apierrors.IsNotFound(err)` is enough.

pkg/framework/nodes.go
```diff
@@ -71,7 +72,11 @@ func GetNodesFromMachineSet(client runtimeclient.Client, machineSet *mapiv1beta1
...
+			// We don't care about not found errors.
+			// Callers should account for the number of nodes being correct or not.
```
Danil-Grigorev, line 77: It still will be better to put a log message in there. The method signature suggests it should be a matching list to the assigned Machines.
```golang
klog.Infof("No Node object found for machine %s", machine.GetName())
```

Danil-Grigorev (commented):
Makes sense to me, just some comments.

JoelSpeed:
@Danil-Grigorev Good feedback, I included your suggestions

PR #202, opened 2020-12-15
Bug 1914837: Re-enable termination handler tests

JoelSpeed:
Reverts openshift/cluster-api-actuator-pkg#197 which disabled the termination handler tests to allow us to migrate from the deletion implementation to the conditions implementation


JoelSpeed:
Attached this bug as the tests should only pass once this bug is resolved

JoelSpeed:
I ran the spot test locally against a GCP 4.7 cluster and it worked as expected, not sure why it would be failing in CI
/test e2e-gcp-operator

JoelSpeed:

The test is permafailing, I've got a fix for that in #213. I think this is safe to override as the specific test this has re-enabled has been stable through all these restests

PR #203, opened 2020-12-16
Change minimal providerSpec for vSphere

alexander-demichev:
Change minimal providerSpec for vSphere. All fields in vSphere's provider spec might be required depending on installation, we can leave only secret references empty.

### DIFF
--- a/pkg/infra/webhooks.go
+++ b/pkg/infra/webhooks.go
@@ -233,26 +233,17 @@ func minimalGCPProviderSpec(ps *mapiv1.ProviderSpec) (*mapiv1.ProviderSpec, erro
 }
 
 func minimalVSphereProviderSpec(ps *mapiv1.ProviderSpec) (*mapiv1.ProviderSpec, error) {
-	fullProviderSpec := &vsphere.VSphereMachineProviderSpec{}
-	err := json.Unmarshal(ps.Value.Raw, fullProviderSpec)
+	providerSpec := &vsphere.VSphereMachineProviderSpec{}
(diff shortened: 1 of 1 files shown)


JoelSpeed:
The test has passed in the latest run through of the test suite, there are other issues in vSphere CI though and at the moment it seems our timeouts for creating the machines are not long enough so we are getting a lot of failures across the board for VMs not starting in time. Not sure what we can do about that

/retest

alexander-demichev:
We are still getting timeouts from vsphere. I need to rework the last 2 commits in order for bigger timeouts to be set for vsphere only.

Makefile
```diff
@@ -68,7 +68,7 @@ test-e2e: ## Run openshift specific e2e test
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
 	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" -skip="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler|\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
```
JoelSpeed, line 71: Why are we removing the parallelism from these tests? This is making the test suite take over 2 hours on azure and therefore it is perma failing

JoelSpeed:
/hold cancel

I'm happy that this is sufficient and all the tests have passed now

JoelSpeed:

This PR will not affect azure and we have previously seen this pass

PR #204, opened 2021-01-14
Bug 1913341: [OCPCLOUD-1072] Add Disruptive test selection and mark proxy tests as disruptive.

Danil-Grigorev:
Proxy tests are containing some basic post condition identifying cluster as recovered.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,14 +59,18 @@ build-e2e:
 
 .PHONY: test-e2e
 test-e2e: ## Run openshift specific e2e test
+	# Run Disruptive tests first. Those might include things causing all masters to restart, and should run serially
+
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
+
 	# Run operator tests first to preserve logs for troubleshooting test
 	# failures and flakes.
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" -skip="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler|\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
 	# TODO: parallelise autoscaler
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" -skip="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
 	# After success, merge all JUnit files into one
 	hack/junitmerge.sh
 
--- a/pkg/framework/framework.go
+++ b/pkg/framework/framework.go
@@ -124,14 +124,25 @@ func LoadClientset() (*kubernetes.Clientset, error) {
 	return kubernetes.NewForConfig(cfg)
 }
 
-func IsStatusAvailable(client runtimeclient.Client, name string) bool {
+func WaitForStatusAvailableShort(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitShort)
+}
+
+func WaitForStatusAvailableMedium(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitMedium)
+}
+
+func WaitForStatusAvailableOverLong(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitOverLong)
+}
+
+func expectStatusAvailableIn(client runtimeclient.Client, name string, timeout time.Duration) bool {
 	key := types.NamespacedName{
-		Namespace: MachineAPINamespace,
-		Name:      name,
+		Name: name,
 	}
 	clusterOperator := &configv1.ClusterOperator{}
 
-	if err := wait.PollImmediate(RetryShort, WaitShort, func() (bool, error) {
+	if err := wait.PollImmediate(RetryMedium, timeout, func() (bool, error) {
 		if err := client.Get(context.TODO(), key, clusterOperator); err != nil {
 			klog.Errorf("error querying api for OperatorStatus object: %v, retrying...", err)
 			return false, nil
--- a/pkg/operators/cluster-autoscaler-operator.go
+++ b/pkg/operators/cluster-autoscaler-operator.go
@@ -92,6 +92,6 @@ var _ = Describe("[Feature:Operators] Cluster autoscaler cluster operator status
 		var err error
 		client, err := framework.LoadClient()
 		Expect(err).NotTo(HaveOccurred())
-		Expect(framework.IsStatusAvailable(client, "cluster-autoscaler")).To(BeTrue())
+		Expect(framework.WaitForStatusAvailableShort(client, "cluster-autoscaler")).To(BeTrue())
 	})
 })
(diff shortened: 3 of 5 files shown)


michaelgugino (approved):
/lgtm

Makefile
```diff
@@ -59,14 +59,19 @@ build-e2e:
...
 	# failures and flakes.
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" -skip="Disruptive" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler|Disruptive" || (hack/junitmerge.sh && exit 1)
 	# TODO: parallelise autoscaler
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Autoscaler" -skip="Disruptive" || (hack/junitmerge.sh && exit 1)
```
JoelSpeed, line 74: Which takes precedence here? Skip or Focus?

Makefile
```diff
@@ -59,14 +59,19 @@ build-e2e:
...
+	# TODO: Disabled, until all disruptive tests are prefixed with post-conditions, allowing cluster to recover
+	# hack/ci-integration.sh $(GINKGO_ARGS) -focus="Disruptive" || (hack/junitmerge.sh && exit 1)
+
 	# Run operator tests first to preserve logs for troubleshooting test
 	# failures and flakes.
 	# Feature:Operator tests remove deployments. Thus loosing all the logs
 	# previously acquired.
-	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" || (hack/junitmerge.sh && exit 1)
-	hack/ci-integration.sh $(GINKGO_ARGS) -p -skip="Feature:Operators|Autoscaler" || (hack/junitmerge.sh && exit 1)
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Feature:Operators" -skip="Disruptive" || (hack/junitmerge.sh && exit 1)
```
JoelSpeed, line 71: Should we be more explicit here and make this `[Disruptive]` including the brackets, just in case a test randomly has disruptive in the title anyway?

Makefile
Danil-Grigorev: Definitely skip.

Makefile
Danil-Grigorev: I'm afraid of possibility for this to become a wildcard regular expression.

JoelSpeed: Does the syntax make it a regex, In which case we can escape the brackets right?

Danil-Grigorev: Let's try this. Need another LGTM then.

JoelSpeed:
/approve
/lgtm

Seems to be doing what it says on the tin

Danil-Grigorev:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/204/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1349735203181957120 is failing, the problem still persists.

/retest

Danil-Grigorev:
Wait, I was assuming this proxy change got merged, and didn't take an effect, so marked webhook tests as disruptive. @JoelSpeed I did a change, should I revert it and will you be able to give it a second LGTM?..

Makefile
```diff
@@ -59,14 +59,19 @@ build-e2e:
 
 .PHONY: test-e2e
 test-e2e: ## Run openshift specific e2e test
+	# Run Disruptive tests first. Those might include things causing all masters to restart, and should run serially
+
+	# TODO: Disabled, until all disruptive tests are prefixed with post-conditions, allowing cluster to recover
+	# hack/ci-integration.sh $(GINKGO_ARGS) -focus="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
```
JoelSpeed, line 65: Does this currently run any tests? Is it worth merging it uncommented and just having it be a no-op for now?

Danil-Grigorev: I think we can uncomment it, but revert to the state of https://github.com/openshift/cluster-api-actuator-pkg/commit/d8f139d9f67dfd10dff4b7cfe3ea60c9cefba429 and https://github.com/openshift/cluster-api-actuator-pkg/commit/e3a1a4fb39a61566087d97d2139cadebf9482c87 first. Let'd mark proxy tests as disruptive for now, and consider adding post/pre conditions later @JoelSpeed

JoelSpeed: I don't really follow what you mean by reverting? As far as I can tell, this should be good to merge if we just uncomment this line. What are you concerned about in particular?

Danil-Grigorev: No, just trying to recall conversation in slack on this topic. SGTM and updated, PTAL @JoelSpeed

JoelSpeed:
Code changes look ok to me, want to see how the tests pass/fail before giving approval

Danil-Grigorev:
Previous code changes are sufficient to fix the issue mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1913341#c4 Reverting this to https://github.com/openshift/cluster-api-actuator-pkg/commit/d8f139d9f67dfd10dff4b7cfe3ea60c9cefba429

Makefile
```diff
@@ -59,14 +59,18 @@ build-e2e:
 
 .PHONY: test-e2e
 test-e2e: ## Run openshift specific e2e test
+	# Run Disruptive tests first. Those might include things causing all masters to restart, and should run serially
+
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Disruptive" || (hack/junitmerge.sh && exit 1)
```
JoelSpeed, line 64: Should this use the same focus as the others skip?
```suggestion
	hack/ci-integration.sh $(GINKGO_ARGS) -focus="\[Disruptive\]" || (hack/junitmerge.sh && exit 1)
```

pkg/framework/framework.go
```diff
@@ -125,13 +125,21 @@ func LoadClientset() (*kubernetes.Clientset, error) {
 }
 
 func IsStatusAvailable(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitShort)
+}
+
+func WaitForStatusAvailable(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitMedium)
+}
```
JoelSpeed, line 133: Do we need two helpers with different names like this? This to me implies that one is sync and one is async, but they are both async. Can we make this clearer somehow?

Danil-Grigorev: It is essentially like this. One scenario is expecting an absence of interruption in MAO `IsStatusAvailable` with minimal amount of polling (just to exclude case when is something with API server), and another is actually waiting for Available condition in ~3 min, which is the time required to roll out our components and verify that they are stable.

Makefile
Danil-Grigorev: Fixed, PTAL.

pkg/framework/framework.go
JoelSpeed: Why not name them `WaitForStatusAvailableShort`/`WaitForStatusAvailableLong` in that case? I still think it's confusing that these are basically the same but have different names completely

Danil-Grigorev: Updated PTAL @JoelSpeed

pkg/framework/framework.go
```diff
@@ -124,14 +124,22 @@ func LoadClientset() (*kubernetes.Clientset, error) {
 	return kubernetes.NewForConfig(cfg)
 }
 
-func IsStatusAvailable(client runtimeclient.Client, name string) bool {
+func WaitForStatusAvailableShort(client runtimeclient.Client, name string) bool {
+	return expectStatusAvailableIn(client, name, WaitShort)
+}
+
+func WaitForStatusAvailableLong(client runtimeclient.Client, name string) bool {
```
JoelSpeed, line 131: Since this uses wait medium let's call it medium
```suggestion
func WaitForStatusAvailableMedium(client runtimeclient.Client, name string) bool {
```

Danil-Grigorev:
/hold

Putting a hold on that, still needs some work. The post condition for proxy was not sufficient. Webhooks are still flaking ATM.

JoelSpeed:
Why are we waiting now for KAS and KCM to become available?

Danil-Grigorev:
> Why are we waiting now for KAS and KCM to become available?

This is the only way to ensure the cluster is healthy again. Or we will see flakes in our tests in different places.

lobziik:
Might it make sense to have a separate make target and CI step as well, like we have for origin tests?
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/machine-api-operator/openshift-machine-api-operator-master-presubmits.yaml#L962

PR #207, opened 2021-04-07
update Makefile to run with podman, update Readme

SamuelStuchly:
This PR updates Makefile to be able to run containers with podman as well as docker. 
Also updates Readme.


Makefile
```diff
@@ -7,33 +7,45 @@ export GOFLAGS
...
 vendor:
-	go mod tidy
-	go mod vendor
-	go mod verify
+	$(DOCKER_CMD) ./hack/go-mod.sh
+
 
 .PHONY: check
-check: fmt vet lint test ## Check your code
+check: fmt vet #lint ## Check your code
```
lobziik, line 48: why lint commented out there?

SamuelStuchly: Duo to lint not working. Comment says `Fixing it causes CI to fail, this will be handles in a seperate PR.`. So for now its not used.

.gitignore
```diff
@@ -1 +1,2 @@
 /bin/
+.idea
```
lobziik, line 2: nit: can we add new line at the end of this file?

PR #210, opened 2021-05-03
Bug 1954177: mark webhook tests as pending

elmiko:
This change adds the `P` prefix to the `It` clauses for the webhook
tests. This is being done temporarily to allow for MAO webhook
dependency changes to pass the test suite, then they will be revendored
into the test suite, at which point the tests will be re-enabled.

ref: https://github.com/openshift/machine-api-operator/pull/855


alexander-demichev (approved):
/approve

PR #214, opened 2021-07-28
Bug 1989051: Do not set spot max price if empty

JoelSpeed:
Spot tests are currently broken when the spot price is set to the empty string, let's not set the spot price if it's empty.

```
Machine error: failed to create machine "ci-op-xjhkjx76-d57e2-75kbmvgjwn-b85rl" scope: error unmarshaling JSON: while decoding JSON: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'%!(EXTRA string=failed to get machine config)
```


elmiko (approved):
/approve

PR #222, opened 2021-11-08
Mark MAO deployment tests as disruptive

JoelSpeed:
These tests modify the machine api controllers deployment and the mutating/validating webhook configurations. This means that they cannot be run in parallel with any tests that rely on these components. These tests disrupt the functionality of the cluster and must be run in serial to prevent flakes from other tests.


Fedosin (approved):
/lgtm
/approve

PR #228, opened 2022-02-22
Add E2E test for Machine Deletion Hooks

RadekManak:
Tests that pre-drain and pre-terminate hooks prevent pods being evicted and machine deleted respectively.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,10 @@ test-e2e: ## Run openshift specific e2e test
 	# After success, merge all JUnit files into one
 	hack/junitmerge.sh
 
+.PHONY: test-e2e-lifecyclehooks
+test-e2e-lifecyclehooks:
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Lifecycle" || (hack/junitmerge.sh && exit 1)
+
 test-e2e-tech-preview:
 	hack/ci-integration.sh $(GINKGO_ARGS) -focus="TechPreview"
 
--- a/pkg/framework/jobs.go
+++ b/pkg/framework/jobs.go
@@ -1,0 +1,63 @@
+package framework
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName string,
+	testLabel string, nodeSelector string, podLabel string) *batchv1.Job {
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      workloadJobName,
+			Namespace: "default",
+			Labels:    map[string]string{testLabel: ""},
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  workloadJobName,
+							Image: "busybox",
+							Command: []string{
+								"sleep",
+								"86400", // 1 day
+							},
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									"memory": memoryRequest,
+									"cpu":    resource.MustParse("500m"),
+								},
+							},
+						},
+					},
+					RestartPolicy: corev1.RestartPolicy("Never"),
+					Tolerations: []corev1.Toleration{
+						{
+							Key:      "kubemark",
+							Operator: corev1.TolerationOpExists,
+						},
+					},
+				},
+			},
+			BackoffLimit: pointer.Int32Ptr(4),
+			Completions:  pointer.Int32Ptr(njobs),
+			Parallelism:  pointer.Int32Ptr(njobs),
+		},
+	}
+	if nodeSelector != "" {
+		job.Spec.Template.Spec.NodeSelector = map[string]string{
+			nodeSelector: "",
+		}
+	}
+	if podLabel != "" {
+		job.Spec.Template.ObjectMeta.Labels = map[string]string{
+			podLabel: "",
+		}
+	}
+	return job
+}
--- a/pkg/infra/lifecyclehooks.go
+++ b/pkg/infra/lifecyclehooks.go
@@ -1,0 +1,189 @@
+package infra
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+
+	machinev1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	lifecycleHooksTestLabel           = "test.lifecyclehooks.label"
+	lifecycleHooksPodLabel            = "e2e.lifecyclehooks.pod.label"
+	lifecyclehooksWorkerNodeRoleLabel = "machine.openshift.io/lifecyclehooks-e2e-worker"
+	lifecycleWorkloadJobName          = "e2e-lifecyclehooks-workload"
+	pollingInterval                   = 3 * time.Second
+)
+
+var _ = Describe("[Feature:Machine] Lifecycle Hooks should", func() {
+	var client runtimeclient.Client
+	var machineSet *machinev1.MachineSet
+	var workload *batchv1.Job
+	var pod corev1.Pod
+
+	BeforeEach(func() {
+		var err error
+
+		By("Creating the machineset")
+		client, err = framework.LoadClient()
+		// Build machine set parameters
+		expectedReplicas := 1
+		machineSetParams := framework.BuildMachineSetParams(client, expectedReplicas)
+		// Create a label for node and add to machine set parameters
+		machineSetParams.Labels[lifecyclehooksWorkerNodeRoleLabel] = ""
+		// Create machine set
+		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+		Expect(err).ToNot(HaveOccurred())
+		// Wait for machine to be running
+		framework.WaitForMachineSet(client, machineSet.GetName())
+
+		By("Running a workload on the machine")
+		// Run a pod on this machine
+		workloadMemRequest := resource.MustParse("100m")
+		workload = framework.NewWorkLoad(int32(expectedReplicas), workloadMemRequest,
+			lifecycleWorkloadJobName, lifecycleHooksTestLabel, lifecyclehooksWorkerNodeRoleLabel, lifecycleHooksPodLabel)
+		Expect(client.Create(context.Background(), workload)).To(Succeed(), "Could not create workload job")
+
+		By("Waiting for job pod to start running on machine.")
+		Eventually(func() (bool, error) {
+			jobPodList, err := framework.GetPods(client, map[string]string{lifecycleHooksPodLabel: ""})
+			if err != nil {
+				return false, err
+			}
+			if len(jobPodList.Items) == expectedReplicas {
+				pod = jobPodList.Items[0]
+				return pod.Status.Phase == corev1.PodRunning, nil
+			}
+			return false, nil
+		}, framework.WaitLong, pollingInterval).Should(BeTrue(), "Pod did not start running on machine")
+	})
+
+	AfterEach(func() {
+		By("Deleting the machineset")
+		cascadeDelete := metav1.DeletePropagationForeground
+		Expect(client.Delete(context.Background(), machineSet, &runtimeclient.DeleteOptions{
+			PropagationPolicy: &cascadeDelete,
+		})).To(Succeed())
+		By("Deleting workload job")
+		Expect(client.Delete(context.Background(), workload, &runtimeclient.DeleteOptions{
+			PropagationPolicy: &cascadeDelete,
+		})).To(Succeed())
+	})
+
+	It("pause lifecycle actions when present", func() {
+		machines, err := framework.GetMachinesFromMachineSet(client, machineSet)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(machines).To(HaveLen(1), "There should be only one machine")
+		machine := machines[0]
+		podKey := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		machineKey := types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+
+		By("Setting lifecycle hooks on the machine")
+		predrainHook := machinev1.LifecycleHook{
+			Name:  "cluster-api-actuator-pkg/pre-drainHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		preterminateHook := machinev1.LifecycleHook{
+			Name:  "cluster-api-actuator-pkg/pre-terminateHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		Eventually(func() (bool, error) {
+			if err = client.Get(context.Background(), machineKey, machine); err != nil {
+				return false, err
+			}
+			machine.Spec.LifecycleHooks.PreDrain = []machinev1.LifecycleHook{predrainHook}
+			machine.Spec.LifecycleHooks.PreTerminate = []machinev1.LifecycleHook{preterminateHook}
+			if err := client.Update(context.Background(), machine); err != nil {
+				return false, err
+			}
+			return true, nil
+		}, framework.WaitShort, pollingInterval).Should(BeTrue(),
+			"Could not add lifecycle hooks to machine")
+
+		By("Deleting the machine")
+		// Delete the machine by scaling down the machineset to zero
+		framework.ScaleMachineSet(machineSet.Name, 0)
+
+		By("Checking that workload pod is running on machine")
+		// pre-drain hook should prevent pod from being evicted
+		Eventually(func() (bool, error) {
+			if err := client.Get(context.Background(), podKey, &pod); err != nil {
+				return false, err
+			}
+			if err := client.Get(context.Background(), machineKey, machine); err != nil {
+				return false, err
+			}
+			// Check that machine drainable false condition is set
+			for _, condition := range machine.Status.Conditions {
+				if condition.Type == machinev1.MachineDrainable && condition.Status == corev1.ConditionFalse {
+					return pod.Status.Phase == corev1.PodRunning, nil
+				}
+			}
+			return false, nil
+		}, framework.WaitMedium, pollingInterval).Should(BeTrue(),
+			"Workload pod was evicted from the machine or drainable condition is not set")
+
+		By("Removing pre-drain hook")
+		Eventually(func() (bool, error) {
+			if err := client.Get(context.Background(), machineKey, machine); err != nil {
+				return false, err
+			}
+			machine.Spec.LifecycleHooks.PreDrain = []machinev1.LifecycleHook{}
+			if err := client.Update(context.Background(), machine); err != nil {
+				return false, err
+			}
+			return true, nil
+		}, framework.WaitShort, pollingInterval).Should(BeTrue(), "Could not delete pre-drain hook")
+
+		By("Checking that workload pod is evicted from the machine")
+		// Check that pod is evicted, but machine is still present
+		Eventually(func() bool {
+			err = client.Get(context.Background(), podKey, &pod)
+			return apierrors.IsNotFound(err)
+		}, framework.WaitMedium, pollingInterval).Should(BeTrue(), "Pod was not evicted from machine")
+		Eventually(func() (bool, error) {
+			if err := client.Get(context.Background(), machineKey, machine); err != nil {
+				return false, err
+			}
+			// Machine phase should be "Deleting"
+			// but pre-terminate hook should prevent deletion and set terminable false condition
+			for _, condition := range machine.Status.Conditions {
+				if condition.Type == machinev1.MachineTerminable && condition.Status == corev1.ConditionFalse {
+					return *machine.Status.Phase == "Deleting", nil
+				}
+			}
+			return false, nil
+		}, framework.WaitMedium, pollingInterval).Should(BeTrue(),
+			"Machine was deleted or terminable condition is not set")
+
+		By("Removing pre-terminate hook")
+		Eventually(func() (bool, error) {
+			if err := client.Get(context.Background(), machineKey, machine); err != nil {
+				return false, err
+			}
+			machine.Spec.LifecycleHooks.PreTerminate = []machinev1.LifecycleHook{}
+			if err = client.Update(context.Background(), machine); err != nil {
+				return false, err
+			}
+			return true, nil
+		}, framework.WaitShort, pollingInterval).Should(BeTrue(),
+			"Could not delete pre-termiante hook")
+
+		By("Checking that machine is deleted")
+		Eventually(func() bool {
+			err = client.Get(context.Background(), machineKey, machine)
+			return apierrors.IsNotFound(err)
+		}, framework.WaitLong, pollingInterval).Should(BeTrue(), "Machine was not deleted")
+	})
+})
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -14,7 +14,6 @@ import (
 	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
 	caov1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1"
 	caov1beta1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1beta1"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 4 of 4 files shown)


Makefile
```diff
@@ -87,6 +87,10 @@ test-e2e: ## Run openshift specific e2e test
 	# After success, merge all JUnit files into one
 	hack/junitmerge.sh
 
+.PHONY: test-e2e-lifecyclehooks
+lifecyclehooks:
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Lifecycle" || (hack/junitmerge.sh && exit 1)
+
```
JoelSpeed, line 93: Given you've included this in `e2e_test.go` it should be run as part of that suite, probably don't need a separate target for it right?

Makefile
```diff
@@ -87,6 +87,10 @@ test-e2e: ## Run openshift specific e2e test
 	# After success, merge all JUnit files into one
 	hack/junitmerge.sh
 
+.PHONY: test-e2e-lifecyclehooks
+lifecyclehooks:
+	hack/ci-integration.sh $(GINKGO_ARGS) -focus="Lifecycle" || (hack/junitmerge.sh && exit 1)
```
alexander-demichev, line 92: this command is useful to run the test suite separately without having to comment code.

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	lifecycleHooksTestLabel           = "test.lifecyclehooks.label"
+	LifecyclehooksWorkerNodeRoleLabel = "machine.openshift.io/lifecyclehooks-e2e-worker"
```
alexander-demichev, line 25:
```suggestion
	lifecyclehooksWorkerNodeRoleLabel = "machine.openshift.io/lifecyclehooks-e2e-worker"
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
+package lifecyclehooks
```
JoelSpeed, line 1: Wondering if we need a separate package for this or whether we should just move it into `infra`, WDYT?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/openshift/api/machine/v1beta1"
```
JoelSpeed, line 13: We would normally import this as `machinev1` or `machinev1beta1` as an alias, might be good to keep that consistent

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string) *batchv1.Job {
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      workloadJobName,
+			Namespace: "default",
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Job",
+			APIVersion: "batch/v1",
+		},
```
JoelSpeed, line 39: You don't need to specify this information as it will be gathered by the client as you are using a typed structure (*batchv1.Job)

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+								},
+							},
+						},
+					},
+					RestartPolicy: corev1.RestartPolicy("Never"),
+					Tolerations: []corev1.Toleration{
+						{
+							Key:      "kubemark",
+							Operator: corev1.TolerationOpExists,
+						},
```
JoelSpeed, line 67: Interesting, I don't know much about kubemark, where you using kubemark to test this?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+const (
+	lifecycleHooksTestLabel           = "test.lifecyclehooks.label"
+	LifecyclehooksWorkerNodeRoleLabel = "machine.openshift.io/lifecyclehooks-e2e-worker"
+	workloadJobName                   = "e2e-lifecyclehooks-workload"
+	pollingInterval                   = 3 * time.Second
+)
+
+func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string) *batchv1.Job {
```
alexander-demichev, line 30: I see you copied this function from https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/autoscaler/autoscaler.go#L42, can you move it to the framework package and make public `NewWorkLoad()`? this way we will avoid code duplication

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+var _ = Describe("[Feature:Machine] Lifecycle Hooks should", func() {
+	var client runtimeclient.Client
+	var machineSet *v1beta1.MachineSet
+	var machine *v1beta1.Machine
+	var workload *batchv1.Job
+	var pod *corev1.Pod
+	ctx := context.Background()
+	var err error
```
JoelSpeed, line 91: I would probably specify this closer to where it's used rather than "globablly" here. You should only specify things in this list that are shared between the `BeforeEach`, `AfterEach` and `It` blocks

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+
+	BeforeEach(func() {
+		By("Creating the machineset")
+		client, err = framework.LoadClient()
+		// Build machine set parameters
+		expectedReplicas := 1
+		machineSetParams := framework.BuildMachineSetParams(client, expectedReplicas)
+		// Create a label for node and add to machine set parameters
+		targetedNodeLabel := fmt.Sprintf("%v", LifecyclehooksWorkerNodeRoleLabel)
```
JoelSpeed, line 100: Is this label not already a string? I don't think we need to print it (or maybe even use a variable here at all)

pkg/lifecyclehooks/lyfecyclehooks.go
JoelSpeed, line 25: Does this need to be public?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+		Expect(err).ToNot(HaveOccurred())
+		// Wait for machine to be running
+		framework.WaitForMachineSet(client, machineSet.GetName())
+		By("Running a workload on the machine")
+		// Run a pod on this machine
+		workloadMemRequest := resource.MustParse("100000000")
+		workload = newWorkLoad(int32(expectedReplicas), workloadMemRequest, targetedNodeLabel)
+		Expect(client.Create(ctx, workload)).Should(Succeed())
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 112: This seems like it's not needed?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		// Create machine set
+		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+		Expect(err).ToNot(HaveOccurred())
+		// Wait for machine to be running
+		framework.WaitForMachineSet(client, machineSet.GetName())
+		By("Running a workload on the machine")
+		// Run a pod on this machine
+		workloadMemRequest := resource.MustParse("100000000")
+		workload = newWorkLoad(int32(expectedReplicas), workloadMemRequest, targetedNodeLabel)
+		Expect(client.Create(ctx, workload)).Should(Succeed())
```
JoelSpeed, line 111: Nit: This is more idiomatic/flows better IMO
```suggestion
		Expect(client.Create(ctx, workload)).To(Succeed())
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+
+	BeforeEach(func() {
+		By("Creating the machineset")
+		client, err = framework.LoadClient()
+		// Build machine set parameters
+		expectedReplicas := 1
+		machineSetParams := framework.BuildMachineSetParams(client, expectedReplicas)
+		// Create a label for node and add to machine set parameters
+		targetedNodeLabel := fmt.Sprintf("%v", LifecyclehooksWorkerNodeRoleLabel)
+		machineSetParams.Labels[targetedNodeLabel] = ""
```
alexander-demichev, line 101:
```suggestion
		machineSetParams.Labels[lifecyclehooksWorkerNodeRoleLabel] = ""
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+			Namespace: "default",
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Job",
+			APIVersion: "batch/v1",
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{lifecycleHooksTestLabel: ""},
```
alexander-demichev, line 43: you can pass labels as an argument to make reusable in autoscaler tests

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		// Create a pre-drain hook for the machine
+		predrainHook := &v1beta1.LifecycleHook{
+			Name:  "cluster-api-actuator-pkg/pre-drainHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		machines, err := framework.GetMachinesFromMachineSet(client, machineSet)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(machines).To(HaveLen(1))
+		machine = machines[0]
+		machine.Spec.LifecycleHooks.PreDrain = []v1beta1.LifecycleHook{*predrainHook}
```
alexander-demichev, line 145:
```suggestion
		machine.Spec.LifecycleHooks.PreDrain = []v1beta1.LifecycleHook{predrainHook}
		machine.Spec.LifecycleHooks.PreTerminate = []v1beta1.LifecycleHook{preterminateHook}
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		By("Deleting workload job")
+		Expect(client.Delete(ctx, workload, &runtimeclient.DeleteOptions{
+			PropagationPolicy: &cascadeDelete,
+		}))
+	})
+
+	It("pause lifecycle actions when present", func() {
+		By("Setting lifecycle hooks on the machine")
+		// Create a pre-drain hook for the machine
+		predrainHook := &v1beta1.LifecycleHook{
```
alexander-demichev, line 137:
```suggestion
		predrainHook := v1beta1.LifecycleHook{
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		workload = newWorkLoad(int32(expectedReplicas), workloadMemRequest, targetedNodeLabel)
+		Expect(client.Create(ctx, workload)).Should(Succeed())
+		Expect(err).ToNot(HaveOccurred())
+		By("Waiting for job pod to start running on machine.")
+		Eventually(func() bool {
+			jobPodList, err := framework.GetPods(client, map[string]string{lifecycleHooksTestLabel: ""})
+			Expect(err).ToNot(HaveOccurred())
+			pod = &jobPodList.Items[0]
+			return len(jobPodList.Items) == int(expectedReplicas) && pod.Status.Phase == corev1.PodRunning
+		}, framework.WaitMedium, pollingInterval).Should(BeTrue())
```
JoelSpeed, line 119: Don't make assertions inside an eventually function, they fail the test straight away so transient errors break the eventually loop, instead do:
```suggestion
		Eventually(func() (bool, error) {
			jobPodList, err := framework.GetPods(client, map[string]string{lifecycleHooksTestLabel: ""})
			if err != nil {
			  return false, err
			}
			pod := jobPodList.Items[0]
			return len(jobPodList.Items) == int(expectedReplicas) && pod.Status.Phase == corev1.PodRunning, nil
		}, framework.WaitMedium, pollingInterval).Should(BeTrue())
```
The eventually will only succeed when the second parameter is nil, ie there is no error

I don't think pod needs to be a pointer here btw so I changed that too

pkg/lifecyclehooks/lyfecyclehooks.go
alexander-demichev: you can add both hooks in one update call

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return job
+}
+
+var _ = Describe("[Feature:Machine] Lifecycle Hooks should", func() {
+	var client runtimeclient.Client
+	var machineSet *v1beta1.MachineSet
+	var machine *v1beta1.Machine
+	var workload *batchv1.Job
+	var pod *corev1.Pod
```
JoelSpeed, line 89: How many of these actually need to be here, vs, can be declared inline at the time they're used?

pkg/lifecyclehooks/lyfecyclehooks.go
JoelSpeed, line 137: Does this need to be a pointer?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+			Name:  "cluster-api-actuator-pkg/pre-drainHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		machines, err := framework.GetMachinesFromMachineSet(client, machineSet)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(machines).To(HaveLen(1))
+		machine = machines[0]
+		machine.Spec.LifecycleHooks.PreDrain = []v1beta1.LifecycleHook{*predrainHook}
+		err = client.Update(context.Background(), machine)
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 147: This could be susceptible to a transient error (eg someone else changed it on the server side), so there is a pattern for making this bullet proof of doing the update inside an eventually.

You seem to have done this already below so maybe we can do it here too?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		//Create a pre-terminate hook for the machine
+		preterminateHook := &v1beta1.LifecycleHook{
```
JoelSpeed, line 149: Does this need to be a pointer?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+			Name:  "cluster-api-actuator-pkg/pre-terminateHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		Eventually(func() bool {
```
JoelSpeed, line 153: As above, make this return `(bool, error)`

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		}, framework.WaitShort, pollingInterval).Should(BeTrue())
+		By("Deleting the machine")
+		// Delete the machine by scaling down the machineset to zero
+		framework.ScaleMachineSet(machineSet.Name, 0)
+		By("Checking that workload pod is running on machine")
+		// pre-drain hook should prevent pod from being evicted
+		key := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		err = client.Get(ctx, key, pod)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
```
alexander-demichev, line 174: can we have an `Eventually()` block here to make sure that pod was not stopped for at least a couple of minutes?

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+			return true
+		}, framework.WaitShort, pollingInterval).Should(BeTrue())
+		By("Deleting the machine")
+		// Delete the machine by scaling down the machineset to zero
+		framework.ScaleMachineSet(machineSet.Name, 0)
+		By("Checking that workload pod is running on machine")
+		// pre-drain hook should prevent pod from being evicted
+		key := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		err = client.Get(ctx, key, pod)
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 173: Could inline this

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+		By("Removing pre-drain hook")
+		key = types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+		err = client.Get(ctx, key, machine)
+		Expect(err).ToNot(HaveOccurred())
+		machine.Spec.LifecycleHooks.PreDrain = []v1beta1.LifecycleHook{}
+		err = client.Update(context.Background(), machine)
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 181: This should be inside an eventually

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		By("Checking that workload pod is running on machine")
+		// pre-drain hook should prevent pod from being evicted
+		key := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		err = client.Get(ctx, key, pod)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+		By("Removing pre-drain hook")
+		key = types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+		err = client.Get(ctx, key, machine)
+		Expect(err).ToNot(HaveOccurred())
```
alexander-demichev, line 178:
```suggestion
		Expect(client.Get(ctx, key, machine)).To(Succeed())
```

pkg/lifecyclehooks/lyfecyclehooks.go
alexander-demichev, line 173:
```suggestion
		Expect(client.Get(ctx, key, pod)).To(Succeed())
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		By("Deleting the machine")
+		// Delete the machine by scaling down the machineset to zero
+		framework.ScaleMachineSet(machineSet.Name, 0)
+		By("Checking that workload pod is running on machine")
+		// pre-drain hook should prevent pod from being evicted
+		key := types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+		err = client.Get(ctx, key, pod)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
+		By("Removing pre-drain hook")
```
JoelSpeed, line 175: Do we need to add some delay before we do this? How do we know that the hook is blocking the removal and not just a timing thing?

Might be worth checking the conditions (we have Drainable and Terminable) to check they are in the expected state

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		err = client.Get(ctx, key, machine)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(*machine.Status.Phase).To(Equal("Deleting")) // Machine phase should be "Deleting"
+		// but pre-terminate hook should prevent deletion
+		By("Removing pre-terminate hook")
+		key = types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+		err = client.Get(ctx, key, machine)
+		machine.Spec.LifecycleHooks.PreTerminate = []v1beta1.LifecycleHook{}
+		err = client.Update(context.Background(), machine)
+		Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 199: Should be an eventually

pkg/lifecyclehooks/lyfecyclehooks.go
alexander-demichev, line 147:
```suggestion
		Expect(client.Update(context.Background(), machine)).To(Succeed())
```
when function only returns an error you can check the error value without assigning it to a variable

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		machine.Spec.LifecycleHooks.PreDrain = []v1beta1.LifecycleHook{}
+		err = client.Update(context.Background(), machine)
+		Expect(err).ToNot(HaveOccurred())
+		By("Checking that workload pod is evicted from the machine")
+		// Check that pod is evicted, but machine is still present
+		Eventually(func() bool {
+			key = types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}
+			err = client.Get(ctx, key, pod)
+			return apierrors.IsNotFound(err)
```
alexander-demichev, line 187:
```suggestion
			return apierrors.IsNotFound(client.Get(ctx, key, pod))
```

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+		Expect(client.Delete(ctx, machineSet, &runtimeclient.DeleteOptions{
+			PropagationPolicy: &cascadeDelete,
+		})).To(Succeed())
+		By("Deleting workload job")
+		Expect(client.Delete(ctx, workload, &runtimeclient.DeleteOptions{
+			PropagationPolicy: &cascadeDelete,
+		}))
+	})
+
+	It("pause lifecycle actions when present", func() {
```
JoelSpeed, line 134: If possible, add some blank new lines throughout this function where it makes sense to break this down into sections of different actions, eg between each of the `By` sections.

Alternatively, you can put the logic within the `By` which makes it clear what the `By` applies to

```
By("...", funcI() {
  // Do some logic here
})

By("...", funcI() {
  // Do some logic here
})
```

JoelSpeed (commented):
This is looking good, I've added a bunch of suggestions for improving it, most of it is stylistic stuff, let me know if you have any questions

pkg/lifecyclehooks/lyfecyclehooks.go
```diff
@@ -0,0 +1,207 @@
...
+			return apierrors.IsNotFound(err)
+		}, framework.WaitShort, pollingInterval).Should(BeTrue())
+		key = types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+		err = client.Get(ctx, key, machine)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(*machine.Status.Phase).To(Equal("Deleting")) // Machine phase should be "Deleting"
+		// but pre-terminate hook should prevent deletion
+		By("Removing pre-terminate hook")
+		key = types.NamespacedName{Namespace: machine.Namespace, Name: machine.Name}
+		err = client.Get(ctx, key, machine)
```
alexander-demichev, line 196:
```suggestion
		Expect(client.Get(ctx, key, machine)).To(Succeed())
```

alexander-demichev (commented):
/ok-to-test

pkg/infra/lifecyclehooks.go
```diff
@@ -0,0 +1,226 @@
...
+		machineSetParams.Labels[lifecyclehooksWorkerNodeRoleLabel] = ""
+		// Create machine set
+		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+		Expect(err).ToNot(HaveOccurred())
+		// Wait for machine to be running
+		framework.WaitForMachineSet(client, machineSet.GetName())
+
+		By("Running a workload on the machine")
+		// Run a pod on this machine
+		workloadMemRequest := resource.MustParse("100000000")
```
JoelSpeed, line 99: What is this value? Perhaps we should make this `100m` instead? (I think that's correct)

pkg/infra/lifecyclehooks.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		Eventually(func() (bool, error) {
+			Expect(client.Get(context.Background(), machineKey, machine)).To(Succeed())
+			machine.Spec.LifecycleHooks.PreDrain = []machinev1.LifecycleHook{predrainHook}
+			machine.Spec.LifecycleHooks.PreTerminate = []machinev1.LifecycleHook{preterminateHook}
+			err = client.Update(context.Background(), machine)
+			if apierrors.IsConflict(err) {
+				return false, err
+			}
+			Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 154: This can just be condensed to
```suggestion
			if err := client.Update(context.Background(), machine); err != nil {
				return false, err
			}
```

pkg/infra/lifecyclehooks.go
```diff
@@ -0,0 +1,226 @@
...
+		predrainHook := machinev1.LifecycleHook{
+			Name:  "cluster-api-actuator-pkg/pre-drainHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		preterminateHook := machinev1.LifecycleHook{
+			Name:  "cluster-api-actuator-pkg/pre-terminateHook",
+			Owner: "cluster-api-actuator-pkg",
+		}
+		Eventually(func() (bool, error) {
+			Expect(client.Get(context.Background(), machineKey, machine)).To(Succeed())
```
JoelSpeed, line 147: You can't make assertions directly like this, you either need to return the error from this function, or pass a `g` into the function and make assertions on that.

Currently if this fails once, it will fail the whole test

Check https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode for an example of using assertions inside, else just `if err != nil { return bool, error}` which can be less confusing

JoelSpeed (commented):
Please check throughout the test to make sure we aren't make global assertions within an Eventually block, this doesn't work and fails the whole test rather than one iteration of the eventually. Better to just return the error to make the eventually retry

pkg/infra/lifecyclehooks.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+const (
+	lifecycleHooksTestLabel           = "test.lifecyclehooks.label"
+	lifecyclehooksWorkerNodeRoleLabel = "machine.openshift.io/lifecyclehooks-e2e-worker"
+	workloadJobName                   = "e2e-lifecyclehooks-workload"
+	pollingInterval                   = 3 * time.Second
+)
+
+func newWorkLoad(njobs int32, memoryRequest resource.Quantity, nodeSelector string, podLabel string) *batchv1.Job {
```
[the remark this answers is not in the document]
RadekManak: It was copied from autoscaler, but changed in few places. I have added three parameters to the function to be able to reuse it and moved it to framework package.

pkg/infra/lifecyclehooks.go
RadekManak: Oh, changed it in just one place. Another commit incomming.

alexander-demichev (approved):
/approve
/retest

pkg/framework/jobs.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName string,
```
JoelSpeed, line 11: Nit, this should have some sort of comment
```suggestion
// NewWorkload generates a batchv1.Job to deploy a pod which will sleep for one day.
// This can be used to add workloads to the cluster to affect scheduling and excercise
// autoscaling within tests.
func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName string,
```

JoelSpeed (commented):
One nit, can you apply my suggestion and squash the commits, then I'm happy to add the LGTM tag, thanks @RadekManak

RadekManak:
/hold
Test failed on gcp. Probably just gcp being slow.

RadekManak:
/hold cancel
Increased timeout that was cousing test to fail on GCP. Added some test failure comments. Also pre-drain hook had flipped boolean check and would not have caught errors.

alexander-demichev (approved):
/lgtm

RadekManak:
Increasing timeouts once more because tests failed. It seeems that things take longer when the the entire tes suite is running instead of just one test.

PR #230, opened 2022-04-27
[OCPCLOUD-1436] AWS IMDSv2 tests

lobziik:
AWS IMDSv2 e2e tests

this patch

- adds helper for cache platfrom and speedup tests a little
extends mock metadata service to ensure
- that termination handler can always work with IMDSv2
- adds aws specific e2e tests for this feature


elmiko (commented):
this looks good to me, i like the cleanups as well. i have one minor question about a helper function name, but it's not a blocker for me.
/approve

elmiko:
although one of the test failures is coming from the e2e spot tests, i'm not convinced it's related to this pr. i'm not 100% sure though ;)

pkg/framework/framework.go
```diff
@@ -56,6 +56,23 @@ func GetInfrastructure(c runtimeclient.Client) (*configv1.Infrastructure, error)
...
+func GetPlatform(c runtimeclient.Client) (configv1.PlatformType, error) {
+	// platform won't change during test run and might be cached
+	if platform != "" {
+		return platform, nil
+	}
+	infra, err := GetInfrastructure(c)
+	if err != nil {
+		return "", err
+	}
+	platform = infra.Status.PlatformStatus.Type
```
JoelSpeed, line 72: PlatformStatus is a pointer, you must guard against a nil pointer before you deref it here

pkg/framework/pods.go
```diff
@@ -13,3 +17,54 @@ func GetPods(client runtimeclient.Client, selector map[string]string) (*corev1.P
...
 	return pods, err
 }
+
+type PodCleanupFunc func() error
+
+type PodLastLogFunc func(container string, lines int, previous bool) (string, error)
+
+// SpinPodOnNode spins a pod according passed spec on particular node.
+// returns created pod object, function for retrieve last logs, cleanup function and error if occurred
+func SpinPodOnNode(clientset *kubernetes.Clientset, node *corev1.Node, namespace string, podSpec corev1.PodSpec) (*corev1.Pod, PodLastLogFunc, PodCleanupFunc, error) {
```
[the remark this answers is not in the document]
JoelSpeed: I would go for `RunPodOnNode` or `CreatePodOnNode`

pkg/providers/aws.go
```diff
@@ -0,0 +1,122 @@
...
+						Command: []string{"curl"},
+						Args:    []string{"-v", amiIdMetadataEndpoint},
+					},
+				},
+			}
+			pod, lastLog, cleanupPod, err := framework.SpinPodOnNode(clientset, nodes[0], framework.MachineAPINamespace, podSpec)
+			Expect(err).ToNot(HaveOccurred())
+			defer cleanupPod()
+
+			By("Ensure curl pod rdy")
```
JoelSpeed, line 84: Please write in full english here, not need to abbreviate ready

lobziik:
> although one of the test failures is coming from the e2e spot tests, i'm not convinced it's related to this pr. i'm not 100% sure though ;)

It's related to this pr, this patch breaks it. 
Precisely this -> 
https://github.com/openshift/cluster-api-actuator-pkg/pull/230/files#diff-e48816ad183ec302bf4d14b6334db1d3bdcd9a7e634e069ab9dc5f96cb09e27cR47

colon matters.

pkg/framework/framework.go
lobziik: I don't feel that i MUST do it there :D 
This thing will crash at the very start of the test suite with a stacktrace which will point right there.

 But good catch, I will add guard, np.

pkg/framework/pods.go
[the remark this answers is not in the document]
lobziik: renamed

PR #232, opened 2022-06-09
Bug 2060068: add security groups to the minimal AWS prover spec

Fedosin:
We are going to make this field mandatore in the spec, so tests must specify it.

### DIFF
--- a/pkg/infra/webhooks.go
+++ b/pkg/infra/webhooks.go
@@ -188,6 +188,7 @@ func minimalAWSProviderSpec(ps *machinev1.ProviderSpec) (*machinev1.ProviderSpec
 				Placement:          fullProviderSpec.Placement,
 				Subnet:             *fullProviderSpec.Subnet.DeepCopy(),
 				IAMInstanceProfile: fullProviderSpec.IAMInstanceProfile.DeepCopy(),
+				SecurityGroups:     fullProviderSpec.SecurityGroups,
 			},
 		},
 	}, nil


alexander-demichev (approved):
/approve

elmiko (commented):
/lgtm

JoelSpeed:
I have questions about this fix:
* How can we change a field to be mandatory once it has already shipped?
* Do we have any validations of this field in the webhooks today?
* Do we do any defaulting of this field in the webhooks today?
* Why does this need to be required when it previously has been optional?

/hold

Fedosin:
@JoelSpeed 

> How can we change a field to be mandatory once it has already shipped?

Well, technically security groups are mandatory now, because if you don't specify them, the machine has networking issues and doesn't work properly (hence the bug!). So, what we try to achieve is to report the error earlier, during initial validation, without creating a real instance.

> Do we have any validations of this field in the webhooks today?

Not in the webhooks, but before the machine controller begins provisioning the machine, it fetches info about its SGs from AWS. If there is nothing, it logs an error, but doesn't return the error: https://github.com/openshift/machine-api-provider-aws/blob/main/pkg/actuators/machine/instances.go#L95-L97 

> Do we do any defaulting of this field in the webhooks today?

No, we haven't. I don't mind adding there some default SG filters as it improves the situation, but we need to make sure that provided filters return something anyway.

> Why does this need to be required when it previously has been optional?

As I said in the beginning, it has always been mandatory, as these values are set by default in the installer, and a machine created without SGs doesn't work properly.

JoelSpeed:
/hold cancel

I think we will want to warn in the webhooks when no security groups are provided, let's not attempt a default as that's not going to be easy to get right. Can you please make sure to add that warning

JoelSpeed:

This change doesn't affect Azure and it has some issues with stability in CI at the moment

PR #238, opened 2022-07-20
add failed machine detection to WaitForMachineSets

elmiko:
this change makes it so that any test using WaitForMachineSets will exit
if any of the machines come up in failed phase. it will print the failed
machines and exit the test.

### DIFF
--- a/pkg/framework/machinesets.go
+++ b/pkg/framework/machinesets.go
@@ -294,7 +294,8 @@ func getScaleClient() (scale.ScalesGetter, error) {
 
 // WaitForMachineSet waits for the all Machines belonging to the named
 // MachineSet to enter the "Running" phase, and for all nodes belonging to those
-// Machines to be ready.
+// Machines to be ready. If a Machine is detected in "Failed" phase, the test
+// will exit early.
 func WaitForMachineSet(c client.Client, name string) {
 	machineSet, err := GetMachineSet(c, name)
 	Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


pkg/framework/machinesets.go
```diff
@@ -312,6 +313,16 @@ func WaitForMachineSet(c client.Client, name string) {
 				name, len(machines), int(replicas))
 		}
 
+		failed := FilterMachines(machines, MachinePhaseFailed)
+		if len(failed) > 0 {
+			// if there are failed machines, print them out and exit the test
+			msg := fmt.Sprintf("found %d Machines in failed phase: ", len(failed))
+			for _, m := range failed {
+				msg = fmt.Sprintf("%s %s", msg, m.Name)
+			}
```
JoelSpeed, line 322: I wonder if we can dump the Machine status out here as well, it would be good to see why the Machine failed. Don't we normally put an error message on the status when the machine fails?

elmiko: that's a good suggestion, i'll poke around to see if it can be more consistent with other failure messages.

elmiko: i looked through some must-gathers and it looks like we do put something in the `status.conditions`, i wonder if we should just print all the available `message` fields from those for each machine?

elmiko: i'm doing a little more digging in MAO to see if i can come up with a strategy for this.

elmiko:
@JoelSpeed i did some digging and it looks like we do set `Machine.status.errorReason` and/or `Machine.status.errorMessage` depending on the type of failure. it looks like we will set these whenever there is a failure cause available while updating status[0], so that seems fairly conclusive. i don't think we need to parse the conditions, although that would be an additional layer we could add.

[0] https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/machine/controller.go#L467

JoelSpeed:
If we can fix the govet issues I think this LGTM

elmiko:
oh weird, i didn't see that govet error before. i'll take a look

elmiko:
updated to fix the go vet error

JoelSpeed:
I'd like to see what the test results look like once this run is done, but otherwise LGTM

elmiko:
sounds good, i'm cautiously optimistic that this will help us identify test failures better across the board.

PR #239, opened 2022-09-02
Ensure tests can run when Pod security admission policy is enabled

JoelSpeed:
While reviewing some test failures, I noticed that our workloads weren't setting up correctly.

I found the following log in the KCM:
```
"Event occurred" object="default/pdb-workload" fieldPath="" kind="ReplicationController" apiVersion="v1" type="Warning" reason="FailedCreate" message="Error creating: pods \"pdb-workload-cm797\" is forbidden: violates PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"work\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container \"work\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"work\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"work\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
```

To mitigate this problem, we need to run the pods in a privileged namespace, such as the Machine API namespace.

### DIFF
--- a/pkg/framework/jobs.go
+++ b/pkg/framework/jobs.go
@@ -13,7 +13,7 @@ func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName s
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      workloadJobName,
-			Namespace: "default",
+			Namespace: MachineAPINamespace,
 			Labels:    map[string]string{testLabel: ""},
 		},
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/hold cancel
/retest-required

This appears to have fixed the issue we were having but there were other failures

JoelSpeed:
Have reviewed the test results and while the autoscaler tests failed, the rest of the suite is fixed

PR #240, opened 2022-09-07
Ensure proxy test pods run in privileged namespace

JoelSpeed:
This test is currently not working as expected (and also not failing?) as the mitm-proxy daemonset was being deployed to the default namespace.

If instead we deploy to our own namespace, we don't need to worry about the pod security admission policy and can fix the tests for the short term.

In the long term we will want to work out how to reduce these privileges and remove the need for the privileged namespace

### DIFF
--- a/pkg/operators/machine-api-operator.go
+++ b/pkg/operators/machine-api-operator.go
@@ -253,9 +253,6 @@ var _ = Describe("[Serial][Feature:Operators][Disruptive] When cluster-wide prox
 
 		By("waiting for machine-api-controller deployment to reflect unconfigured cluster-wide proxy")
 		Expect(framework.WaitForProxyInjectionSync(client, maoManagedDeployment, framework.MachineAPINamespace, false)).To(BeTrue())
-
-		err = framework.DestroyClusterProxy(client)
-		Expect(err).ToNot(HaveOccurred())
 	})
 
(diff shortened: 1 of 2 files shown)


pkg/operators/machine-api-operator.go
```diff
@@ -205,7 +205,7 @@ var _ = Describe("[Feature:Operators] Machine API cluster operator status should
 	})
 })
 
-var _ = Describe("[Serial][Feature:Operators][Disruptive] When cluster-wide proxy is configured, Machine API cluster operator should ", func() {
+var _ = FDescribe("[Serial][Feature:Operators][Disruptive] When cluster-wide proxy is configured, Machine API cluster operator should ", func() {
```
lobziik, line 208: Not sure that this test should be focused...

lobziik (commented):
Dunno how i reviewed it before... 
lets remove approve for now, focus on this particular test doesnt seems to be right. Also it's still failing, need investigate, will try to take a look tomorrow.

/approve cancel

lobziik:
/approve

failed autoscaler test on aws does not seem relevant to this change.

damdo (commented):
/lgtm

JoelSpeed:

These are failing in the autoscaler tests and the proxy tests worked as expected, so I think this is good to go

PR #241, opened 2022-09-27
update autoscaler max nodes test

elmiko:
this change makes it so that the max nodes test will observe only the ready nodes in the cluster when determining if the maximum has been reached. the autoscaler will only count ready nodes when performing its calculations, when this is combined with tests that might not clean up the cluster properly, or may leave nodes in a non schedulable state, then the test need to take into account only the ready nodes when calculating the maximum size.

this PR also changes the workload job names that are run in the autoscaler tests to ensure that they are unique from each other.

### DIFF
--- a/pkg/framework/jobs.go
+++ b/pkg/framework/jobs.go
@@ -41,6 +41,10 @@ func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName s
 							Key:      "kubemark",
 							Operator: corev1.TolerationOpExists,
 						},
+						{
+							Key:    ClusterAPIActuatorPkgTaint,
+							Effect: corev1.TaintEffectPreferNoSchedule,
+						},
 					},
 				},
 			},
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ else
 	  -e "GO111MODULE=$(GO111MODULE)" \
 	  -e "GOFLAGS=$(GOFLAGS)" \
 	  -e "GOPROXY=$(GOPROXY)" \
-	  openshift/origin-release:golang-1.18
+	  registry.ci.openshift.org/openshift/release:golang-1.18
   IMAGE_BUILD_CMD = $(ENGINE) build
 endif
 
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -46,7 +46,7 @@ func clusterAutoscalerResource(maxNodesTotal int) *caov1.ClusterAutoscaler {
 	// and that has high least common multiple to avoid a case
 	// when a node is considered to be empty even if there are
 	// pods already scheduled and running on the node.
-	unneededTimeString := "23s"
+	unneededTimeString := "60s"
 	return &caov1.ClusterAutoscaler{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "default",
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -293,8 +293,9 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 			Expect(client.Create(ctx, asr)).Should(Succeed())
 			cleanupObjects[asr.GetName()] = asr
 
-			By(fmt.Sprintf("Creating scale-out workload: jobs: %v, memory: %s", expectedReplicas, workloadMemRequest.String()))
-			workload := framework.NewWorkLoad(expectedReplicas, workloadMemRequest, workloadJobName, autoscalingTestLabel, targetedNodeLabel, "")
+			uniqueJobName := fmt.Sprintf("%s-scale-from-zero", workloadJobName)
+			By(fmt.Sprintf("Creating scale-out workload %s: jobs: %v, memory: %s", uniqueJobName, expectedReplicas, workloadMemRequest.String()))
+			workload := framework.NewWorkLoad(expectedReplicas, workloadMemRequest, uniqueJobName, autoscalingTestLabel, targetedNodeLabel, "")
 			cleanupObjects[workload.GetName()] = workload
 			Expect(client.Create(ctx, workload)).Should(Succeed())
 
@@ -353,9 +354,10 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 			}
 
 			jobReplicas := expectedReplicas * int32(2)
-			By(fmt.Sprintf("Creating scale-out workload: jobs: %v, memory: %s",
-				jobReplicas, workloadMemRequest.String()))
-			workload := framework.NewWorkLoad(jobReplicas, workloadMemRequest, workloadJobName, autoscalingTestLabel, targetedNodeLabel, "")
+			uniqueJobName := fmt.Sprintf("%s-cleanup-after-scale-down", workloadJobName)
+			By(fmt.Sprintf("Creating scale-out workload %s: jobs: %v, memory: %s",
+				uniqueJobName, jobReplicas, workloadMemRequest.String()))
+			workload := framework.NewWorkLoad(jobReplicas, workloadMemRequest, uniqueJobName, autoscalingTestLabel, targetedNodeLabel, "")
 			cleanupObjects[workload.GetName()] = workload
 			Expect(client.Create(ctx, workload)).Should(Succeed())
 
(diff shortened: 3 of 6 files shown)


elmiko:
https://github.com/openshift/kubernetes-autoscaler/pull/241 will want to know when this merges

also, i am amused that these both have the same pull number

elmiko:
updated to add a filter for unschedulable nodes as well

elmiko:
looking at the scale to/from zero tests, it appears that sometimes the autoscaler does not think it should remove the nodes even after the test workload has been deleted. my current hypothesis is that /something/ is being scheduled to those nodes (could be a core operator or similar) that is causing the autoscaler to not delete them. if we look at successful runs of this test we see this in the output:

```
STEP: Waiting for machineSet replicas to scale in. Current replicas are 3, expected 0.
I0725 15:27:08.616361   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: removing empty node "ip-10-0-228-209.us-east-2.compute.internal"
I0725 15:27:08.629561   13940 autoscaler.go:224] ip-10-0-228-209.us-east-2.compute.internal: marked the node as toBeDeleted/unschedulable
I0725 15:27:08.631358   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: removing empty node "ip-10-0-231-145.us-east-2.compute.internal"
I0725 15:27:08.650275   13940 autoscaler.go:224] dns-default-fhp8g: deleting pod for node scale down
I0725 15:27:08.658258   13940 autoscaler.go:224] ip-10-0-231-145.us-east-2.compute.internal: marked the node as toBeDeleted/unschedulable
I0725 15:27:08.669076   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: removing empty node "ip-10-0-198-234.us-east-2.compute.internal"
I0725 15:27:08.678413   13940 autoscaler.go:224] dns-default-n2pbv: deleting pod for node scale down
I0725 15:27:08.686565   13940 autoscaler.go:224] ip-10-0-198-234.us-east-2.compute.internal: marked the node as toBeDeleted/unschedulable
I0725 15:27:08.701542   13940 autoscaler.go:224] dns-default-hcwdz: deleting pod for node scale down
I0725 15:27:09.821073   13940 autoscaler.go:224] ip-10-0-228-209.us-east-2.compute.internal: node removed by cluster autoscaler
I0725 15:27:09.833105   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: empty node ip-10-0-228-209.us-east-2.compute.internal removed
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0.
I0725 15:27:10.628454   13940 autoscaler.go:224] ip-10-0-231-145.us-east-2.compute.internal: node removed by cluster autoscaler
I0725 15:27:10.628496   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: empty node ip-10-0-231-145.us-east-2.compute.internal removed
I0725 15:27:11.227274   13940 autoscaler.go:224] ip-10-0-198-234.us-east-2.compute.internal: node removed by cluster autoscaler
I0725 15:27:11.233313   13940 autoscaler.go:224] cluster-autoscaler-status: Scale-down: empty node ip-10-0-198-234.us-east-2.compute.internal removed
STEP: Waiting for machineSet replicas to scale in. Current replicas are 0, expected 0.
```

this output shows that the autoscaler emitted events to remove all the nodes. but, if we look at a failure, we see this:

```
 STEP: Waiting for machineSet replicas to scale in. Current replicas are 3, expected 0.
I0928 14:50:10.567005   13834 autoscaler.go:224] cluster-autoscaler-status: Scale-down: removing empty node "ip-10-0-225-90.ec2.internal"
I0928 14:50:10.584090   13834 autoscaler.go:224] ip-10-0-225-90.ec2.internal: marked the node as toBeDeleted/unschedulable
I0928 14:50:10.588194   13834 autoscaler.go:224] dns-default-9wmsn: deleting pod for node scale down
STEP: Waiting for machineSet replicas to scale in. Current replicas are 3, expected 0.
I0928 14:50:11.769735   13834 autoscaler.go:224] ip-10-0-225-90.ec2.internal: node removed by cluster autoscaler
I0928 14:50:11.776553   13834 autoscaler.go:224] cluster-autoscaler-status: Scale-down: empty node ip-10-0-225-90.ec2.internal removed
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0.
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0.
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0.
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0.
I0928 14:50:23.583593   13834 autoscaler.go:224] oauth-openshift-fd8c57bd5-kmmn8: pod didn't trigger scale-up: 1 node(s) didn't match Pod's node affinity/selector
I0928 14:50:23.589126   13834 autoscaler.go:224] apiserver-5775898b65-djq2v: pod didn't trigger scale-up: 1 node(s) didn't match Pod's node affinity/selector
I0928 14:50:23.593609   13834 autoscaler.go:224] route-controller-manager-5bfc67fb97-qrrdq: pod didn't trigger scale-up: 1 node(s) didn't match Pod's node affinity/selector
I0928 14:50:23.598202   13834 autoscaler.go:224] apiserver-5b96744cdb-g6xbs: pod didn't trigger scale-up: 1 node(s) didn't match Pod's node affinity/selector
STEP: Waiting for machineSet replicas to scale in. Current replicas are 2, expected 0. 
```

we can see that the autoscaler only considers removing a single node and also it thinks that these other pods should be scheduled somewhere. i'm not sure if those other pods are causing part of the issue as they might schedule onto the new machineset and prevent a scale down. we have had similar bugs related to OLM pods, but i had thought we solved those issues.

it also seems like this problem is intermittent as the test will pass sometimes.

elmiko:
i'm starting to think we need to bump the k8s version here as well. seeing a bunch of failures around PDBs having the wrong version.

elmiko:
updated the deps for k8s 1.25 and MAO/CAO, let's see how this works

elmiko:
updated with rebase of cao

elmiko:
seems like the previous test run did not include the latest CAO changes
/retest

JoelSpeed:
AWS jobs passed so looks like the autoscaler tests are fixed now
/lgtm
/approve

damdo (commented):
/lgtm

JoelSpeed:

Need to investigate these webhook issues, but I think the autoscaler improvements here are valuable to merge

elmiko:
that last couple failures on e2e-aws-operator appear to have been because the autoscaler scale to zero test was waiting for a node to finish scaling in, but i never see a corresponding event from the autoscaler to signal that it will remove the node. this makes me think that either something is on that node that is preventing it from scaling down or the workload hasn't finished yet (which would be odd).

elmiko:
i'm hacking on a small change to enforce a taint on the machinesets we create that the test workload can tolerate. i have it working locally but need to add a few more tests.
i guess i also need a rebase or something here.

elmiko:
not sure what happened here, i tried to rebase my changes and now it's all messed up

elmiko:
i see what i did, i'll fix it shortly

pkg/framework/jobs.go
```diff
@@ -41,6 +41,10 @@ func NewWorkLoad(njobs int32, memoryRequest resource.Quantity, workloadJobName s
 							Key:      "kubemark",
 							Operator: corev1.TolerationOpExists,
 						},
+						{
+							Key:    ClusterAPIActuatorPkgTaint,
+							Effect: corev1.TaintEffectPreferNoSchedule,
```
JoelSpeed, line 46: I think we want a NoSchedule rather than prefer, prefer means we might still get random workloads on here right? You said you had some issues with this IIRC? Maybe you can expand on those here

JoelSpeed (commented):
The taints and tolerations prevent other workloads running on our hosts, but they don't guarantee that our workloads run on our hosts.
We should add a specific label to the Machine template (so it ends up on the node) and then add a `nodeSelector` to the workloads pods so that they only schedule on nodes with that label

pkg/framework/jobs.go
elmiko: i tried NoSchedule at it did not work as i expected, which is why i backed off the PreferredNoSchedule. i have a feeling something else is not properly tolerating NoSchedule and causing the nodes to get deprovisioned. we should figure it out, but this current patch is an incremental improvement imo.

elmiko:
>  We should add a specific label to the Machine template (so it ends up on the node) and then add a `nodeSelector` to the workloads pods so that they only schedule on nodes with that label

we are doing that already, unless i am misunderstanding

https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/framework/jobs.go#L52

JoelSpeed:
> > We should add a specific label to the Machine template (so it ends up on the node) and then add a `nodeSelector` to the workloads pods so that they only schedule on nodes with that label
> 
> we are doing that already, unless i am misunderstanding
> 
> https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/framework/jobs.go#L52

Positional arguments are hard!

/lgtm

PR #242, opened 2022-09-30
Migrate PDB to policy v1

JoelSpeed:
Policy v1beta1 stopped being served as a part of K8s 1.25.
This means this test is now failing since the 1.25 rebase landed.

This is blocking the tests from getting to the autoscaler tests


elmiko (commented):
/lgtm

JoelSpeed:

The test results show that the drain test is now fixed, however, the autoscaler tests are still broken so we aren't passing just yet

PR #243, opened 2022-09-30
OCPCLOUD-1427: Use BalancingIgnoredLabels in autoscaler test

RadekManak:
The autoscaler tests were removing UUID label from the machines to allow them to balance properly when scaling. I've changed that to use the new BalancingIgnoredLabels CRD option to tests it.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -395,6 +395,10 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func() 
 			By("Creating ClusterAutoscaler")
 			clusterAutoscaler = clusterAutoscalerResource(100)
 			clusterAutoscaler.Spec.BalanceSimilarNodeGroups = pointer.Bool(true)
+			// Ignore this label to make test nodes similar
+			clusterAutoscaler.Spec.BalancingIgnoredLabels = []string{
+				"e2e.openshift.io",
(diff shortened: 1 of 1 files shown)


JoelSpeed:
@RadekManak I suspect given the failures across all platforms that this isn't actually working as expected, can you investigate the test failures and see if you can see what went wrong? Is the autoscaler being configured correctly for example

RadekManak:
I think this is broken as well. Will take a look when I have time.

RadekManak:
There was a check in the test to make sure the labels are the same. I am now ignoring it in the comparison to test the Ignored labels feature.

JoelSpeed:
Good catch, lets see if the tests pass 👀

JoelSpeed:
/test e2e-aws-operator

/lgtm
/approve

GCP passed so I think this should be good

RadekManak:
Let's give this PR another try, now that the test is improved.

elmiko:
i wonder if we want to merge #272 before this?

RadekManak:
I heard you talking about these changes and thought it was merged. Let's wait for that PR.
/hold

elmiko:
ack, thanks Radek!

JoelSpeed:
The dependent PR is now merged, but this now has merge conflicts which will need to be resolved

elmiko (commented):
/lgtm

PR #245, opened 2022-10-06
Check webhook configurations are synced before running webhook tests

JoelSpeed:
We quite often see that the webhook tests in this suite fail, eg we expect an error but nothing happens.
We aren't checking in the `BeforeEach` that the webhook configurations we depend on are actually present and correct.
This PR adds a check to wait for a short period to make sure the webhook configurations are synced up.

This will ensure that, if any test prior to this one disrupted the webhooks, that we give the operator some time to restore the webhooks before we attempt to use them.


damdo (commented):
Thanks
/lgtm

JoelSpeed:

Related tests passed, both failed on autoscaler tests

/approve

PR #246, opened 2022-10-20
Wait for Machines to be deleted before completeing each test case

JoelSpeed:
@elmiko recently reported that he was seeing cordoned Machines interfering with tests in the autoscaler suite. When running the suite against a test cluster I observed that MachineSets were deleted but the suite moved onto new tests before the old Machines had gone away.

When writing tests we should always make sure we leave the environment clean _before_ we declare the test finished.

This PR adds ensures that every test we have that creates a MachineSet, has a `WaitForMachineSetsDeleted` call in an after each when the test finishes. This helper looks for Machines owned by the MachineSet and only finishes successfully once all of those Machines have gone away.

This should ensure we lave the Machine state clean after these tests.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -198,8 +198,23 @@ var _ = Describe("[Feature:Machines] Autoscaler should", func() {
 	})
 
 	AfterEach(func() {
+		var machineSets []*machinev1.MachineSet
+
 		for name, obj := range cleanupObjects {
+			if machineSet, ok := obj.(*machinev1.MachineSet); ok {
+				// Once we delete a MachineSet we should make sure that the
+				// all of its machines are deleted as well.
+				// Collect MachineSets to wait for.
+				machineSets = append(machineSets, machineSet)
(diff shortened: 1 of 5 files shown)


damdo (commented):
/lgtm

JoelSpeed:
Going to run these again to see if we get similar results, would like to try this with a patch for taints/tolerations on our workload machinesets as I think other pods being scheduled on our nodes is interfering with the tests
/retest-required

JoelSpeed:
Both of these tests failed around this block https://github.com/openshift/cluster-api-actuator-pkg/blob/aa33a2b7aca310c56a246400e39c729cc8667e60/pkg/autoscaler/autoscaler.go#L596-L605

Seems the workloads aren't triggering scale up but we don't have any logs from the autoscaler to tell us why. Mike is working on making sure we have taints, node selectors etc to fix this issue

damdo (commented):
/lgtm

JoelSpeed:
The tests passed! 😱 
/approve

PR #247, opened 2022-10-21
Remove unused event counters and its auxiliary code from the autoscaler tests

lobziik:
Apparently this code is not using anywhere. Maybe worth to remove it.


elmiko (commented):
thanks @lobziik 
/approve

PR #249, opened 2022-10-24
Disable proxy test until periodics are set up

JoelSpeed:
We are shortly going to introduce periodic tests for this test suite, and this test is the prime candidate for moving to periodic. I would like to disable it for now to reduce the likelihood of our test suites timing out (which we are seeing frequently on Azure).

This test takes upwards of 20 minutes to complete and because of its disruptive nature, cannot be run in parallel. It is very expensive to run on presubmits and it doesn't provide that much value to us. We haven't had any issues with the proxy configuration since it was set up some releases ago so I'm confident that disabling this for a few weeks won't cause an issue.

Periodics jira tracking: https://issues.redhat.com/browse/OCPCLOUD-1732


damdo (commented):
/lgtm

damdo (commented):
/lgtm

JoelSpeed:

AWS passed! Azure failed on the penultimate autoscaler test which we've been seeing frequently. Lets merge this as I think it's helping the pass rate

PR #254, opened 2022-11-07
Replace busybox image from dockerhub with ubi-minimal from RH registry

lobziik:
There are rate limits for dockerhub, which causes tests to fail.

https://github.com/openshift/machine-api-operator/pull/1082 - vsphere-operator job here, for example


damdo (commented):
/lgtm

elmiko (commented):
nice change
/approve

PR #255, opened 2022-11-11
OCPCLOUD-1733: Try alternative instance type when spot has no capacity

RadekManak:
This PR rewrites machineSet setup logic for spot test to allow it to detect when one of the machines cannot scale up due to insufficient spot capacity. If that happens, then it gets deleted, and it attempts again with a new machineSet that has a different instance type. 
This is attempted 3 times. If it cannot scale after that, the test fails.

### DIFF
--- a/pkg/framework/machinesets.go
+++ b/pkg/framework/machinesets.go
@@ -2,16 +2,22 @@ package framework
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/google/uuid"
 	. "github.com/onsi/gomega"
 
+	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/scale"
@@ -35,6 +41,17 @@ type MachineSetParams struct {
 
 const (
 	machineAPIGroup = "machine.openshift.io"
+)
+
+var (
+	// ErrMachineNotProvisionedInsufficientCloudCapacity is used when we detect that the machine is not being provisioned due to insufficient provider capacity.
+	ErrMachineNotProvisionedInsufficientCloudCapacity = errors.New("machine creation failed due to insufficient cloud provider capacity")
+
+	// errTestForPlatformNotImplemented is used when platform specific test is run on a platform that does not have it implemented.
+	errTestForPlatformNotImplemented = errors.New("test for current platform not implemented")
+
+	// errMachineInMachineSetFailed is used when one of the machines in the machine set is in a failed state.
+	errMachineInMachineSetFailed = errors.New("machine in the machineset is in a failed phase")
 )
 
 func BuildMachineSetParams(client runtimeclient.Client, replicas int) MachineSetParams {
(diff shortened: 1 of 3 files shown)


pkg/framework/machinesets.go
```diff
@@ -37,6 +43,10 @@ const (
 	machineAPIGroup = "machine.openshift.io"
 )
 
+var (
+	ErrMachineCreateFailedInsufficientCapacity = errors.New("machine creation failed due to insufficient capacity")
```
JoelSpeed, line 47: Does this need to be exported? Please make sure to add a comment describing what this is

pkg/infra/spot.go
```diff
@@ -65,14 +60,27 @@ var _ = Describe("[Feature:Machines] Running on Spot", func() {
...
+				machineSet, err = framework.CreateMachineSet(client, machineSetParams)
+				Expect(err).ToNot(HaveOccurred())
+				delObjects[machineSet.Name] = machineSet
+
+				err = framework.WaitForSpotMachineSet(client, machineSet.GetName())
+				if err == framework.ErrMachineCreateFailedInsufficientCapacity {
+					// If machineSet cannot scale up due to insufficient capacity, try again with different machineSetParams
+					framework.WaitForMachineSetDelete(client, machineSet)
+					continue
+				}
```
JoelSpeed, line 78: If we add a new `By()` in here we can add something to the test output when we have to recreate a new MachineSet. Will be helpful to identify when we have to iterate

JoelSpeed (commented):
Spot tests passed and the approach looks good, will need another set of eyes for a thorough review but the outline looks great

pkg/framework/machinesets.go
```diff
@@ -109,6 +120,75 @@ func CreateMachineSet(c client.Client, params MachineSetParams) (*machinev1.Mach
...
+// BuildMachineSetParamsList creates a list of MachineSetParams based on the given machineSetParams with modified instance type.
+func BuildAlternativeMachineSetParams(machineSetParams MachineSetParams, platform configv1.PlatformType) []MachineSetParams {
+	baseMachineSetParams := machineSetParams
+	baseProviderSpec := baseMachineSetParams.ProviderSpec.DeepCopy()
+
+	output := []MachineSetParams{}
+	switch platform {
+	case configv1.AWSPlatformType:
+		// Using cheapest compute optimized instances that meet openshift minimum requirements (4 vCPU, 8GiB RAM)
+		alternativeInstanceTypes := []string{"c5.xlarge", "c5a.xlarge", "m5.xlarge"}
```
lobziik, line 132: Just thought: m6i.large also might be added here, I think.

pkg/framework/machinesets.go
```diff
@@ -369,6 +449,106 @@ func WaitForMachineSetDelete(c runtimeclient.Client, machineSet *machinev1.Machi
...
+// hasInsufficientCapacity return true if the machine cannot be provisioned due to insufficient spot capacity.
+func hasInsufficientCapacity(m *machinev1.Machine, platform configv1.PlatformType) bool {
+	switch platform {
+	case configv1.AWSPlatformType:
+		awsProviderStatus := machinev1.AWSMachineProviderStatus{}
+		if m.Status.ProviderStatus != nil {
+			err := json.Unmarshal(m.Status.ProviderStatus.Raw, &awsProviderStatus)
+			Expect(err).ToNot(HaveOccurred())
+			for _, condition := range awsProviderStatus.Conditions {
+				if condition.Type == string(machinev1.MachineCreation) && condition.Status == metav1.ConditionFalse && strings.Contains(condition.Message, "InsufficientInstanceCapacity") {
```
lobziik, line 530: nit: I think about moving error texts to constants, wdyt?

lobziik: Maybe generalize this condition and extract it to some function, instead of having two quite similar but slightly different looong lines?

lobziik (commented):
overall LGTM
Just couple of nits

pkg/framework/machinesets.go
```diff
@@ -369,6 +449,106 @@ func WaitForMachineSetDelete(c runtimeclient.Client, machineSet *machinev1.Machi
...
+					reason = string(*m.Status.ErrorReason)
+				}
+				message := "failureMessage not present in Machine.status"
+				if m.Status.ErrorMessage != nil {
+					message = string(*m.Status.ErrorMessage)
+				}
+				klog.Errorf("Failed machine: %s, Reason: %s, Message: %s", m.Name, reason, message)
+			}
+		}
+		Expect(len(failed)).To(Equal(0))
```
Fedosin, line 487: Probably we shouldn't call ginkgo functions here directly - just return an error if len(failed) > 0

pkg/framework/machinesets.go
```diff
@@ -369,6 +449,106 @@ func WaitForMachineSetDelete(c runtimeclient.Client, machineSet *machinev1.Machi
 	WaitForMachineSetsDeleted(c, machineSet)
 }
 
+// WaitForSpotMachineSet waits for all Machines belonging to the machineSet to be running and their nodes to be ready.
+// Unlike WaitForMachineSet, this function does not fail the test when machine cannoct be provisioned due to insufficient spot capacity.
+func WaitForSpotMachineSet(c client.Client, name string) error {
+	machineSet, err := GetMachineSet(c, name)
+	Expect(err).ToNot(HaveOccurred())
```
Fedosin, line 456: I would recommend to return an error here instead of calling `Expect`.

pkg/framework/machinesets.go
```diff
@@ -109,6 +120,75 @@ func CreateMachineSet(c client.Client, params MachineSetParams) (*machinev1.Mach
...
+	baseProviderSpec := baseMachineSetParams.ProviderSpec.DeepCopy()
+
+	output := []MachineSetParams{}
+	switch platform {
+	case configv1.AWSPlatformType:
+		// Using cheapest compute optimized instances that meet openshift minimum requirements (4 vCPU, 8GiB RAM)
+		alternativeInstanceTypes := []string{"c5.xlarge", "c5a.xlarge", "m5.xlarge"}
+		for _, instanceType := range alternativeInstanceTypes {
+			updatedProviderSpec, err := updateProviderSpecAWSInstanceType(baseProviderSpec, instanceType)
+			Expect(err).ToNot(HaveOccurred())
```
lobziik, line 135: Good catch form @Fedosin here: https://github.com/openshift/cluster-api-actuator-pkg/pull/255/files#r1026531625

+1 from me, lets avoid using ginkgo stuff here, since it might be called outside of `It`

pkg/framework/machinesets.go
RadekManak: Yes, it is used from the spot package. Added comment

pkg/framework/machinesets.go
RadekManak: I think you meant the current default m6i.xlarge as m6i.large only has two cores. I think using different instance types for the spot test will give lower interruption rate.

pkg/framework/machinesets.go
```diff
@@ -369,6 +449,106 @@ func WaitForMachineSetDelete(c runtimeclient.Client, machineSet *machinev1.Machi
...
+				}
+			}
+		}
+	case configv1.AzurePlatformType:
+		azureProviderStatus := machinev1.AzureMachineProviderStatus{}
+		if m.Status.ProviderStatus != nil {
+			err := json.Unmarshal(m.Status.ProviderStatus.Raw, &azureProviderStatus)
+			Expect(err).ToNot(HaveOccurred())
+			for _, condition := range azureProviderStatus.Conditions {
+				if condition.Type == string(machinev1.MachineCreated) && strings.Contains(condition.Message, "SkuNotAvailable") {
```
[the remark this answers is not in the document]
RadekManak: The error message is the most important here. There was that inverted Condition bug I found. It is fixed now, so I've added the test here.

pkg/framework/machinesets.go
```diff
@@ -369,6 +459,125 @@ func WaitForMachineSetDelete(c runtimeclient.Client, machineSet *machinev1.Machi
...
+	default:
+		return false, errTestForPlatformNotImplemented
+	}
+	return false, nil
+}
+
+// hasInsufficientCapacity return true if there is an insufficient spot capacity condition.
+func hasInsufficientCapacityCondition(conditions []metav1.Condition, platform configv1.PlatformType) (bool, error) {
+	for _, condition := range conditions {
+		if condition.Type == string(machinev1.MachineCreation) && condition.Status == metav1.ConditionFalse {
```
lobziik, line 567: > condition.Type == string(machinev1.MachineCreation)

I believe `machinev1.MachineCreated` also should be counted here.
https://github.com/openshift/api/blob/master/machine/v1beta1/types_provider.go#L135 - see definition.

lobziik (commented):
Mostly LGTM. Just one thing regarding conditions check.

pkg/framework/machinesets.go
RadekManak: You are right, broke this when refactoring.

RadekManak:
Rebased on top of master.

lobziik (approved):
/lgtm

lobziik:
spot tests passed on aws

/lgtm

lobziik:

These tests do not run on vsphere. Just underlying infra flake, no necessity to retest

PR #256, opened 2022-11-14
[OCPCLOUD-1732] Add test-e2e-periodic target in the makefile

Fedosin:
This PR adds a target to run e2e periodic tests only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,6 @@ all: check
 vendor:
 	$(DOCKER_CMD) ./hack/go-mod.sh
 
-
 .PHONY: check
 check: fmt vet #lint ## Check your code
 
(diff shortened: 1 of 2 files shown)


JoelSpeed:
For each test that you've marked as a periodic, can you please explain why you think this warrants being a periodic rather than running on presubmit?

I would like to be very deliberate about the tests that we mark as periodic, they should have good reasons behind them.

For example, it may be useful to make a test plan similar to the one for the CPMS which explains why each of these tests should be periodic and the behavioural subsets we cover in presubmits

lobziik:
Is this still actual after https://github.com/openshift/cluster-api-actuator-pkg/pull/258 was merged?

There I marked proxy test as periodic: https://github.com/openshift/cluster-api-actuator-pkg/pull/258/files#diff-ddf400001089ed146e8291e2e81b9fa305e9e371d81a389561ee6c61a11b8a5bR229

/hold

Makefile
```diff
@@ -74,6 +73,10 @@ build-e2e:
 test-e2e: ## Run openshift specific e2e test
 	hack/ci-integration.sh $(GINKGO_ARGS) --label-filter='!(periodic || spot-instances)' -p
 
+.PHONY: test-e2e-periodic
+test-e2e-periodic: ## Run openshift specific periodic e2e test
+	hack/ci-integration.sh $(GINKGO_ARGS) --label-filter='periodic && !spot-instances' -p
```
lobziik, line 78: think `--label-filter='periodic'` should be enough

lobziik (changes requested):
Think we don't need to explicitly exclude spot-tests

JoelSpeed:

The autoscaler test that's failing is unrelated to this change. I can see the proxy test is still being skipped and the new make target looks good to me, don't think we need this to pass

PR #258, opened 2022-11-15
OCPCLOUD-1729: Update ginkgo to v2, restructure test running 

lobziik:
this PR:

- Updates gingko to v2
- Marked tests with labels, and filter out proxy and spot tests.
- Introduces a special label for periodic.
- Marks autoscaler and serial tests with ginkgo serial label, now tests might be run from a single entrypoint.

Some tests are marked with Serial, according to the gingko doc it should be running one by one after the rest of the suite passed.
Namely autoscaler tests and serial disruptive MAPI operator tests will be run after the rest of the suite.


lobziik:
cc @JoelSpeed

pkg/operators/machine-api-operator.go
```diff
@@ -24,289 +24,298 @@ var (
 	maoManagedDeployment = "machine-api-controllers"
 )
 
-var _ = Describe("[Feature:Operators][Disruptive] Machine API operator deployment should", func() {
-	defer GinkgoRecover()
+var _ = Describe(
+	"[Serial][Feature:Operators][Disruptive] Machine API operator deployment should",
```
JoelSpeed, line 28: Should the prefixes be dropped from these?
```suggestion
	"Machine API operator deployment should",
```

pkg/operators/machine-api-operator.go
```diff
@@ -24,289 +24,298 @@ var (
...
+	Label(framework.LabelDisruptive, framework.LabelOperators, framework.LabelMachines),
+	Serial,
+	func() {
+		defer GinkgoRecover()
```
JoelSpeed, line 32: You don't actually need this, though I know it was already there

JoelSpeed (commented):
What's the motivation for keeping `[Feature:Machines]` in the description of all of the tests? Why did that not become a label?

JoelSpeed:
Looking at the test results, it seems the tests aren't running the serial tests at the end, have you tried setting it to parallel execution, I think it's running everything as serial right now?

Also not seeing any JUnit output

lobziik:
> Looking at the test results, it seems the tests aren't running the serial tests at the end, have you tried setting it to parallel execution, I think it's running everything as serial right now?

Yup, it is. I figured it out yesterday, and marked this PR as WIP. There is still a couple of things that need to be sorted.

lobziik:
> Also not seeing any JUnit output
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-actuator-pkg/258/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-azure-operator/1592546034800660480/artifacts/e2e-azure-operator/test/artifacts/e2e_report.xml

pkg/operators/machine-api-operator.go
lobziik: Removed

lobziik:
Updated this pr:
- removed `[Feature:Machines]` thingies
- Ensured that tests runs in parallel (I did run that manually on my machine against aws cluster, worked as I expected, got it passed)
- updated gatherer for work with new ginkgo spec report
- extended some timeouts a little bit (observed failure due to timeouts in IMDS and autoscaler tests, added two minutes to that)
- added 5m grace period to give room for gatherer and cleanup logic in AfterEach nodes (for cases if some test failed and other parallel tests were interrupted). The default was the 30s, I did set it for 5m.

pkg/framework/framework.go
```diff
@@ -48,10 +48,11 @@ const (
 )
 
 var (
-	WaitShort    = 1 * time.Minute
-	WaitMedium   = 3 * time.Minute
-	WaitLong     = 15 * time.Minute
-	WaitOverLong = 30 * time.Minute
+	WaitShort      = 1 * time.Minute
+	WaitMedium     = 3 * time.Minute
+	WaitOverMedium = 3 * time.Minute
```
JoelSpeed, line 53: This is the same as medium no?

JoelSpeed:
Test results look great, they almost seem to be running faster with this new approach as well 🤔

pkg/framework/framework.go
lobziik: that intended to be 5 mins, fixed

JoelSpeed:
/approve
Interested to see how the tests pass/fail on this run

elmiko:
this generally LGTM, i examined the azure-operator failures and i think it's related to slow machine creation during the tests. it seems like the test is failing while wait for machines to become nodes. perhaps we need to make the wait 30 minutes instead of 15?

lobziik:
>this generally LGTM, i examined the azure-operator failures and i think it's related to slow machine creation during the tests. it seems like the test is failing while waiting for machines to become nodes. perhaps we need to make the wait 30 minutes instead of 15?

It did hit global suite timeout. Due to slow machine creation, I believe.
This behaviour in ginkgo v2 changed, now 90m passed in `-timeout` flag is the limit for the entire suite.
Increased it in https://github.com/openshift/cluster-api-actuator-pkg/pull/258/commits/9bcb3f26d0c574d38127a23383708c731ca73414, I hope it should be enough.

JoelSpeed:
/lgtm

This last run, the suite passed in 1h40, so yep, timeout was causing an issue, but then the gather step failed afterwards. I'm happy with the changes, let's iterate if we find other issues

PR #260, opened 2022-12-01
OCPCLOUD-1743 Reduce amount of machines creation

RadekManak:
This PR reduces the amount of machine creation to the minimum that is required. 
- It moves the "scales up and down while respecting MaxNodesTotal [Slow][Serial]" test own context. Reducing MaxNodesTotal from 12 to 8. Also adds Eventualy for "Ensuring there are 6 machines in the cluster" at the beginning of the test. I hope that this will reduce the flakiness of this test.
- Infra test are grouped into 'When("machineset has x replicas")' contexts that set up the initial machine set replica count
    Tests with changes:
    1. "have ability to additively reconcile taints from machine to nodes" 2 -> 1 machine
    2. In "grow and decrease when scaling different machineSets simultaneously" reduces machineset2 to 2 replicas from 3
    3. "reject invalid machinesets" 2 -> 0 replicas by not using shared setup that created unused machineset as it is only webhook test.
- Reduces the scale from zero replicas from 3 to 2
- "cleanup deletion information after scale down" Reduces workload on machinesets from 3 to 2
- Reduces the spot test machineset replicas from 3 to 1
- Reduces "should return an error when removing required fields from the MachineSet providerSpec" replicas form 1 to 0
- In "should remediate unhealthy nodes" Uses maxUnhealthy instead of maxUnhealthy-1. This allows the test to use just 1 healthy and 1 unhealthy machine. + 1 replacement machine. Instead of 1 unhealthy, 3 healthy machines.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -214,6 +214,8 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func() 
 			}, framework.WaitMedium, pollingInterval).Should(BeTrue())
 		})
 
+		// Machines required for test: 2
+		// Reason: This tests checks that autoscaler is able to scale from zero. It requires 2 machines to ensure it scales to the correct number of nodes based on the workload size.
 		It("It scales from/to zero", func() {
 			// Only run in platforms which support autoscaling from/to zero.
 			clusterInfra, err := framework.GetInfrastructure(client)
@@ -238,7 +240,7 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func() 
 
 			framework.WaitForMachineSet(client, machineSet.GetName())
 
-			expectedReplicas := int32(3)
+			expectedReplicas := int32(2)
 			By(fmt.Sprintf("Creating a MachineAutoscaler backed by MachineSet %s/%s - min:%v, max:%v",
 				machineSet.GetNamespace(), machineSet.GetName(), 0, expectedReplicas))
 			asr := machineAutoscalerResource(machineSet, 0, expectedReplicas)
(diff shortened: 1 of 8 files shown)


pkg/providers/aws.go
```diff
@@ -121,18 +121,24 @@ var _ = Describe("MetadataServiceOptions", framework.LabelCloudProviderSpecific,
 		})
 	}
 
+	// Machines required for test: 0
```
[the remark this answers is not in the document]
JoelSpeed: NVM I see what's happening

pkg/autoscaler/autoscaler.go
```diff
@@ -279,6 +281,9 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 			}, framework.WaitLong, pollingInterval).Should(BeTrue())
 		})
 
+		// Machines required for test: 4
+		// Reason: Needs to scale down to minReplicas = 1. Scales 1 -> 2 -> 1. Uses 2 machineSets => 4 machines.
+		// This test copies reproduction steps for a a bug. I have not found a reason why it uses two machineSets. Should be moved to periodic.
```
elmiko, line 286: we could probably reduce this to a single machineset. it seems like the main thing we are testing here is that the deletion taint is added to the machine properly.

elmiko (commented):
this looks really great to me @RadekManak , nice work!

imo, we could reduce the deletion cleanup test down to a single machineset.

pkg/autoscaler/autoscaler.go
RadekManak: Wasn't sure if there was any special interaction with two machine sets. If you feel it will still test what it should, then I am happy with changing it too. Changed it to use single machineset.

RadekManak:
- Changed the "cleanup deletion information after scale down [Slow]" test to use only one machineSet instead of two.
- Fixed machine update not being retried on conflict in one test.

pkg/autoscaler/autoscaler.go
```diff
@@ -279,145 +281,218 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+						return true, nil
+					}
+					if _, exists := m.ObjectMeta.Annotations[machineDeleteAnnotationKey]; exists {
 						return false, nil
 					}
-				}
-				return true, nil
-			}, framework.WaitLong, pollingInterval).Should(BeTrue())
+					return true, nil
+				}, framework.WaitMedium, pollingInterval).Should(BeTrue())
```
elmiko, line 354: just curious why this was switched to a `WaitMedium` instead of `WaitLong`

elmiko (commented):
generally looks good to me, just a question

pkg/autoscaler/autoscaler.go
RadekManak: It was not. This is just confusing diff. There are three eventually blocks next to each other. The first uses WaitLong and the other two use WaitMedium.

elmiko: ack, thanks!

elmiko (commented):
/lgtm

PR #261, opened 2022-12-02
OCPCLOUD-1827: Fix lint issues reported by golangci-lint

Fedosin:
This PR adds golangci-lint support and enables additional linters for the component.


pkg/framework/gatherer/oc-cli.go
```diff
@@ -68,31 +70,38 @@ func (oc *CLI) outputs(stdOutBuff, stdErrBuff *bytes.Buffer) (string, string, er
...
 	}
+
 	err = cmd.Wait()
 
 	stdOutBytes := stdOutBuff.Bytes()
 	stdErrBytes := stdErrBuff.Bytes()
 	stdOut := strings.TrimSpace(string(stdOutBytes))
 	stdErr := strings.TrimSpace(string(stdErrBytes))
 
-	switch err.(type) {
```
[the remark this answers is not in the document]
Fedosin: The message was
```
pkg/framework/gatherer/oc-cli.go:78:9: type switch on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
```

pkg/framework/framework.go
```diff
@@ -202,45 +208,48 @@ func NewCLI() (*gatherer.CLI, error) {
...
+func getCliOutputFilesPath() string {
 	if isOpenshiftCI() {
-		return filepath.Join(os.Getenv(ARTIFACT_DIR), "machine-api-e2e-suite"), nil
+		return filepath.Join(os.Getenv(artifactDir), "machine-api-e2e-suite")
 	}
 
 	cwd, err := os.Getwd()
 	if err != nil {
-		return "", nil
+		return ""
```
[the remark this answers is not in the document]
Fedosin: It's a good question... I would prefer to return an error here.

pkg/framework/framework.go
```diff
@@ -227,20 +237,22 @@ func isOpenshiftCI() bool {
...
 func getCliOutputFilesPath() (string, error) {
 	if isOpenshiftCI() {
-		return filepath.Join(os.Getenv(ARTIFACT_DIR), "machine-api-e2e-suite"), nil
+		return filepath.Join(os.Getenv(artifactDir), "machine-api-e2e-suite"), nil
 	}
 
 	cwd, err := os.Getwd()
 	if err != nil {
-		return "", nil
+		return "", err
```
JoelSpeed, line 245: Does this not cause a linter error, I thought all errors must be wrapped?

pkg/infra/infra.go
```diff
@@ -167,13 +172,12 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
...
 		specReport := CurrentSpecReport()
-		if specReport.Failed() == true {
+		if specReport.Failed() {
 			Expect(gatherer.WithSpecReport(specReport).GatherAll()).To(Succeed())
 		}
 
 		By("Deleting the new MachineSet")
-		err := client.Delete(context.Background(), machineSet)
-		Expect(err).ToNot(HaveOccurred())
+		Expect(client.Delete(context.Background(), machineSet)).ToNot(HaveOccurred())
```
JoelSpeed, line 180: Can we change `ToNot(HaveOccurred())` to `To(Succeed())` in these cases please?

pkg/infra/spot.go
```diff
@@ -268,9 +271,10 @@ func setSpotOnProviderSpec(platform configv1.PlatformType, params framework.Mach
 func setSpotOnAWSProviderSpec(params framework.MachineSetParams, maxPrice string) error {
 	spec := machinev1.AWSMachineProviderConfig{}
 
-	err := json.Unmarshal(params.ProviderSpec.Value.Raw, &spec)
-	if err != nil {
-		return fmt.Errorf("error unmarshalling providerspec: %v", err)
+	var err error
```
JoelSpeed, line 274: I don't think this is needed here is it? Can we move the decl closer to where it's needed

pkg/operators/cluster-autoscaler-operator.go
```diff
@@ -54,8 +54,7 @@ var _ = Describe("Cluster autoscaler operator should", framework.LabelOperators,
 			},
 		}
 
-		err := client.Create(context.TODO(), invalidCA)
-		Expect(err).To(HaveOccurred())
+		Expect(client.Create(context.TODO(), invalidCA)).To(HaveOccurred())
```
JoelSpeed, line 57: This should be `ToNot(Succeed())` IMO

pkg/providers/aws.go
```diff
@@ -62,11 +62,12 @@ var _ = Describe("MetadataServiceOptions", framework.LabelCloudProviderSpecific,
 	})
 
 	createMachineSet := func(metadataAuth string) (*machinev1.MachineSet, error) {
+		var err error
```
JoelSpeed, line 65: Is this needed? Can we move it to where it is needed please

JoelSpeed (commented):
I'd like if possible to clean up some of the assertions.

When you have a function, it should be `Expect(func()).To(Succeed())` or `Expect(func()).ToNot(Succeed()` rather than `To(HaveOccurred())` or `ToNot(HaveOccurred()`, the occurred ones are more for errors in variables, when you read the code, it makes sense, expect and error to have occurred, expect a function to have occurred doesn't make as much sense

pkg/framework/framework.go
Fedosin: It's because I temporarily disabled `wrapcheck` linter. There are too many errors to wrap, so it could be done in a separate PR.

JoelSpeed: You could wrap this one as good practice ahead of the new linter being set up

pkg/autoscaler/autoscaler.go
```diff
@@ -168,7 +169,7 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 
 		BeforeEach(func() {
 			gatherer, err = framework.NewGatherer()
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).To(Succeed())
```
JoelSpeed, line 172: This was correct before 😅 `Expect(err).ToNot(HaveOccurred())` is correct, it reads like a sentence

elmiko (commented):
nice cleanup
/lgtm

PR #262, opened 2022-12-05
[OCPCLOUD-1728] Listing and document tests within the MAPI suite

lobziik:


docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+The test is expecting machineset to be scaled up to 3 machines and accept the payload. Then, created Job is removing and the machineset is expected to be scaled down back to 0 replicas. 
+
+Test considered failed if scale up or down did not happen within 15 min timeout.
+
+#### Autoscaler should use a ClusterAutoscaler that has 100 maximum total nodes count cleanup deletion information after scale down [Slow]
+* Parallel: No
+* ExecTime: ~13m on AWS
+* Recommendation:
+  * This test is a bit cryptic at first glance and requires at least some in-code explanation about the meaning of checks for taints and annotations presence. Having some documentation references will be beneficial here.
+  * amount of desired machines might be reduced from 3 to 2 for each machineset
```
elmiko, line 400: i also wonder if this functionality is being exercised by the other tests

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+Test considered failed if scale up or down did not happen, or after scaling down there are taints/annotations leftovers within a 3 minutes timeout.
+
+#### Autoscaler should use a ClusterAutoscaler that has 12 maximum total nodes count and balance similar nodes enabled scales up and down while respecting MaxNodesTotal [Slow][Serial]
+* Parallel: No
+* ExecTime: ~13m on AWS
+* Recommendation:
+  * Since this test does not leverage or checking `balance similar nodes` feature, this might be extracted to separate group with reducing `MaxNodesTotal` number to ~8. This should speed up this test a bit and reduce resource consumption.
+  * This test requires revise and extension of checks. Workload destiny need to be checked too. As i understand this test at the moment of writing, part of the workload should not be scheduled due to `MaxNodesTotal` limit.
+  * Gather and report a number of nodes when the test starts
+  * Might be merged with from/to zero scaling test ([Autoscaler should use a ClusterAutoscaler that has 100 maximum total nodes count It scales from/to zero](#autoscaler-should-use-a-clusterautoscaler-that-has-100-maximum-total-nodes-count-it-scales-fromto-zero)).
```
elmiko, line 418: +1

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+with `MaxNodesTotal` + 1 desired pods in it. Then waits cluster to be scaled up till `MaxNodesTotal` nodes number and checks that this constraint was not violated.
+
+Then workload deletes and test waits for the machineset to be scaled back to 1 replica.
+
+#### Autoscaler should use a ClusterAutoscaler that has 12 maximum total nodes count and balance similar nodes enabled places nodes evenly across node groups [Slow]
+* Parallel: No
+* ExecTime: ~10m on AWS
+* Recommendation:
+  * Check kubernetes-autoscaler deployment for the `balance-similar-node-groups` flag presence first
+  * Currently max machineset size for machine autoscaler resource is 3 within this test. I suggest to increase it to 4 or 6 for being more sure that `balance-similar-node-groups` does its thing and that the test pass is not a coincidence.
```
elmiko, line 430: +1, also as discussed in standup, we might want to create another machineset, with zero replicas, that does not match the job to ensure that we aren't placing nodes in the empty machineset.

elmiko (commented):
this doc generally looks great to me @lobziik , thanks!

i reviewed the autoscaler parts a little closer and tend to agree with your recommendations. i also added a comment from our discussion.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+In `BeforeEach` machineset creation and wait happen (this machineset has 2 desired replicas configured, so 2 machines will be created for each test here).
+In `AfterEach` created machineset is being deleted.
+
+#### Managed cluster should drain node before removing machine resource
+* Parallel: Yes
+* ExecTime: >12m on AWS
+* Recommendations:
+  * Update resource creation helper functions, and add parameters to specify workload names, replicas count, selectors, etc. Now it's almost impossible to say what is going on without looking into these functions.
+  * Specify selectors more strictly for the created workloads, now it's theoretically possible that created workload will be scheduled somewhere else and we could get false positive results.
+  * Add more meaningful comments inside the test code, now it's not easy to figure out what is going on there.
```
JoelSpeed, line 247: Do we want to have these as step delineation using `By` clauses?

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+  * Fix logging here, it is quite noisy, and do flood reports.
```
JoelSpeed, line 248: +1 this has been on my radar for a while, for some reason it prints out every pod right? We don't need that

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+
+This test checks nodes are properly draining during machine deletion and this process also respects PDBs (pod disruption budgets).
+During this test - a replication controller creating with 20 desired replicas, then PDB creating for this workload.
+After the replication controller is ready, machineset marking for deletion. Along with the deletion process test checking that pods amount on the node 
+is going down to 1 and RC has at most one non-ready replica.
+
+Test considered failed in case if more than 1 replica within RC detected as not ready during machineset deletion. That would mean that PDB constraints were violated.
```
JoelSpeed, line 255: Where do the pods go to when the machineset is removed?  Do we really need to run 20 pods?

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+
+This test takes machines from the machineset which was created in `BeforeEach` block and marks them for deletion,
+then waits for machines to be deleted and recreated by the machineset. Takes quite a while, because 2 machines should be deleted and recreated then.
+
+#### Managed cluster should grow and decrease when scaling different machineSets simultaneously
+* Parallel: Yes
+* ExecTime: >10m on AWS
+* Recommendations:
+  * Reduce created machines number to 1 for this test, currently there are 2
+  * Remove this test, similiar one included into origin suite. Aside of that we running this and other tests which creates and scales MSes in parallel, so we are excercising this functionality
```
JoelSpeed, line 272: I think this makes sense to move to a periodic, I think that's been discussed in Mikhails periodic recommendations

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+  * Move this to another webhook tests
+
+This test just tries to create machineset with empty provider config, but due to machineset is creating in `BeforeEach` block, it takes a while.
+We are creating and waiting for 2 machines here for nothing...
+
+#### Managed cluster should have ability to additively reconcile taints from machine to nodes
+* Parallel: Yes
+* ExecTime: >6m on AWS
+* Recommendation:
+  * Add check also for labels, or create new one which would exercise labels and annotations as well
```
JoelSpeed, line 292: We only reconcile the labels and annotations when the node is first created, via nodelink, so I don't think this makes sense does it?

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+5) Removes the pre-drain hook
+6) Checks that the workload pod was evicted, but the machine is still present
+7) Removes the pre-delete hook
+8) Checks the machine is deleted
+
+## MachineHealthCheck
+These tests check MachineHealthCheck functionality. Tests listed below share common `BeforeEach` and `AfterEach` blocks.
+Within `BeforeEach` machineset with 4 desired replicas creating. In `AfterEach` resources created during a test are attempted to be cleaned up.
+
+Recommendation: Refactor these tests to run them in order for and use the same machinesets. Given that we are setting up conditions on machines, we could perform
```
JoelSpeed, line 357: I like this idea, lets definitely do this!

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+* Parallel: No
+* ExecTime: ~13m on AWS
+* Recommendation:
+  * This test is a bit cryptic at first glance and requires at least some in-code explanation about the meaning of checks for taints and annotations presence. Having some documentation references will be beneficial here.
+  * amount of desired machines might be reduced from 3 to 2 for each machineset
+
+Within this test machineset with 1 desired replica and with a specific set of labels is created.
+After, a Job with 6 desired replicas is attempting to be scheduled on nodes with labels configured for previously created machinesets.
+
+The test is expecting both machinesets to be scaled up to 3 machines and accept the payload. Then, created Job is removing and the machineset is expected to be scaled down back to 1 replica.
```
JoelSpeed, line 405: This seems like it has a big overlap with the previous test right? Do we want to consolidate the two of them/maybe remove the first? Is the first covering something that isn't covered in this test?

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+
+After machinesets will be scaled back to 1 replica test checks if no cluster-autoscaler specific annotations are presented on machines and no autoscaler specific taints are presented on respective nodes from these machinesets.
+
+Test considered failed if scale up or down did not happen, or after scaling down there are taints/annotations leftovers within a 3 minutes timeout.
+
+#### Autoscaler should use a ClusterAutoscaler that has 12 maximum total nodes count and balance similar nodes enabled scales up and down while respecting MaxNodesTotal [Slow][Serial]
+* Parallel: No
+* ExecTime: ~13m on AWS
+* Recommendation:
+  * Since this test does not leverage or checking `balance similar nodes` feature, this might be extracted to separate group with reducing `MaxNodesTotal` number to ~8. This should speed up this test a bit and reduce resource consumption.
```
JoelSpeed, line 415: Reducing the total here seems reasonable, I wonder if 8 was chosen for some specific reason before?

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+# Tests
+
+## Operators
+
+### Machine API
+
+#### Machine API operator deployment should be available
+* Parallel: No (sits in disruptive suite)
+* ExecTime: ~5s on AWS
+* Recommendation: Might be extracted and run in parallel with the other part of the suite.
```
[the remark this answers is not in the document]
lobziik: Sounds reasonable to me.

> This seems like a pre-requisite for a bunch of our other tests right?

For machine API operator-related tests, yes. Having it as a prerequisite makes sense to me.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+* ExecTime: ~2s on AWS
+
+Checks cluster operator status to be available, not progressing and not degraded.
+
+#### When cluster-wide proxy is configured, Machine API cluster operator should  create machines when configured behind a proxy
+* Parallel: No
+* ExecTime: ~30m on AWS
+* Recommendation:
+  - ~make this test periodic (done already)~
+  - extend this test - add check for the CCCMO also
```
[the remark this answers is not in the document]
lobziik: I don't think so, since cluster autoscaler deals with machine resource through the kubernetes API.

Machine controllers and CCMs interact with cloud APIs, and setting up proxy properly is crucial here.
IIRC we have had bugs for CCMs and proxy issues.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+### Cluster Machine Approver
+
+Recommendation: This suite needs to be extended with some sort of disruptive suite (change/delete CMA deployment, ensure that it's preserved).
+Another variant is to remove this test at all since a cluster will not finish installation if the operator will be in the degraded state.
+
+#### Cluster Machine Approver deployment should be available
+* Parallel: Yes
+* ExecTime: ~2s on AWS
+
+This test checks status of the `openshift-cluster-machine-approver/machine-approver` deployment.
```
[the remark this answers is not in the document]
lobziik: Yup, might be. 

This might be merged with another CMA test, which checks the deployment. Aside from that two, we do not have any specific CMA tests

docs/tests.md
lobziik: 12 you mean? I don't know. Looking into tests I didn't figure out any specific reason. @elmiko maybe you know?

elmiko: we could certainly reduce this number, i think i chose 12 because it was inspired by the original values we were using to grow the machine sets.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+This test query `openshift-machine-api/machine-api-controllers` deployment, saves manifest and scales deployment to zero replicas.
+After, tests wait for the deployment to be restored and available (available replicas > 0).
+This test checks operator behaviour to manage and maintain MAPI components deployment state.
+
+#### Machine API operator deployment should reconcile mutating webhook configuration
+#### Machine API operator deployment should reconcile validating webhook configuration
+* Parallel: No (each)
+* ExecTime: ~5s on AWS (each)
+* Recommendation:
+  - Extract this test, make it parallel
```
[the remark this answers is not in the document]
lobziik: I could imagine some nested ordered container in the webhook suite. But yeah, this will require some experimentation.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+    - this two tests might be generalized in a way and merged into one, the only difference is a resource type
+
+This test checks that webhook configurations are presented in cluster then delete it. Waits for configurations to be restored and 
+being equal to what is encoded within the vendored MAPO repo.
+
+This test checks operator behaviour to manage and maintain MAPI webhooks state.
+
+#### Machine API operator deployment should maintains spec after mutating webhook configuration change and preserve caBundle
+#### Machine API operator deployment should maintains spec after validating webhook configuration change and preserve caBundle
+* Parallel: No (each)
```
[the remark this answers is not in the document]
lobziik: This tests delete webhook. `reconcile webhook configuration` doesn't touch anything.

docs/tests.md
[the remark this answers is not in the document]
lobziik: Or yes, pre-step. Will be convenient if we will decide to add another CMA-related tests.

docs/tests.md
```diff
@@ -0,0 +1,439 @@
...
+
+### Cluster Autoscaler
+
+Recommendation: This suite need to be extended with some sort of disruptive suite (change/delete autoscaler deployment, ensure that it's preserved by the operator)
+
+#### Cluster autoscaler operator should reject invalid ClusterAutoscaler resources early via webhook
+* Parallel: Yes
+* ExecTime: ~5s on AWS
+* Recommendation:
+  - Remove from this suite, move this test to the cluster autoscaler repo
```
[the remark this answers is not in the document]
lobziik: For this particular webhook - no. I'm thinking about adding a test which specifically checks webhooks configuration for all our components (MAPI, CAO, opesnshift specific CAPI perhaps)

docs/tests.md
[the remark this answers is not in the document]
JoelSpeed: Could they be combined? Would that make things faster?

docs/tests.md
elmiko: looking at @RadekManak PR, i remember why it was 12. 6 for the original cluster + 6 for the new workload we created.

docs/tests.md
```diff
@@ -243,13 +256,15 @@ In `AfterEach` created machineset is being deleted.
...
   * Fix logging here, it is quite noisy, and do flood reports.
+  * Reduce amount of pods created
+  * Check where pods was actually scheduled
 
 This test checks nodes are properly draining during machine deletion and this process also respects PDBs (pod disruption budgets).
-During this test - a replication controller creating with 20 desired replicas, then PDB creating for this workload.
-After the replication controller is ready, machineset marking for deletion. Along with the deletion process test checking that pods amount on the node 
+During this test - machineset with two desired replicas and a replication controller with 20 desired replicas are creating, then PDB creating for this workload.
+Machines from the machinesets marking with labels to be sure that pods created by the replication controller will be scheduled there.
+After the replication controller is ready, one of the machines from the machineset marking for deletion. Along with the deletion process test checking that pods amount on the node
```
JoelSpeed, line 269:
```suggestion
After the replication controller is ready, one of the machines from the machineset marking for deletion. Along with the deletion process test checking that number of pods on the node
```

docs/tests.md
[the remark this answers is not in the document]
lobziik: I don't think that it is worth to bother with optimizing this. This tests doesn't take much time.

docs/tests.md
lobziik: I thought about some human-readable description within the test itself, just as a comment.
This bullet point might be omitted, if we will be happy enough after refactoring describing in the first point.

docs/tests.md
lobziik: First one testing scaling from/to zero. This one checks that machine which stays after scaling down is not polluted with unwanted labels/annotations.

But it might be combined with `MaxNodesTotal` test.

elmiko: > But it might be combined with MaxNodesTotal test.

+1

JoelSpeed:
/approve

I'm happy with this as is right now, I expect it will evolve over time

Can we make sure that the recommendations turn into Jira cards in a new epic so that we can make sure action them going forward

docs/tests.md
```diff
@@ -0,0 +1,464 @@
...
+#### Autoscaler should use a ClusterAutoscaler that has 100 maximum total nodes count It scales from/to zero
+* Parallel: No
+* ExecTime: >7m on AWS
+* Recommendation:
+  * amount of desired machines might be reduced from 3 to 2
+
+Within this test machineset with the initial desired 0 replicas and with a specific set of labels is created.
+After, a Job with 3 desired replicas is attempting to be scheduled on nodes with labels configured for previously created machineset.
+
+The test is expecting machineset to be scaled up to 3 machines and accept the payload. Then, created Job is removing and the machineset is expected to be scaled down back to 0 replicas.
```
elmiko, line 415: minor nit
```suggestion
The test is expecting machineset to be scaled up to 3 machines and accept the payload. Then, created Job is removed and the machineset is expected to be scaled down back to 0 replicas. 
```

elmiko (commented):
content-wise i think things look good here. i've made some suggestions that i think help the readability, and a couple questions that i had while reading.

docs/tests.md
```diff
@@ -0,0 +1,464 @@
...
+In `BeforeEach` machineset creation and wait happen (this machineset has 2 desired replicas configured, so 2 machines will be created for each test here).
+In `AfterEach` created machineset is being deleted.
+
+#### Managed cluster should drain node before removing machine resource
+* Parallel: Yes
+* ExecTime: >12m on AWS
+* Recommendations:
+  * Update resource creation helper functions, and add parameters to specify workload names, replicas count, selectors, etc. Now it's almost impossible to say what is going on without looking into these functions.
+  * Add more meaningful comments inside the test code, now it's not easy to figure out what is going on there.
+  * Fix logging here, it is quite noisy, and do flood reports.
```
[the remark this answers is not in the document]
lobziik: It produces a lot of rubbish log messages which are mostly unreadable. I will reword this.

docs/tests.md
```diff
@@ -0,0 +1,464 @@
...
+  * Reduce amount of pods created
+  * Check where pods was actually scheduled
+
+This test checks nodes are properly draining during machine deletion and this process also respects PDBs (pod disruption budgets).
+During this test - machineset with two desired replicas and a replication controller with 20 desired replicas are creating, then PDB creating for this workload.
+Machines from the machinesets marking with labels to be sure that pods created by the replication controller will be scheduled there.
+After the replication controller is ready, one of the machines from the machineset marking for deletion. Along with the deletion process test checking that number of pods on the node
```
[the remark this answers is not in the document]
lobziik: I reworded this a bit, hope it would be better now.

docs/tests.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Cluster Autoscaler
+
+Recommendation: This suite need to be extended with some sort of disruptive suite (change/delete autoscaler deployment, ensure that it's preserved by the operator)
+
+#### Cluster autoscaler operator should reject invalid ClusterAutoscaler resources early via webhook
+* Parallel: Yes
+* ExecTime: ~5s on AWS
+* Recommendation:
+  - Remove from this suite, move this test to the cluster autoscaler repo
```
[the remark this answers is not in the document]
lobziik: Yeah, you are right. When I initially dug around this, I didn't read through CAO code properly. Actual validation logic is covered with unittests there, this two are definitely useful for checking that webhooks are set up properly and respective controllers are responding.

I rewrote this section, PTAL.

docs/tests.md
```diff
@@ -0,0 +1,464 @@
...
+  - Or review and extend, but better remove
+
+This test just checks that creation of ClusterAutoscaler resource with name other than 'default' is rejected
+
+#### Cluster autoscaler operator should reject invalid MachineAutoscaler resources early via webhook
+* Parallel: Yes
+* ExecTime: ~5s on AWS
+* Recommendation:
+  - Remove, and implement this test to the cluster autoscaler repo
```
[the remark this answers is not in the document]
lobziik: https://github.com/openshift/cluster-api-actuator-pkg/pull/262#discussion_r1047114264

elmiko (commented):
thanks for the updates @lobziik 

/lgtm

PR #265, opened 2023-01-10
[OCPCLOUD-1706] Migrate test utils from CPMSO repository

RadekManak:
This PR moves test utilities and resource builders from CPMSO repository into this repository. This will allow us to reuse them in other test suites.

Tooling for unit test is created as this repo currently only has e2e tests. This will need to be enabled in openshift/release after this PR merges.
The MachineInfoBuilder is kept in the CPMSO repo. MachineInfo's are used only in CPMSO. Moving the builder over would create cyclic dependency.


pkg/testutils/resourcebuilder/consts.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package resourcebuilder
+
+const (
+	// OpenshiftMachineAPINamespaceName is the name of the OpenShift
+	// Machine API namespace.
+	OpenshiftMachineAPINamespaceName = "openshift-machine-api"
+
+	testClusterIDValue = "cpms-cluster-test-id"
```
JoelSpeed, line 24: We probably don't want this to be `cpms` anymore, where else is it used?

JoelSpeed (commented):
What is the effect on CPMS of importing these changes? Do we need to make a separate go module to confine the imports to just the testutils module or does it work well as is?

Do we have appropriate linting rules on the testutils?

Do we want to keep the resource builders as a flat structure, or group them by API group/versions as you'd expect for other helpers?

pkg/testutils/resourcebuilder/consts.go
RadekManak: As machine label, machine selector.  I have dropped the cpms prefix.

RadekManak:
> What is the effect on CPMS of importing these changes? Do we need to make a separate go module to confine the imports to just the testutils module or does it work well as is?
I was able to import from CPMSO without issues. 

> Do we have appropriate linting rules on the testutils?
The linter is running on the files. (with slightly different settings)

> Do we want to keep the resource builders as a flat structure, or group them by API group/versions as you'd expect for other helpers?
Probably good idea to structure it now that it will be imported from other repositories. I have restructured it. Let me know if it looks good to you.

damdo:
@RadekManak this PR is merged now FYI, if you want to pull in also those builders.
https://github.com/openshift/cluster-api-actuator-pkg/pull/265#issuecomment-1377460596

RadekManak:
Yes, thanks already done. :)

pkg/testutils/resourcebuilder/machine/v1beta1/openshift_machine_v1beta1_template.go
```diff
@@ -0,0 +1,93 @@
...
+package v1beta1
+
+import (
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-api-actuator-pkg/pkg/testutils/resourcebuilder"
+)
+
+// OpenShiftMachineV1Beta1Template creates a new OpenShift machine template builder.
+func OpenShiftMachineV1Beta1Template() OpenShiftMachineV1Beta1TemplateBuilder {
```
JoelSpeed, line 26: This should be in `machine/v1`

JoelSpeed (commented):
Couple of nits, otherwise LGTM

pkg/testutils/resourcebuilder/config/v1/infrastructure.go
```diff
@@ -0,0 +1,170 @@
...
+*/
+
+package v1
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Infrastructure creates a new awsplacementgroup builder.
```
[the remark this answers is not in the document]
RadekManak: Not sure, It was like this in the CPMSO repo as well. Fixed.

Makefile
```diff
@@ -69,6 +73,10 @@ vet: ## Apply go vet to all go files
 build-e2e:
 	$(DOCKER_CMD) go test -c -o "$(BUILD_DEST)" github.com/openshift/cluster-api-actuator-pkg/pkg/
 
+.PHONY: unit
+unit: ## Run unit tests
```
elmiko, line 77: do we need to add a prow configuration to exercise this new target?

elmiko (commented):
looks mostly good to me, just a question

Makefile
RadekManak: Yes, this needs to be set up. The tests do pass locally right now.

Makefile
```diff
@@ -7,7 +7,11 @@ export GOFLAGS
 GOPROXY ?=
 export GOPROXY
 
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.25
+
```
damdo, line 12: I saw Daniel is doing similar work but moving things to 1.26.
Do we want to do it here already or do we prefer to do so in a followup PR?

go.mod
```diff
@@ -3,20 +3,22 @@ module github.com/openshift/cluster-api-actuator-pkg
...
 	github.com/onsi/gomega v1.24.1
-	github.com/openshift/api v0.0.0-20220930192943-199dc0598fd1
+	github.com/openshift/api v0.0.0-20221004120407-c46852673d03
 	github.com/openshift/cluster-autoscaler-operator v0.0.1-0.20220930224803-cdc49524de94
 	github.com/openshift/library-go v0.0.0-20220930165906-a18683c20aee
 	github.com/openshift/machine-api-operator v0.2.1-0.20220922080915-f7f326323f37
 	k8s.io/api v0.25.1
 	k8s.io/apimachinery v0.25.1
-	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/client-go v0.25.1
```
damdo, line 17: Same for these

pkg/testutils/cleanup.go
```diff
@@ -0,0 +1,150 @@
+/*
```
damdo, line 1: Similarly to @JoelSpeed I also do wonder if it would be worth confining `testutils` in its own go module, to confine it from the rest of the e2e suites and avoid importing all of it everywhere.

damdo (commented):
Generally LGTM. Thanks for putting this together.
Left mainly a comment regarding importing.

Makefile
RadekManak: Testutils is now based on 1.26. E2E suite will require machine-api-operator to be updated first.

pkg/testutils/cleanup.go
RadekManak: I have made it into its own module. I've had some issues getting unit tests and liters to run on both modules, but I got it to work.

Makefile
RadekManak: I have https://github.com/openshift/release/commit/09ee970b2de485ddfa1dda98d9628df5ab9944db prepared, but we need to merge this first. There is no unit target on master, so it won't pass until this PR adds it.

testutils/resourcebuilder/meta/v1/conditions.go
```diff
@@ -0,0 +1,70 @@
...
+func (c StatusConditionBuilder) WithReason(reason string) StatusConditionBuilder {
+	c.reason = reason
+	return c
+}
+
+// WithMessage sets the message for the status condition builder.
+func (c StatusConditionBuilder) WithMessage(message string) StatusConditionBuilder {
+	c.message = message
+	return c
+}
```
JoelSpeed, line 70: We should rename this to just `Condition` from `StatusCondition` now that it's in its own folder

JoelSpeed (commented):
I think I would advise considering the `testutils` module as if it were a completely independent repository. Set up its own makefile and hack scripts for running the tests, and then link to those targets from the top level using `make -C testutils`, it will simplify a lot of the bash/makefile logic and keep them better isolated

Makefile
```diff
@@ -7,7 +7,11 @@ export GOFLAGS
 GOPROXY ?=
 export GOPROXY
 
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
+ENVTEST_K8S_VERSION = 1.26
+
 PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+ENVTEST = go run ${PROJECT_DIR}/vendor/sigs.k8s.io/controller-runtime/tools/setup-envtest
```
JoelSpeed, line 14: Are these new additions used now we've moved things to a separate makefile?

testutils/Makefile
```diff
@@ -0,0 +1,74 @@
...
+lint: ## Go lint your code
+	( GOLANGCI_LINT_CACHE=$(PROJECT_DIR)/.cache $(GOLANGCI_LINT) run --config ../.golangci.yaml --timeout 10m )
+
+.PHONY: fmt
+fmt: ## Go fmt your code
+	$(DOCKER_CMD) hack/go-fmt.sh .
+
+.PHONY: goimports
+goimports: ## run goimports
+	$(DOCKER_CMD) hack/goimports.sh .
```
JoelSpeed, line 66: Are these not covered by GolangCI Lint? My understanding was they are both run as part of its suite

go.mod
```diff
@@ -5,18 +5,19 @@ go 1.19
...
 	k8s.io/klog v1.0.0
 	k8s.io/utils v0.0.0-20220823124924-e9cbc92d1a73
 	sigs.k8s.io/controller-runtime v0.13.0
+	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20221222054525-3c4deba74e5c
```
JoelSpeed, line 20: Is this correct? I didn't think setup-entest was needed at this level anymore?

testutils/Makefile
RadekManak: You are correct. I have kept them because, they were being executed by CI. There are currently two targets that run on this repo. (excluding E2Es). `check` that on main branch runs fmt and vet, and goimports. The golanCI lint has been added recently and is not being executed by CI. If you want, we can remove these and only keep golangci lint and unit.

Makefile
RadekManak: Forgot to remove this. Thanks

go.mod
RadekManak: Forgot to remove this. Thanks

testutils/Makefile
JoelSpeed: If that's the case, then yep, lets keep them for now but I think we should migrate to golangci-lint as the single source moving forward, that can be a follow up PR though

JoelSpeed:
/approve

/assign @damdo 

Dam for LGTM on this as we talked about it during standup

damdo (commented):
Thanks for putting this together Radek.

/lgtm

PR #267, opened 2023-02-10
[OCPCLOUD-1715]: Add exception annotations to mhc tests

odvarkadaniel:
This PR annotates all the `Expect` block with exceptions so we can easily trace bugs in the test suite.

### DIFF
--- a/pkg/machinehealthcheck/machinehealthcheck.go
+++ b/pkg/machinehealthcheck/machinehealthcheck.go
@@ -41,16 +41,16 @@ var _ = Describe("MachineHealthCheck", framework.LabelMachineHealthChecks, func(
 		var err error
 
 		gatherer, err = framework.NewGatherer()
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "failed to create a new StateGatherer")
 
 		client, err = framework.LoadClient()
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "failed to create a new controller-runtime client")
 
 		machineSetParams := framework.BuildMachineSetParams(client, expectedReplicas)
 
 		By("Creating a new MachineSet")
 		machineSet, err = framework.CreateMachineSet(client, machineSetParams)
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "failed to create a new machineSet resource")
 
 		framework.WaitForMachineSet(client, machineSet.GetName())
 	})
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
I don't think we need to use so many format strings in these messages, the errors will be printed by the matcher anyway so including the error in the string in most of these cases is duplicating information in the logs, are we sure we want that? CC @elmiko

elmiko:
good question @JoelSpeed , i did something similar in my changes as well. i tried running some tests with ginkgo/gomega to see what the errors looked like and it wasn't clear to me that they were printing as neatly as i liked, that's why i added it to the exception message.

i don't have strong feelings about it though, i do think we should agree on a general style for these.

JoelSpeed:
My understanding is that the general format for the errors printed out would be
```
< stack to get to this failure>

< context message being added in this PR >

Expected <something>

<operator>

<the thing that didn't match>
```

So in the case of an error, it would be unexpected error, print errro after the context message, so I don't know we need to duplicate it within the context message, we certainly haven't where we have used this in CPMS

elmiko:
> So in the case of an error, it would be unexpected error, print errro after the context message, so I don't know we need to duplicate it within the context message, we certainly haven't where we have used this in CPMS

sounds good to me, i'll update my patch as well

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -41,31 +41,34 @@ var _ = Describe("MachineHealthCheck", framework.LabelMachineHealthChecks, func(
...
 
 		framework.WaitForMachineSet(client, machineSet.GetName())
 	})
 
 	AfterEach(func() {
 		specReport := CurrentSpecReport()
 		if specReport.Failed() {
-			Expect(gatherer.WithSpecReport(specReport).GatherAll()).To(Succeed())
+			err := gatherer.WithSpecReport(specReport).GatherAll()
+			Expect(err).To(Succeed(), "failed to gather spec report")
```
JoelSpeed, line 62: Can we merge these back together? And the other cases where these have been split?

odvarkadaniel: Yup, sure. Good point.

odvarkadaniel: Reverted this change.

JoelSpeed:
Looks good, are there any `Eventually` or `Consistently` assertions within this package, would be good to make sure those are annotated as well

odvarkadaniel:
> Looks good, are there any `Eventually` or `Consistently` assertions within this package, would be good to make sure those are annotated as well

Nope. There are just `Expect` in this code. Checked earlier while working on this change. 
If you are ok with the annotations, this is good to go @JoelSpeed.

PR #268, opened 2023-02-14
OCPCLOUD-1868: Update k8s to 1.26

RadekManak:


pkg/autoscaler/event.go
```diff
@@ -63,7 +63,11 @@ func newEventWatcher(clientset kubernetes.Interface) *eventWatcher {
 		},
 	})
 
-	return &w
+	if err != nil {
+		return nil, err
```
JoelSpeed, line 67: Always wrap errors please
```suggestion
		return nil, fmt.Errorf("could not add event handler: %w", err)
```

PR #269, opened 2023-02-14
OCPCLOUD-1714: Add exception annotations to infra tests

damdo:
This change adds exception text to the gomega assertions in the infra test suite so that we can more easily detect where failures are occurring.

### DIFF
--- a/pkg/infra/infra.go
+++ b/pkg/infra/infra.go
@@ -157,22 +157,22 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
 		var err error
 
 		gatherer, err = framework.NewGatherer()
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "StateGatherer should be able to be created")
 
 		client, err = framework.LoadClient()
-		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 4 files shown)


JoelSpeed (commented):
Are there any eventually or consistently assertions within this section of the tests?

damdo:
@JoelSpeed almost all are already decorated yes. There were a couple still missing which I added now.

pkg/infra/infra.go
```diff
@@ -253,7 +253,7 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
 				klog.Infof("Did not find all expected taints on the node. Missing: %v", expectedTaints.Difference(observedTaints))
 
 				return false
-			}, framework.WaitMedium, 5*time.Second).Should(BeTrue())
+			}, framework.WaitMedium, 5*time.Second).Should(BeTrue(), "Should find all the expected taints on the Node")
```
JoelSpeed, line 256: As a future todo, we should rewrite this with Komega so that we can use gomega assertions to compare the desired taints, as that would allow better output than a bool here

JoelSpeed (commented):
/approve
/lgtm

pkg/infra/infra.go
damdo: +1

PR #272, opened 2023-02-23
improve balance similar nodes tests

elmiko:
The balance similar nodes test has been quite flaky recently, this change implements a number of fixes to improve the stability of this test. The test has been updated to do the following:

* ensure the machinesets match on the labels they will apply to nodes
* add a wait for the machinesets to take the autoscaler annotations
* add a 1 minute wait to allow for the nodes to gain all their capacity
* add a wait for the nodes to have the same resource capacity keys

These changes are being added to address issues where the cluster autoscaler is getting node records that appear different due to the timing of when the hugepages resources are added.


pkg/autoscaler/autoscaler.go
```diff
@@ -437,12 +439,12 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
 
-			// balance similar nodes requires that all the participating MachineSets have the same
+			// balance similar nodes requires that all the participating Nodes have the same
 			// instance types and labels. while the instance types should be the same, we want to
 			// ensure that no extra labels have been added.
 			// TODO it would be nice to check instance types as well, this will require adding some deserialization code for the machine specs.
-			By("Ensuring both MachineSets have the same labels")
-			Expect(reflect.DeepEqual(transientMachineSets[0].Labels, transientMachineSets[1].Labels)).Should(BeTrue())
+			By("Ensuring both MachineSets have the same .spec.template.spec.labels")
+			Expect(reflect.DeepEqual(transientMachineSets[0].Spec.Template.Spec.Labels, transientMachineSets[1].Spec.Template.Spec.Labels)).Should(BeTrue())
```
JoelSpeed, line 447: Should this expectation have a context added, or be reworked to compare the two maps?
```suggestion
			Expect(transientMachineSets[0].Spec.Template.Spec.Labels).To(Equal(transientMachineSets[1].Spec.Template.Spec.Labels), "Test machine sets should have the same labels")
```

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +459,64 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+			// after they have been initialized. this has been seen with `hugepages-1Gi` and `hugepages-2Mi`. so we
+			// wait until the new nodes have the same resource keys before progressing.
+			By("Waiting for the new Nodes to have similar resources")
+			Eventually(func() (bool, error) {
+				nodes1, err := framework.GetNodesFromMachineSet(client, transientMachineSets[0])
+				if err != nil {
+					return false, err
+				}
+				if len(nodes1) != 1 {
+					return false, fmt.Errorf("Got more nodes than expected for MachineSet %s", transientMachineSets[0].GetName())
```
JoelSpeed, line 490: `more` is not necessarily true, you could have got no node right?

pkg/autoscaler/autoscaler.go
elmiko: sounds good to me

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +459,64 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+					current, err := framework.GetMachineSet(client, machineSet.GetName())
+					if err != nil {
+						return false, err
+					}
+
+					annotations := current.GetAnnotations()
+					_, minFound := annotations[caMinSizeAnnotation]
+					_, maxFound := annotations[caMaxSizeAnnotation]
+					return minFound && maxFound, nil
+				}, framework.WaitShort, pollingInterval).Should(BeTrue(), "MachineSet %s did not acquire autoscaling annotations", machineSet.GetName())
```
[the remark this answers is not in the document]
elmiko: happy for the suggestion, i wasn't aware of doing it this way. thanks

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +459,64 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+			}
+
+			By("Waiting 1 minute to give the new nodes extra time to register their resource capacity")
+			time.Sleep(1 * time.Minute)
```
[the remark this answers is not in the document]
elmiko: well, we don't know exactly "what" to look for, today it's hugepages who knows what it is tomorrow. i figured this was the easiest way to do it.

pkg/autoscaler/autoscaler.go
elmiko: good point

elmiko: i ended up refactoring this code to be a little less repetitive

elmiko:
updated to address comments

elmiko:
looks like i might have introduced a problem with komega

```
Default Komega's client is not set. Use SetClient to set it.
```

pkg/autoscaler/autoscaler.go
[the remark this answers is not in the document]
elmiko: if we want to specifically wait for the hugepages values i can add that. i know the sleep seems a little enigmatic and non-deterministic, but we are dealing with a distributed asynchronous system  XD

pkg/autoscaler/autoscaler.go
```diff
@@ -477,7 +528,12 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 						return 0, err
 					}
 
-					return int(pointer.Int32Deref(current.Spec.Replicas, 0)), nil
+					replicas := pointer.Int32Deref(current.Spec.Replicas, 0)
+					if replicas > expectedReplicas {
+						return 0, StopTrying("exceeded expected replicas in a single MachineSet")
+					} else {
+						return int(replicas), nil
+					}
```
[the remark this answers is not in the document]
elmiko: yeah, i was just being explicit

pkg/autoscaler/autoscaler.go
```diff
@@ -468,6 +518,7 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 			cleanupObjects[workload.GetName()] = workload
 			Expect(client.Create(ctx, workload)).Should(Succeed())
 
+			// TODO make this asynchronous so that we can early exit when one MachineSet is scaled over the expected amount
```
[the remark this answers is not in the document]
elmiko: not necessarily, but i can if we want to do it now.

elmiko:
updated to replace the sleep with a check for the hugepages values, i hope this doesn't cause problems for us later... (<-- famous last words)

elmiko:
i'm still seeing some failures to balance, gonna need to dig in further

elmiko:
i want to add a little more detail to the balance test so that we can see when it overbalances.

/hold

elmiko:
i updated the step where it waits to scale out so that it doesn't need to be asynchronous, gives more information, and can exit early if one of the machinesets gets over-scaled.

/hold cancel

elmiko:
fixing a lint issue

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +461,75 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+				for _, n := range nodes {
+					_, ok1Gi := n.Status.Capacity[corev1.ResourceHugePagesPrefix+"1Gi"]
+					_, ok2Mi := n.Status.Capacity[corev1.ResourceHugePagesPrefix+"2Mi"]
+					if !ok1Gi || !ok2Mi {
+						return false, nil
+					}
+				}
+
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue(), "Node capacity resources did not contain hugepages-1Gi and hugepages-2Mi")
```
JoelSpeed, line 498: Potential rewrite to improve the output here, the improvement would be that if they don't match, the capacity map would be printed out so we could see what is there
```suggestion
			Eventually(func() ( ]*corev1.Node, error) {
				nodes := []*corev1.Node{}

				for _, machineSet := range transientMachineSets {
					if n, err := framework.GetNodesFromMachineSet(client, machineSet); err != nil {
						return false, err
					} else if len(n) != 1 {
						return false, fmt.Errorf("expected 1 node for MachineSet %s, found %d", machineSet.GetName(), len(n))
					} else {
						nodes = append(nodes, n[0])
					}
				}

				return nodes, nil
			}, framework.WaitMedium, pollingInterval).Should(HaveEach(
				HaveField("Status.Capacity", SatisfyAny(
					HaveKey(corev1.ResourceHugePagesPrefix+"1Gi"),
					HaveKey(corev1.ResourceHugePagesPrefix+"2Mi"),
				)),
			), "Node capacity resources did not contain hugepages-1Gi and hugepages-2Mi")
```

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +461,75 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+
+				// if any of the entries does not equal 2, it means that only one of the nodes had that resource
+				for _, v := range resources {
+					if v != 2 {
+						return false, nil
+					}
+				}
+
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue(), "Node capacity resources did not match")
```
JoelSpeed, line 531: We could rewrite this a little to do the final matching using gomega, in that case you'd be able to see which resources didn't match, might help out a little?
```suggestion
			Eventually(func() (map[string]int, error) {
				nodes := []*corev1.Node{}

				for _, machineSet := range transientMachineSets {
					if n, err := framework.GetNodesFromMachineSet(client, machineSet); err != nil {
						return false, err
					} else if len(n) != 1 {
						return false, fmt.Errorf("expected 1 node for MachineSet %s, found %d", machineSet.GetName(), len(n))
					} else {
						nodes = append(nodes, n[0])
					}
				}

				// add the resources from each node's capacity, counting the number of each
				resources := map[corev1.ResourceName]int{}
				for _, n := range nodes {
					for k := range n.Status.Capacity {
						resources[k] += 1
					}
				}

				// Check that each resource in the list has a count of two, if any of the entries does not equal 2, it means that only one of the nodes had that resource
				return resources, nil
			}, framework.WaitMedium, pollingInterval).Should(HaveEach(2), "Node capacity resources did not match")
```
The `HaveEach` matcher will ignore the keys in the map but check that each value equals `2`.

Eg of a failure
```
        Node capacity resources did not match
        Expected
            <map[string]int | len:3>: {"a": 1, "b": 2, "c": 2}
        to contain element matching
            <int>: 2
```

pkg/autoscaler/autoscaler.go
```diff
@@ -468,18 +541,27 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+					replicas := int(pointer.Int32Deref(ms.Spec.Replicas, 0))
+					if replicas > expectedReplicas {
+						return allreplicas, StopTrying(fmt.Sprintf("exceeded expected replicas in MachineSet %s", ms.GetName()))
+					}
+
+					allreplicas = append(allreplicas, replicas)
+				}
+
+				return allreplicas, nil
+			}, framework.WaitOverMedium, pollingInterval).Should(ConsistOf([]int{expectedReplicas, expectedReplicas}), "Failed to balance properly")
```
JoelSpeed, line 564: The following would also work
```suggestion
			}, framework.WaitOverMedium, pollingInterval).Should(HaveEach(expectedReplicas), "Failed to balance properly")
```

JoelSpeed (commented):
Left some suggestions for improving the failure output, WDYT?

pkg/autoscaler/autoscaler.go
elmiko: i wanted to use `HaveExactElements` , but perhaps we need to bump our version of gomega because my compiler wasn't finding it.

https://onsi.github.io/gomega/#haveexactelementselement-interface

pkg/autoscaler/autoscaler.go
elmiko: i like printing the map when they don't match

elmiko:
updated to include latest suggestions

elmiko:
@JoelSpeed i think the e2e-aws-operator failure is interesting in this case, this is what i was wondering about when we talked about waiting for the hugepages variables to exist. not sure why these nodes are so different.

pkg/autoscaler/autoscaler.go
```diff
@@ -457,6 +461,65 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+						nodes = append(nodes, n[0])
+					}
+				}
+
+				return nodes, nil
+			}, framework.WaitMedium, pollingInterval).Should(HaveEach(
+				HaveField("Status.Capacity", SatisfyAny(
+					HaveKey(corev1.ResourceHugePagesPrefix+"1Gi"),
+					HaveKey(corev1.ResourceHugePagesPrefix+"2Mi"),
+				)),
```
elmiko, line 494: @JoelSpeed did i mess up this logic?

it seems like the test is failing but the output contains the expected fields.

JoelSpeed: The issue is typing, the keys you've put in are `string` where the keys in the map are `corev1.ResourceName`, which, happens to be `type ResourceName string`, so, a `BeEquivalentTo` will solve your issue, as this will cast the key before it compares

JoelSpeed, line 494:
```suggestion
				HaveField("Status.Capacity", SatisfyAny(
					HaveKey(BeEquivalentTo(corev1.ResourceHugePagesPrefix+"1Gi")),
					HaveKey(BeEquivalentTo(corev1.ResourceHugePagesPrefix+"2Mi")),
				)),
```

JoelSpeed: I tested this with a simple unit test btw
```
var _ = It("should pass the check", func() {
	nodeBuilder := corev1resourcebuilder.Node()

	node1 := nodeBuilder.Build()
	node2 := nodeBuilder.Build()

	resources := corev1.ResourceList{
		corev1.ResourceCPU:                   resource.MustParse("4"),
		corev1.ResourceMemory:                resource.MustParse("16Gi"),
		corev1.ResourceName("hugepages-2Mi"): resource.MustParse("2Gi"),
		corev1.ResourceName("hugepages-1Gi"): resource.MustParse("1Gi"),
	}

	node1.Status.Capacity = resources
	node2.Status.Capacity = resources

	Expect([]corev1.Node{*node1, *node2}).To(HaveEach(
		HaveField("Status.Capacity", SatisfyAny(
			HaveKey(BeEquivalentTo(corev1.ResourceHugePagesPrefix+"1Gi")),
			HaveKey(BeEquivalentTo(corev1.ResourceHugePagesPrefix+"2Mi")),
		)),
	))
})
```

elmiko: wow, good call

elmiko:
updated the hugepages resource check section to only check the Node.Status.Capacity ResourceLists, and also incorporate Joel's fixes for the typing.

elmiko:
reverted the change to return only the `ResourceList` from the node

elmiko:
test outputs are looking much better imo
/retest

elmiko:
the vsphere test is failing suspiciously early

elmiko:
i've seen a couple failures of the balance test but they look like timeouts waiting for the infrastructure to finish creating the instances

elmiko:
looks like some build system failure happened

/retest-required

elmiko:
updated with latest suggestions to make replicas a map and update that map before exit

JoelSpeed:
Looks good, lets see if it passes

elmiko:
the balance test only failed on azure and it looks like it did not scale out in time. we are only waiting for the replica counts to change and a 5 minute wait should be enough time. not quite sure what the issue here is.

/retest

elmiko:
@JoelSpeed the latest failures are unrelated, i think we will still see occasional failures but hopefully we've reduced the rate of failure now

damdo:
Do we want to cherry pick these fixes?

JoelSpeed:
I think we do
/cherry-pick release-4.13

PR #273, opened 2023-02-27
resourcebuilder: node: add conditions support

damdo:
To be able to create nodes with various conditions (more notably Ready/NotReady).


elmiko (commented):
are these new functions in expectation of a followup PR?

damdo:
Yes we need them in the CPMS for a change on the isReadyMachine behaviour.

PR #278, opened 2023-04-27
Fix spot test alternative machineset logic

RadekManak:
The logic that tries different machineSets when the instance type does not have spot capacity did not remove the old machineSet, but waited for that mahineSet to be deleted. This caused the spot test to fail when there was insufficient capacity for the first instance type it tried.

> MachineSet ci-op-fvk1klgj-efb8d-sn7ndxr57k still exists and does not have a deletion timestamp

This PR adds the missing call to delete the machineSet.


JoelSpeed (commented):
/approve
/lgtm

Lets make sure the spot tests pass before we merge

PR #279, opened 2023-04-27
Make IMDS test more stable

RadekManak:
The curl command is sometimes missing http status header in verbose mode. I am not sure why this happens, but the IMDS works correctly, as seen by it printing correct ami. I hope that this change will make the test more stable by requesting it to print only the status code.

```
  [FAILED] Expected
      <string>: [21 bytes data]
      
100    21  100    21    0     0  21000      0 --:--:-- --:--:-- --:--:-- 21000
      * Closing connection 0
      ami-0e95030500ad17daa
  to contain substring
      <string>: HTTP/1.1 200 OK
```


RadekManak:
Changed code only runs on AWS

JoelSpeed (commented):
/approve
/lgtm

Thanks

PR #280, opened 2023-05-15
OCPCLOUD-2043: Update dependencies to K8s 1.27

damdo:
This updates the dependencies to the latest versions to support K8s 1.27.

TODO: update reference to machine-api-operator once https://github.com/openshift/machine-api-operator/pull/1142 merges


damdo:
/hold
until https://github.com/openshift/machine-api-operator/pull/1142 merges

JoelSpeed:
Have you considered using a global `ctx` variable within the tests instead of calling `context.Background()` the whole time, not ideal but a pattern we've used a lot in tests before

damdo:
Yes I did consider it, although I wasn't sure how long it would take and whether it was worth it.
Anyway had a go at it now and was able to get it sorted in a reasonable amount of effort.

JoelSpeed:
Changes look good, should probably squash the commits up though

JoelSpeed:
/hold cancel

dependent PR merged and this was updated

PR #281, opened 2023-05-31
update autoscaler verbosity to 4

elmiko:
this change makes the autoscaler use a higher verbosity so that the debugging logs are better, it also adds a minor change to the balancing similar nodes test so that we can see the true replica count when a machineset causes the test to fail early.


JoelSpeed:
/approve
/lgtm

Thanks, this should help with the debugging of https://issues.redhat.com/browse/OCPBUGS-14289

JoelSpeed:

We are trying to fix these failures by providing more logging, this shouldn't affect the tests in a negative way so I think lets merge this and then try and get some improved logging in the autoscaler to make progress

PR #282, opened 2023-06-07
OCPBUGS-14289: add a sleep to balance similar nodes test

elmiko:
this change adds a 60 second sleep to the balance similar nodes test so that the new nodes have proper time to acquire their capacity resources, and so that the autoscaler has time to update its records. through experimentation, it appears that 60 seconds is sufficient time for the autoscaler to become aware of the changes. ideally this will be remediated in the future, but for now this change is being proposed to help the stability of this test.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -531,6 +531,19 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func() 
 				return resources, nil
 			}, framework.WaitMedium, pollingInterval).Should(HaveEach(2), "Node capacity resources did not match")
 
+			// in some cases, instance creation is slow for one of the node being introduced
+			// by the previous steps in this test. in these cases, the nodes can be slow to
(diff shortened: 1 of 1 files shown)


elmiko:
hmm, i guess it's back to the drawing board on this one since AWS is failing now

elmiko:
i _think_ this will make the test better, but i would still like to understand why things look so different during the early node creation for the autoscaler. it feels like the autoscaler is running with stale information. i wonder if decreasing the scan interval time (eg down from 15 seconds to 5 seconds) would make a difference.

elmiko:
i've updated the wait time to 1 minute, and also added more information in the comments. i think we should merge this to help keep the testing queues happy, but i also would like to return to this issue once we have the 1.28 rebase from upstream as i have added more logging that will help in the future.

elmiko:
@JoelSpeed ptal, i think this is as good as it will get for now

JoelSpeed (commented):
/approve
/lgtm

We need to time box this investigation for now and we've seen good results adding a sleep. In the future it would be good to understand why the test fails when we don't have the sleep, but we need to move on for now.

PR #283, opened 2023-06-21
Add WithDeletionTimestamp to MachineBuilder

JoelSpeed:
This can be used in unit testing where you aren't using a client, but passing Machine objects around. Need for CPMS testing since I'm moving the API call higher up the stack.


damdo (commented):
/lgtm

PR #285, opened 2023-07-06
openstackFailureDomain: nil rootVolume by default

pierreprinetti:
An empty rootVolume is off-spec and should never happen.


testutils/resourcebuilder/machine/v1beta1/openstack_provider_spec.go
```diff
@@ -30,7 +30,7 @@ func OpenStackProviderSpec() OpenStackProviderSpecBuilder {
 	return OpenStackProviderSpecBuilder{
 		flavor:           "m1.large",
 		availabilityZone: "",
-		rootVolume:       &machinev1alpha1.RootVolume{},
+		rootVolume:       nil,
```
EmilienM, line 33: maybe we could add a comment here to explain why it has to be nil?

EmilienM: Let's have a test, if that's not too complicated. Just in case.

EmilienM (approved):
/lgtm

PR #291, opened 2023-08-22
Add assertion annotations to framework package

nrb:


pkg/framework/machines.go
```diff
@@ -159,5 +159,5 @@ func WaitForMachinesDeleted(c runtimeclient.Client, machines ...*machinev1.Machi
 		}
 
 		return true // Everything was deleted.
-	}, WaitLong, RetryMedium).Should(BeTrue())
+	}, WaitLong, RetryMedium).Should(BeTrue(), "error encounted while deleting Machines.")
```
JoelSpeed, line 162: Maybe this instead? Since this is just waiting?
```suggestion
	}, WaitLong, RetryMedium).Should(BeTrue(), "error encountered while Machines to be deleted.")
```

pkg/framework/machinesets.go
```diff
@@ -67,7 +67,7 @@ func BuildPerArchMachineSetParamsList(ctx context.Context, client runtimeclient.
 	// Get the current workers MachineSets so we can copy a ProviderSpec
 	// from one to use with our new dedicated MachineSet.
 	workers, err := GetWorkerMachineSets(ctx, client)
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).ToNot(HaveOccurred(), "listing machine workers should not error.")
```
JoelSpeed, line 70: This is listing machinesets rather than machines right?

PR #292, opened 2023-08-22
OCPBUGS-17851: Update Azure zone to non-pointer

RadekManak:
Azure's AzureMachineProviderSpec Zone field was updated to be a non-pointer. This PR updates its usage to be compatible again.


elmiko (commented):
/lgtm

elmiko:
hmm, didn't realize that something seems broken in the tests, it seems like maybe an infra flake.

/retest

PR #293, opened 2023-08-25
OCPCLOUD-1717: Add annotations to error output

racheljpg:
This is a PR for [OCPCLOUD-1717](https://issues.redhat.com/browse/OCPCLOUD-1717) to add annotations to error outputs in the AWS E2E tests to make debugging easier.

### DIFF
--- a/pkg/providers/aws.go
+++ b/pkg/providers/aws.go
@@ -35,18 +35,18 @@ var _ = Describe("MetadataServiceOptions", framework.LabelCloudProviderSpecific,
 	BeforeEach(func() {
 		var err error
 		client, err = framework.LoadClient()
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "Failed to load client")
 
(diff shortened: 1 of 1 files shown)


pkg/providers/aws.go
```diff
@@ -35,18 +35,18 @@ var _ = Describe("MetadataServiceOptions", framework.LabelCloudProviderSpecific,
 	BeforeEach(func() {
 		var err error
 		client, err = framework.LoadClient()
-		Expect(err).ToNot(HaveOccurred())
+		Expect(err).ToNot(HaveOccurred(), "Failed to load client")
```
racheljpg, line 38: Should annotations like these be made more clear on which component they're failing to load, or is leaving them like this fine?

elmiko: as long as it is clear where the error originates from, and is not ambiguous with other error messages, then it's fine for me.

elmiko (commented):
thanks @racheljpg !

/lgtm

PR #295, opened 2023-08-28
Update README

racheljpg:
This is a PR to tidy up the README and update a deprecated flag. 
The commands included `$` signs and an example of the expected output, which meant you had to delete both when copying commands from this file into the terminal. I've removed them for easier copying and pasting.
I removed the expected output entirely, unless anyone has any thoughts/preferences on leaving them there underneath their respective commands? 
Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,7 @@ If you do not disable the CVO, when you try to deploy your test code, the CVO wi
 To disable the CVO, scale its deployment to 0 replicas:
 
 ```console
-$ oc scale --replicas 0 -n openshift-cluster-version deployments/cluster-version-operator
-deployment.apps/cluster-version-operator scaled
+oc scale --replicas 0 -n openshift-cluster-version deployments/cluster-version-operator
 ```
 
 Now deploy the code either directly into the cluster, or by running it locally.
@@ -39,8 +38,7 @@ To deploy your test code within the cluster, you must first build and push a con
 Once pushed, override the image within the deployment to deploy your code for testing:
 
 ```console
-$ oc set image -n machine-api-operator deployment/cluster-autoscaler-operator cluster-autoscaler-operator=<YOUR CONTAINER IMAGE>
-deployment.apps/cluster-autoscaler-operator image updated
+oc set image -n machine-api-operator deployment/cluster-autoscaler-operator cluster-autoscaler-operator=<YOUR CONTAINER IMAGE>
 ```
 
 ### Deploy the code to test locally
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,71 +38,36 @@ To deploy your test code within the cluster, you must first build and push a con
...
 Once the operator has been disabled, build your code to test locally and run it on your machine,
 pointing it to the cluster by passing the appropriate `--kubeconfig` flag:
 
 ```console
-$ make build
-  ...
-$ ./bin/cluster-autoscaler-operator --kubeconfig=<PATH/TO/YOUR/CLUSTERS/KUBECONFIG> ...
+make build
+...
+./bin/cluster-autoscaler-operator --kubeconfig=<PATH/TO/YOUR/CLUSTERS/KUBECONFIG> ...
```
racheljpg, line 58: Not quite sure what to do with this and the two lines above - what does the ... do in the terminal? Is it better to split this into two separate commands?

damdo: The three dots there are signalling either waiting or expected output, they are not required IMO.
Feel free to remove them

RadekManak: I think the first three are in place of the output of the first command, and the second three dots are in place of any possible command line arguments you'd want to pass to the program.

README.md
```diff
@@ -39,71 +38,36 @@ To deploy your test code within the cluster, you must first build and push a con
...
-  places nodes evenly across node groups [Slow]
-  /cluster-api-actuator-pkg/pkg/autoscaler/autoscaler.go:605
-•SSSSSSSSSSS
-Ran 4 of 33 Specs in 0.000 seconds
-SUCCESS! -- 0 Passed | 0 Failed | 0 Pending | 29 Skipped
-PASS
-
-Ginkgo ran 1 suite in 1.887727166s
-Test Suite Passed
+NAMESPACE=kube-system ./hack/ci-integration.sh -focus "Autoscaler should" -v --dry-run
```
damdo, line 70: I would keep the `Running Suite: ...` and all the lines below, as they give an example of the output the user can expect from a test run.
Maybe we could put them in a separate code section and mention they are an example output rather than having them in the same block where the command is, to not trick the user. WDYT?

README.md
elmiko: we might want to add some description to this example so that readers know this is show a _different_ operator as an example of how to test.

README.md
elmiko: > Maybe we could put them in a separate code section and mention they are an example output rather than having them in the same block where the command is, to not trick the user. WDYT?

+1 from me

elmiko (commented):
i'm ok with removing the output from most of these, left some comments.

README.md
racheljpg: Thanks everyone for your comments/reviews, I've implemented your changes now. 
@elmiko I'm not sure what you mean by your last comment - a different operator as opposed to which?

README.md
racheljpg: Thanks, I agree with you. I've added it back in now in it's own section

damdo (commented):
Nice one 👏 
/lgtm

README.md
elmiko: my comment is basically speaking to the notion that this part of the instructions (everything that mentions `cluster-autoscaler-operator`) is demonstrating how a user would build a different operator, in this case `cluster-autoscaler-operator`, and then run that operator locally while also running the test suite.

so, imo, we should update this section to describe the scenario that is being given as an example. eg "in this example, we want to run our version of the cluster-autoscaler-operator locally against a remote cluster, and then run the e2e test suite so that we can confirm local changes.". was just a thought while reading all this.

elmiko (commented):
/approve
/lgtm

PR #302, opened 2023-10-24
Let the initial count of nodes for the "scale up and down while respecting MaxNodesTotal" scenario dynamic

aleskandro:
When executing "Scale up and down while respecting MaxNodesTotal" in multi-arch compute nodes clusters with both arm64 and amd64 workers, nodes are more than 6 (8). We can rather consider the initial count by getting the amount of machines running in the cluster before the test runs.

Related to openshift/release#44772

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -688,14 +688,17 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func() 
 		})
 	})
 
-	Context("use a ClusterAutoscaler that has 8 maximum total nodes", framework.LabelPeriodic, func() {
+	Context("use a ClusterAutoscaler that has initial + 2 maximum total nodes", framework.LabelPeriodic, func() {
 		var clusterAutoscaler *caov1.ClusterAutoscaler
-		caMaxNodesTotal := 8
+		var caMaxNodesTotal, initialNodesTotal int
 
 		BeforeEach(func() {
 			gatherer, err = framework.NewGatherer()
 			Expect(err).ToNot(HaveOccurred(), "Failed to create gatherer")
-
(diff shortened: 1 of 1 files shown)


pkg/autoscaler/autoscaler.go
```diff
@@ -688,14 +688,17 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
-		caMaxNodesTotal := 8
+		var caMaxNodesTotal, initialNodesTotal int
 
 		BeforeEach(func() {
 			gatherer, err = framework.NewGatherer()
 			Expect(err).ToNot(HaveOccurred(), "Failed to create gatherer")
-
+			machines, err := framework.GetMachines(ctx, client)
+			Expect(err).ToNot(HaveOccurred(), "Failed to get current machines")
+			initialNodesTotal = len(machines)
```
JoelSpeed, line 700: With the previous assumption, we know that the test framework does not modify existing machines/machinesets. IE, we know that every cluster we test on should have 6 base nodes, and we are adding to that. With this approach, you could infer that there are additional nodes if there's anything left over from other tests when this test starts.

I wonder if there's a way we can make this safer by making sure, for example, all of the machines are in the Running phase? We'd expect the base machines to be running, but perhaps not ones that are being removed after another test.

CC @elmiko keep me honest here, do you agree with my concern?

aleskandro: I'm also considering running the job with 2 amd64 workers and 1 arm64 one to save on costs. But if we agree that counting the Running machines only is reliable too, I would still check in this change as it seems worth it: not to hard code the number of machines, if possible/reliable.

elmiko: hey, sorry me as well. i'll spend some time reviewing

PR #304, opened 2023-10-24
WIP: vendor bump to test Nutanix failure domains

yanhua121:
WIP: vendor bump to test Nutanix failure domains


testutils/resourcebuilder/machine/v1/nutanix_provider_spec.go
```diff
@@ -0,0 +1,132 @@
...
+func (n *NutanixMachineProviderConfigBuilder) WithFailureDomains(withFailureDomains bool) *NutanixMachineProviderConfigBuilder {
+	n.withFailureDomains = withFailureDomains
+	return n
+}
+
+// WithFailureDomainIndex sets the failureDomainIndex field with the input value.
+func (n *NutanixMachineProviderConfigBuilder) WithFailureDomainIndex(index int32) *NutanixMachineProviderConfigBuilder {
+	n.failureDomainIndex = index
+	return n
+}
```
JoelSpeed, line 132: This should take a name to a failure domain rather than an index, this should not rely on the infrastructure object since that can be created with custom failure domains

That or it needs to be passed the failure domain from an external infrastructure object, but it definitely should not be defaulting like that

JoelSpeed (commented):
Looks good, just one concern on the machine provider config builder

testutils/resourcebuilder/machine/v1/nutanix_provider_spec.go
yanhua121: Changed to use failureDomainName instead of index, as suggested.

yanhua121: Changed to allow to set the failureDomains data externally with the "withFailureDomains()" function.

yanhua121:
@elmiko @JoelSpeed The review comments are addressed. Please take another look. thanks.

elmiko (commented):
this looks better to me, i'm happy with this. i'd like @JoelSpeed to give another review tomorrow.

/lgtm

JoelSpeed:

These changes do not effect the E2E in any way

PR #306, opened 2023-12-21
OCPBUGS-9925: Fix tests for OpenStack platform

mandre:
There were several issues, the more serious one was a problem where cleanup would wipe all existing machinesets (the existing ones, before the tests ran) when the tests were skipped.

Moving the cleanup code to a `DeferCleanup()` function fixes it.

### DIFF
--- a/pkg/infra/infra.go
+++ b/pkg/infra/infra.go
@@ -165,6 +165,17 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
 		client, err = framework.LoadClient()
 		Expect(err).ToNot(HaveOccurred(), "Controller-runtime client should be able to be created")
 
+		// Reset the machineSet between each test
+		machineSet = nil
+
+		// Make sure to clean up the resources we created
+		DeferCleanup(func() {
+			if machineSet != nil {
+				By("Deleting the new MachineSet")
+				Expect(client.Delete(ctx, machineSet)).To(Succeed(), "MachineSet should be able to be deleted")
+				framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
+			}
+		})
 	})
 
 	AfterEach(func() {
@@ -172,13 +183,6 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
 		if specReport.Failed() {
 			Expect(gatherer.WithSpecReport(specReport).GatherAll()).To(Succeed(), "StateGatherer should be able to gather resources")
 		}
-
-		if machineSet != nil {
-			By("Deleting the new MachineSet")
-			Expect(client.Delete(ctx, machineSet)).To(Succeed(), "MachineSet should be able to be deleted")
-			framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
-		}
-
 	})
 
 	When("machineset has one replica", func() {
--- a/pkg/infra/infra.go
+++ b/pkg/infra/infra.go
@@ -8,6 +8,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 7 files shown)


mandre:
The CI is finding real bugs now :sweat_smile: 

Fix at https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1803

mandre:
Hey @elmiko, since you inquired about the openstack jobs, PTAL whenever you get a chance.

/assign elmiko

mandre:
e2e-vsphere-operator [has never passed](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-vsphere-operator), and fails on cluster installation, which is outside of the scope of this PR.

elmiko:
thanks @mandre , updates look nice to me and that is wild about the machinesets getting deleted on skips.

/approve
/lgtm

JoelSpeed:
/hold

I want to review this

pkg/infra/infra.go
```diff
@@ -165,20 +166,21 @@ var _ = Describe("Managed cluster should", framework.LabelMachines, func() {
...
 		Expect(err).ToNot(HaveOccurred(), "Controller-runtime client should be able to be created")
 
+		DeferCleanup(func() {
+			if machineSet != nil {
+				By("Deleting the new MachineSet")
+				Expect(client.Delete(ctx, machineSet)).To(Succeed(), "MachineSet should be able to be deleted")
+				framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
+			}
+		})
+
```
JoelSpeed, line 176: This still has a chance to error. `machineSet` is not cleared by the before each, so the `machineSet` could come from a previous test run and be non-nil. Will a delete call fail if the resource doesn't exist? I expect so.

Wouldn't changing this to the below make it safer?
```suggestion
		{
			// Reset the machineSet between each test
			machineSet = nil
			
			// Make sure to clean up the machineSet, if we create one.
			DeferCleanup(func() {
				if machineSet != nil {
					By("Deleting the new MachineSet")
					Expect(client.Delete(ctx, machineSet)).To(Succeed(), "MachineSet should be able to be deleted")
					framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
				}
			}
		})

```

pkg/infra/lifecyclehooks.go
```diff
@@ -84,27 +84,29 @@ var _ = Describe("Lifecycle Hooks should", framework.LabelMachines, func() {
...
+			})).To(Succeed(), "MachineSet should be able to be deleted")
+
+			By("Waiting for the MachineSet to be deleted...")
+			framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
+
+			By("Deleting workload job")
+			Expect(client.Delete(context.Background(), workload, &runtimeclient.DeleteOptions{
+				PropagationPolicy: &cascadeDelete,
+			})).To(Succeed(), "Workload job should be able to be deleted")
+		})
```
JoelSpeed, line 102: Nit: It would be preferable to put the cleanup next to the thing that creates it, I would have split this into several cleanups so that, if something breaks mid setup, we tear down the bits we have already created

The flow should be:
* Create MachineSet
* Defer remove MachineSet
* Wait for MachineSet
* Create workload Job
* Defer remove workload Job
* Wait for workload job running

pkg/infra/spot.go
```diff
@@ -110,31 +110,33 @@ var _ = Describe("Running on Spot", framework.LabelMachines, framework.LabelSpot
...
 			Expect(machineSetReady).To(BeTrue(), "Failed to create a spot backed MachineSet")
 		})
-	})
 
-	AfterEach(func() {
-		specReport := CurrentSpecReport()
-		if specReport.Failed() {
-			Expect(gatherer.WithSpecReport(specReport).GatherAll()).To(Succeed(), "StateGatherer should be able to gather resources")
-		}
+		DeferCleanup(func() {
```
JoelSpeed, line 114: For the sake of making this maintainable, would it make sense to pair this with the line where we clear `delObjects` at the beginning of the before each? Do we risk otherwise not always clearing up if setup breaks mid way through?

pkg/machinehealthcheck/machinehealthcheck.go
```diff
@@ -56,21 +56,23 @@ var _ = Describe("MachineHealthCheck", framework.LabelMachineHealthChecks, func(
...
+
+		DeferCleanup(func() {
+			By("Deleting the MachineHealthCheck resource")
+			Expect(client.Delete(context.Background(), machinehealthcheck)).To(Succeed(), "failed to delete MHC")
+
+			By("Deleting the new MachineSet")
+			Expect(client.Delete(context.Background(), machineSet)).To(Succeed(), "failed to delete machineSet")
+
+			framework.WaitForMachineSetsDeleted(ctx, client, machineSet)
+		})
```
JoelSpeed, line 68: This should come before the `WaitForMachineSet` else it won't run if the wait fails

mandre:
Thanks for the thorough review @JoelSpeed. I believe I have addressed all your comments, PTAL.

JoelSpeed:
/lgtm
/hold cancel

Thanks for fixing that up

PR #312, opened 2024-03-20
OCPCLOUD-2501: Update e2e test to check for updated upstream annotations

racheljpg:
Hello, this is PR for [OCPBUGS-2501](https://issues.redhat.com/browse/OCPCLOUD-2501). 
This is a change to update the current scale from zero tests to check the MachineSets and make sure the new upstream scale from zero annotations are present. Thanks!


pkg/autoscaler/autoscaler.go
```diff
@@ -344,6 +344,17 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 					workloadArch = arch
 					expectedScaledMachineSet = machineSet
 				}
+
+				It("should have expected upstream annotations", func() {
```
elmiko, line 348: we talked about this at standup and think the `It` might be causing issues. try changing it to 

```suggestion
				By("should have expected upstream annotations", func() {
```

pkg/autoscaler/autoscaler.go
```diff
@@ -344,6 +344,17 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+					//checking for the newly added upstream annotations from the CAO
+					expectedAnnotations := map[string]string{
+						"machine.openshift.io/vCPU":     "",
+						"machine.openshift.io/memoryMb": "",
+						"machine.openshift.io/GPU":      "",
+					}
+
+					Expect(machineSet.Annotations).To(HaveValue(Equal(expectedAnnotations)))
```
elmiko, line 356: ultimately we want a different test here.

we probably want to use `HaveKey`, https://pkg.go.dev/github.com/onsi/gomega#HaveKey to test that the keys we want are present.

then for the values we want to do a light validation, so for the CPU and memory values (on the `capacity.cluster.x-k8s.io` annotations) we want to ensure that a `resource.Quantity.MustParse`, https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#MustParse, will succeed. for the GPU count we want to make sure that it is parseable as a number, and for GPU type that it is a string with length greater than 0.

hope this makes sense, but if not you know where to find me =)

elmiko (commented):
couple comments

pkg/autoscaler/autoscaler.go
```diff
@@ -345,6 +347,31 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+						}
+						if _, exists := machineSet.Annotations[annotationsutil.MemoryKey]; !exists {
+							return false, nil
+						}
+						if _, exists := machineSet.Annotations[annotationsutil.GpuCountKey]; !exists {
+							return false, nil
+						}
+						if _, exists := machineSet.Annotations[annotationsutil.GpuTypeKey]; !exists {
+							return false, nil
+						}
```
elmiko, line 369: i'm curious if we changed these to use the ginkgo matchers will we see better error output?

eg instead of `if` clauses, what if we did this:
```
Expect(machineSet.Annotations).To(HaveExistingField(annotationsutil.CpuKey))
```

elmiko (commented):
definitely testing what we want, just wondering if we could get the error output to be a little cleaner.

pkg/autoscaler/autoscaler.go
```diff
@@ -345,6 +347,31 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
 					expectedScaledMachineSet = machineSet
 				}
 			}
+
+			By("should have expected upstream annotations", func() {
```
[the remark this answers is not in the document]
elmiko: also, minor nit, the message for the `By` should read like "by doing X thing", so for example `By("checking for the presence of the scale from zero annotations")`

pkg/autoscaler/autoscaler.go
```diff
@@ -345,6 +365,27 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, Serial, func()
...
+					}
+
+					Expect(ms.Annotations).To(HaveKey(annotationsutil.CpuKey))
+					Expect(ms.Annotations).To(HaveKey(annotationsutil.MemoryKey))
+					Expect(ms.Annotations).To(HaveKey(annotationsutil.GpuCountKey))
+					Expect(ms.Annotations).To(HaveKey(annotationsutil.GpuTypeKey))
+				}
+
+				return true, nil
+			}, framework.WaitMedium, pollingInterval).Should(BeTrue(), "No scale from zero annotations found")
```
elmiko, line 387: we don't strictly need to run this check again on another test, i would be happy with it on only one test, but since it's here i think it's fine to keep.

elmiko (commented):
this is nice, thanks @racheljpg !

/approve
/lgtm

elmiko:
i think the tests are looking good, the openstack failure looked unrelated.

/hold cancel

PR #324, opened 2024-06-26
OCPBUGS-35745: Do not assume the number of nodes

mandre:
This should allow for different type of topology, including compact clusters.

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -3,6 +3,7 @@ package autoscaler
 import (
 	"context"
 	"fmt"
+
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


mandre:
The `Autoscaler should use a ClusterAutoscaler that has 100 maximum total nodes count cleanup deletion information after scale down` test failed with a timeout for both OpenStack and AWS. But looking at the openstack-periodic history, we can see this test previously passed (for example [here](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-cluster-api-actuator-pkg-master-e2e-openstack-operator-periodic/1803234022202019840)), with no code change to the repository since.

Giving it another try.

/test e2e-openstack-operator-periodic-pre e2e-aws-periodic-pre

mandre:
This is ready for reviews.

The CI failure is caused by https://issues.redhat.com//browse/OCPBUGS-36219 and unrelated to this patch.

PR #325, opened 2024-06-26
NOJIRA: adding more logging to deletion taint cleanup test

elmiko:
This change updated the error messages and extends the timeout for the deletion cleanup test to improve the behavior.


nrb:
/lgtm
/approve
/hold
/test e2e-aws-periodic-pre

Previous [failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-api-actuator-pkg/325/pull-ci-openshift-cluster-api-actuator-pkg-master-e2e-aws-periodic-pre/1806301107383701504) seems to be unrelated to this PR. Re-running the test, but would like confirmation from @elmiko that this is different. Remove hold if so.

elmiko:
i think we should leave this on hold for now, i'm not convinced that it solves the problem from the bug and there is also a deeper issue related to the deletion taint cleanup not happening.

elmiko:
i'm cleaning this test up more to help diagnose the underlying problem. i think the test is probably fine, but the failure modes can be clearer.

PR #327, opened 2024-09-05
OCPCLOUD-2703: OWNERS: update subcomponent

nrb:
Signed-off-by: Nolan Brubaker <nolan@nbrubaker.com>

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -18,4 +18,4 @@ reviewers:
 - sub-mod
 - theobarberbany
 component: Cloud Compute
-subcomponent: Other Provider
+subcomponent: Unknown


damdo (commented):
Is this supposed to be a Machine API Provider?
I think this repo will live longer than MAPI, and contains stuff for CPMS and may for CAPI in the future too.

nrb:
The spreadsheet shows MAPI, but it does seem like it's not quite right. @JoelSpeed was this a mistake?

Screenshot 2024-09-05 at 12 31 33 PM

JoelSpeed:
I had put it as MAPI because it's the MAPI E2E suite, but, if you'd rather put it as unknown I'm ok with that, I don't know which it fits better in

damdo:
+1 for Unknown

damdo (commented):
/approve
/lgtm

PR #329, opened 2024-09-23
Add MatchViaJSON matcher to testutils

JoelSpeed:
This adds a new gomega matcher called `MatchViaJSON` that converts objects to JSON before using the gomega `MatchJSON` matcher.

This matcher is useful when structs contain maps that you do not necessarily care about the ordering for. For example, when we are converting machines between mapi and capi and we care about the content of the map, not the order, this matcher can successfully match objects, without the need for the order to be exactly correct.

### DIFF
--- a/testutils/matchers.go
+++ b/testutils/matchers.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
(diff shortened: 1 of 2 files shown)


damdo (commented):
Useful! Thanks!

/approve
/lgtm

PR #331, opened 2024-09-27
OCPCLOUD-2749: testutils: resourcebuilder: extend fields

damdo:
extend fields of resourcebuilder to allow for more flexibility for the following:

- testutils: resourcebuilder: machinev1beta1.AWSProviderSpec extend fields
- testutils: resourcebuilder: machinev1beta1.Machine extend fields
- testutils: resourcebuilder: machinev1beta1.MachineSet extend fields

**Testing**:

To prove this doesn't break existing use cases I ran cluster-control-plane-machine-set-operator unit tests successfully.
```
Ginkgo ran 11 suites in 5m30.502007958s
Test Suite Passed
```

### DIFF
--- a/testutils/resourcebuilder/coalesce.go
+++ b/testutils/resourcebuilder/coalesce.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resourcebuilder
+
+// Coalesce returns the first value (value) provided it is a non-zero value for the type,
+// otherwise it returns a the second value (defaultValue).
+func Coalesce[T comparable](value *T, defaultValue T) T {
+	if value == nil {
+		return defaultValue
+	}
+
+	return *value
+}
--- a/testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
+++ b/testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
@@ -20,48 +20,107 @@ import (
 	"encoding/json"
 
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
 )
 
-// AWSProviderSpec creates a new AWS machine config builder.
-func AWSProviderSpec() AWSProviderSpecBuilder {
-	return AWSProviderSpecBuilder{
-		availabilityZone: "us-east-1a",
-		instanceType:     "m6i.xlarge",
-		securityGroups: []machinev1beta1.AWSResourceReference{
-			{
-				Filters: []machinev1beta1.Filter{
-					{
-						Name: "tag:Name",
-						Values: []string{
-							"aws-security-group-12345678",
-						},
-					},
-				},
+var (
+	defaultAMIID                 = ptr.To[string]("aws-ami-12345678")
+	defaultAvailabilityZone      = "us-east-1a"
+	defaultCredentialsSecretName = "aws-cloud-credentials"
+	defaultDeviceIndex           = int64(0)
+	defaultIAMInstanceProfile    = &machinev1beta1.AWSResourceReference{
+		ID: ptr.To[string]("aws-iam-instance-profile-12345678"),
+	}
+	defaultInstanceType  = "m6i.xlarge"
+	defaultLoadBalancers = []machinev1beta1.LoadBalancerReference{
+		{
+			Type: "network",
+			Name: "aws-nlb-int",
+		},
+		{
+			Type: "network",
+			Name: "aws-nlb-ext",
+		},
+	}
+	defaultBlockDevices = []machinev1beta1.BlockDeviceMappingSpec{
+		{
+			EBS: &machinev1beta1.EBSBlockDeviceSpec{
+				Encrypted:  ptr.To[bool](true),
+				VolumeSize: ptr.To[int64](120),
+				VolumeType: ptr.To[string]("gp3"),
 			},
 		},
-		subnet: machinev1beta1.AWSResourceReference{
+	}
+	// zero-value of the type, to play well with omitempty.
+	defaultMetadataServiceOptions = machinev1beta1.MetadataServiceOptions{}
+	// zero-value of the type, to play well with omitempty.
+	defaultNetworkInterfaceType = machinev1beta1.AWSNetworkInterfaceType("")
+	defaultPlacement            = machinev1beta1.Placement{
+		Region:           defaultRegion,
+		AvailabilityZone: defaultAvailabilityZone,
+	}
+	defaultPlacementGroupName = "" // zero-value of the type, to play well with omitempty.
+	defaultRegion             = "us-east-1"
+	defaultSecurityGroups     = []machinev1beta1.AWSResourceReference{
+		{
 			Filters: []machinev1beta1.Filter{
 				{
 					Name: "tag:Name",
 					Values: []string{
-						"aws-subnet-12345678",
+						"aws-security-group-12345678",
 					},
 				},
 			},
 		},
 	}
+	defaultSubnet = machinev1beta1.AWSResourceReference{
+		Filters: []machinev1beta1.Filter{
+			{
+				Name: "tag:Name",
+				Values: []string{
+					"aws-subnet-12345678",
+				},
+			},
+		},
+	}
+	defaultUserDataSecretName = "aws-user-data-12345678"
+)
+
+// AWSProviderSpec creates a new AWS machine config builder.
+func AWSProviderSpec() AWSProviderSpecBuilder {
+	return AWSProviderSpecBuilder{}
 }
 
 // AWSProviderSpecBuilder is used to build out a AWS machine config object.
+// All the fields of this struct are a pointer to the corresponding original type
+// in the machinev1beta1.AWSMachineProviderConfig. This is done to enable representing
+// the value not being specified in the building chain (so it can be defaulted), versus
+// it being specified, either for setting it to a custom value or to the zero-value of that type.
 type AWSProviderSpecBuilder struct {
-	availabilityZone string
-	instanceType     string
-	securityGroups   []machinev1beta1.AWSResourceReference
-	subnet           machinev1beta1.AWSResourceReference
+	ami                    *machinev1beta1.AWSResourceReference
+	availabilityZone       *string
+	blockDevices           *[]machinev1beta1.BlockDeviceMappingSpec
+	credentialsSecret      **corev1.LocalObjectReference
+	deviceIndex            *int64
+	iamInstanceProfile     **machinev1beta1.AWSResourceReference
+	instanceType           *string
+	keyName                **string
+	loadBalancers          *[]machinev1beta1.LoadBalancerReference
+	metadataServiceOptions *machinev1beta1.MetadataServiceOptions
+	networkInterfaceType   *machinev1beta1.AWSNetworkInterfaceType
+	placement              *machinev1beta1.Placement
+	placementGroupName     *string
+	publicIP               **bool
+	region                 *string
+	securityGroups         *[]machinev1beta1.AWSResourceReference
+	spotMarketOptions      **machinev1beta1.SpotMarketOptions
+	subnet                 *machinev1beta1.AWSResourceReference
+	tags                   *[]machinev1beta1.TagSpecification
+	userDataSecret         **corev1.LocalObjectReference
 }
 
 // Build builds a new AWS machine config based on the configuration provided.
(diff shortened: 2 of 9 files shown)


testutils/resourcebuilder/coalesce.go
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright 2022 Red Hat, Inc.
```
JoelSpeed, line 2: Gone back in time? 👀

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -20,48 +20,90 @@ import (
...
-			},
-		},
-		subnet: machinev1beta1.AWSResourceReference{
+var (
+	defaultInstanceType          = "m6i.xlarge"
+	defaultAvailabilityZone      = "us-east-1a"
+	defaultRegion                = "us-east-1"
+	defaultUserDataSecretName    = "aws-user-data-12345678"
+	defaultCredentialsSecretName = "aws-cloud-credentials"
+	defaultAMIID                 = ptr.To[string]("aws-ami-12345678")
```
JoelSpeed, line 36: Surprised this is a pointer, does it need to be, for the coalesce to work correctly?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -127,26 +154,76 @@ func (m AWSProviderSpecBuilder) BuildRawExtension() *runtime.RawExtension {
...
+// WithAMI sets the AMI for the AWS machine config builder.
+func (m AWSProviderSpecBuilder) WithAMI(ami *machinev1beta1.AWSResourceReference) AWSProviderSpecBuilder {
+	m.ami = ami
+	return m
+}
+
 // WithAvailabilityZone sets the availabilityZone for the AWS machine config builder.
-func (m AWSProviderSpecBuilder) WithAvailabilityZone(az string) AWSProviderSpecBuilder {
+func (m AWSProviderSpecBuilder) WithAvailabilityZone(az *string) AWSProviderSpecBuilder {
 	m.availabilityZone = az
```
JoelSpeed, line 165: Why not take the address of AZ rather than changing the signature?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -127,26 +154,76 @@ func (m AWSProviderSpecBuilder) BuildRawExtension() *runtime.RawExtension {
...
 	return m
 }
 
+// WithBlockDevices sets the BlockDevices for the AWS machine config builder.
+// Use WithBlockDevices(nil), to set them to the zero value.
+func (m AWSProviderSpecBuilder) WithBlockDevices(blockDevices *[]machinev1beta1.BlockDeviceMappingSpec) AWSProviderSpecBuilder {
```
JoelSpeed, line 171: Make this not a pointer, and take the address of blcokDevices?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
damdo: Yes good point! Updated. LMK if the change sounds reasonable and I can extend more fields like so.

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -20,48 +20,94 @@ import (
...
+	return AWSProviderSpecBuilder{
+		availabilityZone: &defaultAvailabilityZone,
+		instanceType:     &defaultInstanceType,
+		securityGroups:   &defaultSecurityGroups,
+		subnet:           &defaultSubnet,
+	}
 }
 
 // AWSProviderSpecBuilder is used to build out a AWS machine config object.
 type AWSProviderSpecBuilder struct {
```
JoelSpeed, line 98: Might just want a godoc to explain why these are all pointers/double pointers

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -71,44 +117,25 @@ func (m AWSProviderSpecBuilder) Build() *machinev1beta1.AWSMachineProviderConfig
...
-		Placement: machinev1beta1.Placement{
-			Region:           "us-east-1",
-			AvailabilityZone: m.availabilityZone,
-		},
-		SecurityGroups: m.securityGroups,
-		Subnet:         m.subnet,
-		UserDataSecret: &corev1.LocalObjectReference{
-			Name: "aws-user-data-12345678",
-		},
+		AMI: CoalesceAWSResourceReference(m.ami, machinev1beta1.AWSResourceReference{
```
JoelSpeed, line 120: The generic Coalesce you wrote doesn't work here?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec_test.go
```diff
@@ -0,0 +1,100 @@
+package v1beta1
+
+import (
+	"testing"
+
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"k8s.io/utils/ptr"
+)
+
+// ------------- DNM: FOR DEMONSTRATION ONLY ----------------
```
JoelSpeed, line 10: Why DNM? Having tests isn't harmful is it?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -127,26 +154,76 @@ func (m AWSProviderSpecBuilder) BuildRawExtension() *runtime.RawExtension {
...
 
 // WithSubnet sets the subnet for the AWS machine config builder.
 func (m AWSProviderSpecBuilder) WithSubnet(subnet machinev1beta1.AWSResourceReference) AWSProviderSpecBuilder {
-	m.subnet = subnet
+	m.subnet = &subnet
+	return m
+}
+
+// WithUserDataSecret sets the UserDataSecret for the AWS machine config builder.
+func (m AWSProviderSpecBuilder) WithUserDataSecret(userDataSecret *corev1.LocalObjectReference) AWSProviderSpecBuilder {
```
JoelSpeed, line 226: Can you add a test that shows what happens when I pass `WithUserDataSecret(nil)`?

testutils/resourcebuilder/machine/v1beta1/coalesce.go
JoelSpeed, line 1: Not sure if these are needed, vs using the generic one?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
damdo: Unfortunately not. That generic Coalesce relies on the provided type being comparable by generics standards.
Slice, map, and function types are not comparable, and so structs that contains them.

so they yield errors like so:
```
[]"github.com/openshift/api/machine/v1beta1".BlockDeviceMappingSpec does not satisfy comparable compiler[InvalidTypeArg](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidTypeArg)
```

Thus I've added some custom Coalesce functions for the types that are not adhering to comparable.

---

Another, more flexible option is to go with `interface{}` rather than generics, WDYT?:

```
// Coalesce returns the first value (value) provided it is a non-zero value for the type,
// otherwise it returns a the second value (defaultValue).
func Coalesce(i interface{}, defaultValue interface{}) interface{} {
	// Use reflection to get the underlying type and value
	val := reflect.ValueOf(i)

	// Check if the interface{} is a pointer
	if val.Kind() != reflect.Ptr {
		panic("Coalesce only accepts pointers to a value")
	}

	// Check if the pointer is nil
	if val.IsNil() {
		return defaultValue
	}

	// Dereference the pointer and return the value it points to
	return val.Elem().Interface()
}
```

And then use it like so:
```
LoadBalancers:      resourcebuilder.Coalesce(m.loadBalancers, defaultLoadBalancers).([]machinev1beta1.LoadBalancerReference),
Placement: resourcebuilder.Coalesce(m.placement, machinev1beta1.Placement{
	Region:           resourcebuilder.Coalesce(m.region, defaultRegion).(string),
	AvailabilityZone: resourcebuilder.Coalesce(m.availabilityZone, defaultAvailabilityZone).(string),
}).(machinev1beta1.Placement),
```

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec_test.go
damdo: I can remove the comment if we are happy with them. Or do we want them to use ginkgo?

testutils/resourcebuilder/machine/v1beta1/coalesce.go
damdo: They are at the moment, see: https://github.com/openshift/cluster-api-actuator-pkg/pull/331#discussion_r1781367355

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec_test.go
JoelSpeed: Would prefer ginkgo if you have the time to convert them across?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
JoelSpeed: No strong opinion either way to be honest, happy to leave as is, but the internal coalesce should be private right?

testutils/resourcebuilder/machine/v1beta1/coalesce.go
```diff
@@ -0,0 +1,53 @@
...
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+)
+
+func CoalesceAWSResourceReference(v1 *machinev1beta1.AWSResourceReference, v2 machinev1beta1.AWSResourceReference) machinev1beta1.AWSResourceReference {
```
JoelSpeed, line 23: Doesn't need to be exported does it?

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec_test.go
damdo: Done

testutils/resourcebuilder/machine/v1beta1/coalesce.go
damdo: Yeah. Unexported now.

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
damdo: Ok, went with generics.

JoelSpeed (commented):
Looking good, do we want to complete the rest of the fields here or follow up for those in a separate PR, the pattern is looking good and the tests are helpful, thanks

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -20,48 +20,100 @@ import (
...
+	}
+)
+
+// AWSProviderSpec creates a new AWS machine config builder.
+func AWSProviderSpec() AWSProviderSpecBuilder {
+	return AWSProviderSpecBuilder{
+		availabilityZone: &defaultAvailabilityZone,
+		instanceType:     &defaultInstanceType,
+		securityGroups:   &defaultSecurityGroups,
+		subnet:           &defaultSubnet,
```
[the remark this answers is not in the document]
damdo: I wasn't sure about this, but I checked the main user of this pkg (CPMSO) and we don't need to 👍 So I removed them.

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -127,26 +169,124 @@ func (m AWSProviderSpecBuilder) BuildRawExtension() *runtime.RawExtension {
...
+
+// WithBlockDevices sets the BlockDevices for the AWS machine config builder.
+// Use WithBlockDevices(nil), to set them to the zero value.
+func (m AWSProviderSpecBuilder) WithBlockDevices(blockDevices []machinev1beta1.BlockDeviceMappingSpec) AWSProviderSpecBuilder {
+	m.blockDevices = &blockDevices
+	return m
+}
+
+// WithCredentialsSecret sets the DeviceIndex for the AWS machine config builder.
+func (m AWSProviderSpecBuilder) WithDeviceIndex(deviceIndex int64) AWSProviderSpecBuilder {
```
JoelSpeed, line 192: Can we make sure that all of the functions are alphabetically ordered, and have the correct godoc, this one doesn't for example

testutils/resourcebuilder/machine/v1beta1/aws_provider_spec.go
```diff
@@ -127,26 +169,124 @@ func (m AWSProviderSpecBuilder) BuildRawExtension() *runtime.RawExtension {
...
+}
+
+// WithMetadataServiceOptions sets the MetadataServiceOptions for the AWS machine config builder.
+func (m AWSProviderSpecBuilder) WithMetadataServiceOptions(opts machinev1beta1.MetadataServiceOptions) AWSProviderSpecBuilder {
+	m.metadataServiceOptions = &opts
+	return m
+}
+
+// WithSpotMarketOptions sets the SpotMarketOptions for the AWS machine config builder.
+func (m AWSProviderSpecBuilder) WithSpotMarketOptions(opts *machinev1beta1.SpotMarketOptions) AWSProviderSpecBuilder {
```
JoelSpeed, line 235: Should come later, after SecurityGroups?

damdo: TY

testutils/resourcebuilder/machine/v1beta1/machine.go
```diff
@@ -30,47 +37,69 @@ func Machine() MachineBuilder {
...
+	generateName          string
+	labels                map[string]string
+	lifecycleHooks        *machinev1beta1.LifecycleHooks
+	machineSpec           *machinev1beta1.MachineSpec
+	machineSpecObjectMeta *machinev1beta1.ObjectMeta
+	name                  string
+	namespace             string
+	providerID            **string
+	providerSpecBuilder   *resourcebuilder.RawExtensionBuilder
+	providerSpec          *machinev1beta1.ProviderSpec
```
JoelSpeed, line 53: Do we have a builder function for this one?

testutils/resourcebuilder/machine/v1beta1/machine.go
```diff
@@ -30,47 +37,69 @@ func Machine() MachineBuilder {
...
-	providerSpecBuilder resourcebuilder.RawExtensionBuilder
-	providerID          *string
+	authoritativeAPI      *machinev1beta1.MachineAuthority
+	annotations           map[string]string
+	creationTimestamp     metav1.Time
+	deletionTimestamp     *metav1.Time
+	generateName          string
+	labels                map[string]string
+	lifecycleHooks        *machinev1beta1.LifecycleHooks
+	machineSpec           *machinev1beta1.MachineSpec
```
JoelSpeed, line 47: Do we have a builder function for this?

testutils/resourcebuilder/machine/v1beta1/machine.go
```diff
@@ -30,47 +37,69 @@ func Machine() MachineBuilder {
...
+	machineSpecObjectMeta *machinev1beta1.ObjectMeta
+	name                  string
+	namespace             string
+	providerID            **string
+	providerSpecBuilder   *resourcebuilder.RawExtensionBuilder
+	providerSpec          *machinev1beta1.ProviderSpec
+	taints                []corev1.Taint
```
JoelSpeed, line 54: Do we have a builder function for this?

damdo:
All 3 commits now pushed for AWSProvider, Machine and MachineSet resourcebuilder extensions.

@JoelSpeed this is ready for a final review

testutils/resourcebuilder/machine/v1beta1/machineset.go
```diff
@@ -88,32 +142,123 @@ func (m MachineSetBuilder) AsWorker() MachineSetBuilder {
 		WithLabel(machineSetMachineTypeLabelName, "worker")
 }
 
+// WithAnnotations sets the annotations for the machineSet on the machineSet builder.
+// DEPRECATED: use WithMachineSetAnnotations() instead.
```
JoelSpeed, line 146: Why is this special and prefixing the labels/annotations with MachineSet? I wouldn't expect that on other resources.

Do we do something similar on CPMS?

testutils/resourcebuilder/machine/v1beta1/machineset.go
```diff
@@ -27,55 +28,108 @@ const (
...
 
+var (
+	defaultAvailableReplicas      = int32(0)
+	defaultDeletePolicy           = ""
+	defaultFullyLabeledReplicas   = int32(0)
+	defaultMinReadySeconds        = int32(0)
+	defaultObservedGeneration     = int64(0)
+	defaultReadyReplicas          = int32(0)
+	defaultReplicasStatus         = int32(0)
+	defaultSynchronizedGeneration = int64(0)
```
JoelSpeed, line 39: Do we really need default values when they are the zero value for the field?

testutils/resourcebuilder/machine/v1beta1/machineset.go
```diff
@@ -27,55 +28,108 @@ const (
...
 		},
 		Spec: machinev1beta1.MachineSetSpec{
-			Replicas: &m.replicas,
+			AuthoritativeAPI: resourcebuilder.Coalesce(m.authoritativeAPI, defaultAuthoritativeAPI),
+			DeletePolicy:     resourcebuilder.Coalesce(m.deletePolicy, defaultDeletePolicy),
+			MinReadySeconds:  resourcebuilder.Coalesce(m.minReadySeconds, defaultMinReadySeconds),
+			Replicas:         m.replicas,
 			Selector: metav1.LabelSelector{
-				MatchLabels: m.labels,
+				MatchLabels: m.machineTemplateLabels,
```
JoelSpeed, line 100: Is there an occasion you can think of where we might want to try an alternative selector?

testutils/resourcebuilder/machine/v1beta1/machineset.go
damdo: We do have MachineSet (top level annotations/labels on the resource), and we then have the machineTemplate ones, which are then propagated to the children Machines. I wanted to have a way to differentiate the two, as in real MachineSet they could be different right?

testutils/resourcebuilder/machine/v1beta1/machineset.go
damdo: The selector should be matchng the labels in the Template.ObjectMeta.Labels below, so this and that one should match.
Although the MachineSet.ObjectMeta.Labels, are different right?

JoelSpeed: Yes, MachineSet.ObjectMeta.Labels are different.

My thinking here was that perhaps we might want to use this builder to provide a case where the selector uses match expressions, or, doesn't match the labels on the template.

I think having a builder that allows providing an entire selector, and then defaulting to what you've included here may be a more flexible solution

testutils/resourcebuilder/machine/v1beta1/machineset.go
JoelSpeed: I agree that these should be different, but wondered why we don't just call these top level ones `WithLabels` and `WithAnnotations`, since that's what we do on other objects isn't it?

The ones within the spec I'm happy to be named differently, but thought we should be consistent on the top level object meta ones

damdo: Yeah I did that to avoid confusion, but I agree having consistency is also desireable.  In that case we will need to set both labels to the same for AsWorker() now.

Addressed that in the latest commit.

Also testcases that currently use MachineSet.WithLabels() will need updating as now  WithLabels() will only update ms.Labels and not also ms.Spec.Selector.MatchLabels. (e.g. https://github.com/search?q=repo%3Aopenshift%2Fmachine-api-operator%20WithLabel&type=code )

testutils/resourcebuilder/machine/v1beta1/machineset.go
damdo: Makes sense. Extended the builder to accept a custom selector with `WithMachineSetSpecSelector()`, this may then differ from the machineSet.Spec.Template.ObjectMeta.Labels, which can separately be set with `WithMachineTemplateLabels()`.

testutils/resourcebuilder/machine/v1beta1/machineset.go
```diff
@@ -34,48 +35,91 @@ func MachineSet() MachineSetBuilder {
...
-	creationTimestamp   metav1.Time
-	providerSpecBuilder resourcebuilder.RawExtensionBuilder
+	annotations                map[string]string
+	authoritativeAPI           machinev1beta1.MachineAuthority
+	creationTimestamp          metav1.Time
+	deletePolicy               string
+	deletionTimestamp          *metav1.Time
+	generateName               string
+	lifecycleHooks             machinev1beta1.LifecycleHooks
+	machineSetLabels           map[string]string
```
JoelSpeed, line 45: Inconsistent with annotations?

JoelSpeed (commented):
/lgtm
/approve

PR #332, opened 2024-10-08
NO-JIRA: Downgrade testutils to k8s v1.30

damdo:
cluster-capi-operator will otherwise need to bump to 1.31 and that's not compatible with CAPI 1.8


JoelSpeed:
> cluster-capi-operator will otherwise need to bump to 1.31 and that's not compatible with CAPI 1.8

How compatible does it actually need to be, operator to operand level? What do we import from CAPI into the operator? It should just be APIs in which case, this shouldn't matter I would have thought

Did you hit specific errors when trying to bump?

damdo:
Yeah I don't necessarily think this will cause issues, as, like you said, in the C-CAPI-O we are importing k8s/CAPI only for API purposes, and there shouldn't be breaking changes there.

I'm haven't yet run CI against https://github.com/openshift/cluster-capi-operator/pull/215 (which forces the 1.31 bump), but I put up a PR to be extra cautious in case they break. Will check if we hit any issues first before carrying on merging this

PR #348, opened 2024-12-04
NO-JIRA: resourcebuilder: configv1/infrastructure: allow setting more status fields

damdo:
Allow setting more status fields for the configv1/infrastructure.

### DIFF
--- a/testutils/resourcebuilder/config/v1/infrastructure.go
+++ b/testutils/resourcebuilder/config/v1/infrastructure.go
@@ -354,6 +354,17 @@ func (i InfrastructureBuilder) WithGenerateName(generateName string) Infrastruct
 	return i
 }
 
+// WithInfrastructureName sets the infrastructureName in the status for the infrastructure builder.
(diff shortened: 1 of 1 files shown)


testutils/resourcebuilder/config/v1/infrastructure.go
```diff
@@ -382,3 +393,20 @@ func (i InfrastructureBuilder) WithNamespace(namespace string) InfrastructureBui
...
+	if i.status == nil {
+		i.status = &configv1.InfrastructureStatus{}
+	}
+
+	i.status.PlatformStatus = &ps
+
+	return i
+}
+
+// WithNamespace sets the status for the infrastructure builder.
```
elmiko, line 408: update the doc string here, also do we need a nil check here as well?

elmiko (commented):
minor nit, otherwise lgtm

testutils/resourcebuilder/config/v1/infrastructure.go
damdo: Ahh thanks for catching this. Fixed.

elmiko (commented):
/lgtm

PR #363, opened 2024-12-24
OCPQE-26835: Auto capi cases on azure: spot/disk encryption/accelerated network

sunzhaohua2:


sunzhaohua2:
The failure is not related to this pr

pkg/framework/utils.go
```diff
@@ -113,3 +113,7 @@ func GetControlPlaneHostAndPort(ctx context.Context, cl client.Client) (string,
 
 	return apiURL.Hostname(), int32(port), nil
 }
+
+func BoolPtr(v bool) *bool {
```
JoelSpeed, line 117: Perhaps use the `ptr.To` pattern from [k8s.io/utils/ptr](https://pkg.go.dev/k8s.io/utils/ptr) instead?

sunzhaohua2: Updated

PR #370, opened 2025-01-07
OCPQE-27708 - Annotationtestsgcp

miyadav:
Adding annotation tests on gcp . This is work in progress , any suggestions are welcome.

### DIFF
--- a/pkg/annotations/gcp.go
+++ b/pkg/annotations/gcp.go
@@ -1,0 +1,266 @@
+package annotations
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+
+	"github.com/openshift/cluster-api-actuator-pkg/pkg/framework"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	namespace = "openshift-machine-api"
+
+	// Timeout constants for service operations.
+	serviceTimeout      = 3 * time.Minute
+	servicePollInterval = 10 * time.Second
+)
+
+var cl client.Client
+
+var _ = Describe("Service Annotation tests GCP", framework.LabelCCM, framework.LabelDisruptive, Ordered, func() {
+	var (
+		ctx             context.Context
+		platform        configv1.PlatformType
+		createdServices []string
+		gcpClient       *framework.GCPClient
+	)
+
+	BeforeAll(func() {
+		cfg, err := config.GetConfig()
+		Expect(err).ToNot(HaveOccurred(), "Failed to GetConfig")
+
+		cl, err = client.New(cfg, client.Options{})
+		Expect(err).NotTo(HaveOccurred(), "Failed to create Kubernetes client for test")
+		komega.SetClient(cl)
+		ctx = framework.GetContext()
+		platform, err = framework.GetPlatform(ctx, cl)
+		fmt.Println("platform is ", platform)
+		Expect(err).ToNot(HaveOccurred(), "Failed to get platform")
+		if platform != configv1.GCPPlatformType {
+			Skip("Skipping GCP E2E tests")
+		}
+
+		// Get GCP credentials from infrastructure
+		project, region, err := framework.GetGCPCredentialsFromInfrastructure(ctx, cl)
+		Expect(err).ToNot(HaveOccurred(), "Failed to get GCP credentials from infrastructure")
+		gcpClient = framework.NewGCPClient(project, region)
+	})
+
+	AfterAll(func() {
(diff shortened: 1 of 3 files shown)


miyadav:
once [PR](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/379) is merged  the new added test shall pass .

sunzhaohua2:
@miyadav Thanks for working on this, it might be better if we could check result from gcp cli，or we dont know if it take affect. Take` cloud.google.com/network-tier annotation `for example, the step may be: 

1. create service with annotation cloud.google.com/network-tier: Standard
2. check with gcp cli that LB with network-tier: Standard
3. update service annotation to network-tier: Premium
4. check with gcp cli that LB with network-tier: Premium
5. remove the annoation
6. check with gcp cli that LB with network-tier: Standard
7. create with invalidvalue
8. check create failed

pkg/annotations/gcp.go
```diff
@@ -0,0 +1,141 @@
...
+		komega.SetClient(cl)
+		ctx = framework.GetContext()
+		platform, err = framework.GetPlatform(ctx, cl)
+		fmt.Println("platform is ", platform)
+		o.Expect(err).ToNot(o.HaveOccurred(), "Failed to get platform")
+		if platform != configv1.GCPPlatformType {
+			g.Skip("Skipping GCP E2E tests")
+		}
+
+		namespace = "default"
```
sunzhaohua2, line 57: maybe set namespace as const and change to another namespace.

pkg/annotations/gcp.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		o.Expect(cl.Create(ctx, service)).To(o.Succeed())
+		createdServices = append(createdServices, service.Name)
+
+		var lastIngressIP string
+		o.Eventually(func() (string, error) {
+			updatedService := &corev1.Service{}
+			err := cl.Get(ctx, client.ObjectKey{Name: service.Name, Namespace: namespace}, updatedService)
+			if err != nil {
+				return "", err
```
sunzhaohua2, line 95: `return "", fmt.Errorf("failed to get service: %v", err)`

pkg/annotations/gcp.go
```diff
@@ -0,0 +1,141 @@
...
+					latestService.Annotations = make(map[string]string)
+				}
+				latestService.Annotations[key] = value
+
+				if key == "cloud.google.com/network-tier" && value != "Standard" && value != "Premium" {
+					o.Expect(cl.Update(ctx, latestService)).ToNot(o.Succeed(), "The annotation 'cloud.google.com/network-tier', if specified, must be either 'Standard' or 'Premium'")
+					continue
+				}
+
+				o.Expect(cl.Update(ctx, latestService)).To(o.Succeed())
```
sunzhaohua2, line 121: `o.Expect(cl.Update(ctx, latestService)).To(o.Succeed(), "Failed to update service with annotation")`

miyadav:
> @miyadav Thanks for working on this, it might be better if we could check result from gcp cli，or we dont know if it take affect. Take`cloud.google.com/network-tier annotation`for example, the step may be:
> 
> 1. create service with annotation cloud.google.com/network-tier: Standard
> 2. check with gcp cli that LB with network-tier: Standard
> 3. update service annotation to network-tier: Premium
> 4. check with gcp cli that LB with network-tier: Premium
> 5. remove the annoation
> 6. check with gcp cli that LB with network-tier: Standard
> 7. create with invalidvalue
> 8. check create failed

We are checking the updated IP ( line [136](https://github.com/openshift/cluster-api-actuator-pkg/blob/9383e0a8f0891e9129baa7b53b35b6a9355fcadd/pkg/annotations/gcp.go#L136) to confirm it took effect . 
I will review more for the updates to confirm from cli.

miyadav:
> @miyadav Thanks for working on this, it might be better if we could check result from gcp cli，or we dont know if it take affect. Take`cloud.google.com/network-tier annotation`for example, the step may be:
> 
> 1. create service with annotation cloud.google.com/network-tier: Standard
> 2. check with gcp cli that LB with network-tier: Standard
> 3. update service annotation to network-tier: Premium
> 4. check with gcp cli that LB with network-tier: Premium
> 5. remove the annoation
> 6. check with gcp cli that LB with network-tier: Standard
> 7. create with invalidvalue
> 8. check create failed

I am not sure how to use gcp cli , do you think something like below , I try running it seeing some issues , not sure if this is helpful as well , we would be required to know many details in such case and would be too deep into networking :thinking: 

```
`func verifyGCPAnnotation(svcName, annotation, expectedValue string) bool {
	var cmd string

	switch annotation {
	case "cloud.google.com/network-tier":
		cmd = fmt.Sprintf("gcloud compute forwarding-rules describe %s --format='value(networkTier)'", svcName)
	case "networking.gke.io/internal-load-balancer-allow-global-access":
		cmd = fmt.Sprintf("gcloud compute forwarding-rules describe %s --format='value(allowGlobalAccess)'", svcName)
	case "alpha.cloud.google.com/load-balancer-backend-share":
		cmd = fmt.Sprintf("gcloud compute backend-services describe %s --format='value(backendShare)'", svcName)
	default:
		fmt.Println("No GCP CLI validation available for annotation:", annotation)
		return true
	}`
```

miyadav:
@sunzhaohua2 I have tried adding gcp cli validation as well now. PTAL when time permits.

huali9:
This basically LGTM, but I'd like to leave it to @sunzhaohua2 as she had some comments before.

PR #376, opened 2025-01-21
OSASINFRA-3689: Add testutils builders for CAPO

stephenfin:
As with #371, we want this logic in `cluster-capi-operator` and might as well keep it here along with the other CAPI provider implementations.


stephenfin:
Words cannot describe how tedious this was to write... :sweat_smile:

stephenfin:
/cc @mandre
/cc @JoelSpeed

I need this to implement the unit tests for the OpenStack `capi2mapi` implementation in `cluster-capi-operator`.

damdo:
> Words cannot describe how tedious this was to write... 😅

I feel the pain, did the same for most of MAPI and CAPI 😂

damdo (commented):
Hey @stephenfin thanks for the PR, the changes looks great.

Sorry to be a bother, but would you be able to set up BDD tests like we did for the AWS provider here: https://github.com/openshift/cluster-api-actuator-pkg/blob/master/testutils/resourcebuilder/cluster-api/infrastructure/v1beta2/awscluster_test.go

Thanks!

stephenfin:
> Sorry to be a bother, but would you be able to set up BDD tests like we did for the AWS provider here: https://github.com/openshift/cluster-api-actuator-pkg/blob/master/testutils/resourcebuilder/cluster-api/infrastructure/v1beta2/awscluster_test.go

I've added these tests. I didn't test every field since I'm thinking Go will do a good enough job ensuring I am using the correct types and we just want to test the patterns. Also, writing these for every field is going to be even more painful :sweat_smile: Please let me know if this isn't sufficient though.

nrb (approved):

PR #377, opened 2025-02-03
OCPCLOUD-2833: Update to k8s 1.32

damdo:
This PR updates the module dependencies to be inline with K8s 1.32 ahead of 4.19.

TODO:
- [x] update MAO dep once https://github.com/openshift/machine-api-operator/pull/1325 merges


testutils/go.mod
```diff
@@ -1,88 +1,87 @@
...
-	k8s.io/api v0.31.1
-	k8s.io/apimachinery v0.31.1
-	k8s.io/client-go v0.31.0
-	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
-	sigs.k8s.io/cluster-api v1.8.5
-	sigs.k8s.io/cluster-api-provider-aws/v2 v2.6.1
+	k8s.io/api v0.31.3
+	k8s.io/apimachinery v0.31.3
+	k8s.io/client-go v0.31.3
+	k8s.io/utils v0.0.0-20241210054802-24370beab758
```
JoelSpeed, line 14: Does the code compile if these are moved to 0.32.1?

go.mod
```diff
@@ -1,97 +1,106 @@
...
-	github.com/openshift/cluster-autoscaler-operator v0.0.1-0.20240509123215-40cadf8a4729
-	github.com/openshift/library-go v0.0.0-20240919205913-c96b82b3762b
-	github.com/openshift/machine-api-operator v0.2.1-0.20240924183942-9c3e4a04009a
+	github.com/onsi/ginkgo/v2 v2.22.2
+	github.com/onsi/gomega v1.36.2
+	github.com/openshift/api v0.0.0-20250131155403-30a036067514
+	github.com/openshift/cluster-api-actuator-pkg/testutils v0.0.0-20250122171707-86066d47a264
+	github.com/openshift/cluster-autoscaler-operator v0.0.1-0.20241204142113-43631b045675
+	github.com/openshift/library-go v0.0.0-20250129210218-fe56c2cf5d70
+	github.com/openshift/machine-api-operator v0.2.1-0.20250122132849-b397e885dafb
```
damdo, line 17: We need not to be on the latest MAO as it would force controller-runtime 0.20, which would break the CAPI API dependencies, which in their API use constructs from c-r which have been removed in c-r 0.20

testutils/go.mod
damdo: For the testutil submodule I don't think we can go up to 0.32, as that would force:
```
go: k8s.io/api@v0.32.1 requires go >= 1.23.0; switching to go1.23.6
```

Which then would force the cluster-capi-operator to be on 1.23, which we've always kept in line with the CAPI providers (they will be on 1.22).

JoelSpeed: 4.19 will be built with Go 1.23

I don't think that means that the go.mod minimum version needs to be Go 1.23, but I certainly don't think moving the minimum go.mod version here matters, we don't import this into any of the CAPI repos do we?

damdo: > 4.19 will be built with Go 1.23

Correct

> we don't import this into any of the CAPI repos do we?
We do in the cluster-capi-operator. I've always tried to keep the minimum version in sync with the CAPI ecosystem. That said it is probably not a hard requirement.

JoelSpeed: I don't think that's going to be a hard requirement, lets not keep that restriction if possible, I don't think that's going to cause us issues

damdo: Made the switch to 1.23 also for testutils.

go.mod
```diff
@@ -1,97 +1,106 @@
...
+	k8s.io/apimachinery v0.32.1
+	k8s.io/client-go v0.32.1
 	k8s.io/klog v1.0.0
-	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
-	sigs.k8s.io/cluster-api v1.8.4
-	sigs.k8s.io/cluster-api-provider-aws/v2 v2.6.1
+	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
+	sigs.k8s.io/cluster-api v1.9.4
+	sigs.k8s.io/cluster-api-provider-aws/v2 v2.7.1
 	sigs.k8s.io/cluster-api-provider-azure v1.17.1
```
theobarberbany, line 26: nit: curious as to why azure and gcp aren't being bumped here

testutils/Makefile
```diff
@@ -6,7 +6,7 @@ GOPROXY ?=
 export GOPROXY
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.28
+ENVTEST_K8S_VERSION = 1.31
```
theobarberbany, line 9: 1.32.1? or are we not using the openshift builds for a reason here 🤔

go.mod
damdo: It doesn't matter much as we only use APIs for those, ideally they should be the minor release that follows the twin CAPI minor release (in this case CAPI 1.9), which is not out for all the providers. I'll bump them again across repos when we have also CAPA and CAPG releases upstream that follow CAPI 1.9, but in the end it is not really important as it is API only stuff.

testutils/Makefile
damdo: We should, originally I didn't bump the whole repo to 1.32 for CAPI compat reasons.
Now that I did, this should be updated too.

Pushed a change to do this.

go.mod
theobarberbany: Gotcha, thanks for explaining! :)

PR #379, opened 2025-02-12
Adds ability to set project to gcp provider builder

theobarberbany:
This change updates the gcp provider spec builder to allow setting of project ids. It retains the current default.


damdo (commented):
/approve
/lgtm

PR #382, opened 2025-02-20
OCPQE-28631: update golang version

sunzhaohua2:


JoelSpeed:
The go version should be dictated by ART, I'm not sure we can update the go version on older releases normally. We may be ok here because it's a testing only repo, but what exactly is the issue that prompted this?

sunzhaohua2:
> The go version should be dictated by ART, I'm not sure we can update the go version on older releases normally. We may be ok here because it's a testing only repo, but what exactly is the issue that prompted this?

Not quite sure the issue, only dockerfile's golang version is not correct for 4.15.

PR #392, opened 2025-03-29
increase the timeout for scaling out

elmiko:
this change makes the timeout 15 minutes for the autoscaler test where we scale out to maximum size. on some platforms, it appears that 5 minutes is not enough time.


damdo (commented):
/approve
/lgtm

PR #403, opened 2025-04-23
[release-4.17] Set project and disks on GCP provider builder.

theobarberbany:


testutils/resourcebuilder/machine/v1beta1/gcp_provider_spec.go
```diff
@@ -28,15 +28,27 @@ import (
 // GCPProviderSpec creates a new GCP machine config builder.
 func GCPProviderSpec() GCPProviderSpecBuilder {
 	return GCPProviderSpecBuilder{
+		disks: []*machinev1beta1.GCPDisk{
```
RadekManak, line 31: Are you sure that adding a new default values won't break any consumers?

RadekManak: Just noticed. It's just moved from below. Ignore.

PR #353, opened 2024-12-20
OCPQE-27646: add label for the proxy case

huali9:
[It] When cluster-wide proxy is configured, Machine API cluster operator should create machines when configured behind a proxy

There are many jobs in qe ci, but this case failed on disconnected jobs when DeployProxy, because the url cannot be accessed on disconnected cluster, so I'd like to mark this case only run in dev env.
@sunzhaohua2 @miyadav @shellyyang1989 PTAL, thanks!

PR #411, opened 2025-06-06
OCPQE-27646: addlabel2

huali9:
Refer https://github.com/openshift/cluster-api-actuator-pkg/pull/353#issuecomment-2948288605, but I checked the code of dynamically checking if it's a disconnected cluster, it's complicated and time consuming, so I prefer to use label
@sunzhaohua2 @miyadav @shellyyang1989 PTAL, thanks!


shellyyang1989:
looks good to me. leaving it to team for review. Thanks

PR #426, opened 2025-08-25
AUTOSCALE-254: upgrade autoscaler e2e tests to detect workload placement

maxcao13:
Adds checks in the autoscaler e2e tests to confirm workloads actually run after machines are provisioned by the autoscaler.


elmiko:
putting this in my queue

elmiko (commented):
this is looking good so far, but something that i would like the test to confirm is that the workload landed on a node associated with the machineset. we might need to pass the machineset to the `WaitForWorkload` function so that we could detect the placement.

maxcao13:
Thanks Mike, that makes sense. I'll put some changes in :-)

elmiko (commented):
this is looking good, one small suggestion that might make this a little cleaner.

PR #431, opened 2025-09-10
NO-ISSUE: Add image option to Azure resourcebuilder

djoshy:
This adds an image option to the Azure resourcebuilder for use in https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/368


damdo:
/verified bypass
This is testing helpers, no need to verify

PR #429, opened 2025-09-05
enhance capi spot/dedicated cases for InsufficientInstanceCapacity error

huali9:

PR #437, opened 2025-09-18
If there is still no capacity after retrying, skip instead of fail

huali9:
Unfortunately, after https://github.com/openshift/cluster-api-actuator-pkg/pull/429 merged, we still observed that spot case failed twice: [1967560045713428480](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/360/pull-ci-openshift-cluster-capi-operator-main-regression-clusterinfra-aws-ipi-techpreview-capi/1967560045713428480) and [1968297529997004800](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-capi-operator/363/pull-ci-openshift-cluster-capi-operator-main-regression-clusterinfra-aws-ipi-techpreview-capi/1968297529997004800) There is still no capacity after 4 times retry. I think we should skip the case not fail it when there is still no capacity after retrying. Another, the case failed in 16 minutes but it should fail quickly, so I also updated the AfterEach function. @damdo @theobarberbany @sunzhaohua2 PTAL, thanks!

### DIFF
--- a/pkg/capi/aws.go
+++ b/pkg/capi/aws.go
@@ -81,9 +81,13 @@ var _ = Describe("Cluster API AWS MachineSet", framework.LabelCAPI, framework.La
 	})
 
 	AfterEach(func() {
-		framework.DeleteCAPIMachineSets(ctx, cl, machineSet)
-		framework.WaitForCAPIMachineSetsDeleted(ctx, cl, machineSet)
-		framework.DeleteObjects(ctx, cl, awsMachineTemplate)
+		if machineSet != nil {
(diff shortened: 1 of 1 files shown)


damdo (commented):
/approve

Leaving the LGTM to other QE members.

damdo:
cc. @sunzhaohua2 @theobarberbany @racheljpg

huali9:
/verified by @huali9 
tested in my local and it works

PR #472, opened 2026-04-02
NO-JIRA: Update WaitForEvent to support events.k8s.io/v1

RadekManak:
Check both core/v1 and events.k8s.io/v1 events, so polling still succeeds on clusters that emit the newer event API.

The older events can be removed later. Support both now to avoid flag day.


RadekManak:
This is not currently necessary. The events published using the new events.k8s.io/v1 can still be read using the old core api.
/close

PR #486, opened 2026-07-09
e2eEnforceNodeGroup

shyadav10:
Jira link: https://redhat.atlassian.net/browse/AUTOSCALE-661

## Summary by CodeRabbit

* **Chores**
  * Updated Go module dependencies (Kubernetes, Cluster API, OpenShift-related CAP components, controller-runtime, and observability libraries) to newer versions for improved compatibility and maintenance.
* **Tests**
  * Added end-to-end scenarios for `ClusterAutoscaler.Spec.EnforceNodeGroupMinSize`: when enabled, scaling increases a `MachineSet` from 0 to meet minimum replicas and reaches Running; when disabled, replicas remain at 0 over a short window.
  * Improved test reporting and cleanup for the `ClusterAutoscaler` lifecycle.


pkg/autoscaler/autoscaler.go
```diff
@@ -1280,4 +1280,174 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, framework.Label
...
+			minReplicas := int32(1)
+			maxReplicas := int32(3)
+
+			By("Creating ClusterAutoscaler with EnforceNodeGroupMinSize enabled")
+
+			clusterAutoscaler = clusterAutoscalerResource(100)
+			enforceMode := caov1.EnforceNodeGroupMinSizeModeEnabled
+			clusterAutoscaler.Spec.EnforceNodeGroupMinSize = &enforceMode
+			Expect(client.Create(ctx, clusterAutoscaler)).Should(Succeed(), "Failed to create ClusterAutoscaler")
+			cleanupObjects[clusterAutoscaler.GetName()] = clusterAutoscaler
```
elmiko: i would move this in to the `BeforeEach` function for this context. that way if we add more tests that need this flag enabled, we can put them in this context.

pkg/autoscaler/autoscaler.go
```diff
@@ -1280,4 +1280,174 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, framework.Label
...
+				current, err := framework.GetMachineSet(ctx, client, machineSet.GetName())
+				if err != nil {
+					return -1, err
+				}
+
+				return *current.Spec.Replicas, nil
+			}, framework.WaitShort, pollingInterval).Should(BeEquivalentTo(int32(0)),
+				"ClusterAutoscaler should not enforce min size when disabled - MachineSet should stay at 0")
+		})
+	})
```
elmiko: i like that you added the negative assertion case here, i think to follow the design pattern in this file i would prefer to see this test in the context that begins on line 182. this way, we can keep the `ClusterAutoscaler` resources in the `beforeEach` for each `Context`.

elmiko (commented):
test looks great @shyadav10 , i'd just like to organize things slightly differently for consistency.

elmiko:
one quick note, we should remove the merge commit from this PR (3cd2ee136c44e66c2924630c1406c9f920591147)

pkg/autoscaler/autoscaler.go
shyadav10: Done :white_check_mark:

pkg/autoscaler/autoscaler.go
shyadav10: Done :white_check_mark:

pkg/autoscaler/autoscaler.go
```diff
@@ -584,6 +584,56 @@ var _ = Describe("Autoscaler should", framework.LabelAutoscaler, framework.Label
 				}, framework.WaitMedium, pollingInterval).Should(BeTrue(), "Node %s has a deletion taint and it should not", machine.Status.NodeRef.Name)
 			}
 		})
+
+		// Machines required for test: 0
+		// Reason: This test verifies that when EnforceNodeGroupMinSize is disabled,
+		// the ClusterAutoscaler does NOT scale up a MachineSet from 0 unless there is workload demand.
+		It("should not enforce minimum size when disabled", func() {
```
elmiko: minor nit, since we moved this let's change the description a little bit to reduce confusion, eg "should not enforce minimum size"

elmiko (commented):
one minor tweak and this is good to go.

pkg/autoscaler/autoscaler.go
shyadav10: done

elmiko:
thanks for all the updates!

/approve
/lgtm

go.mod
hongkailiu, line 9: Do we want to bump it to `v1.13.3` as in Line 29?

hongkailiu: Do we want to bump the following to `v1.13.3` as well?

https://github.com/openshift/cluster-api-actuator-pkg/blob/36a8d884fe63f393532cce22828524a106bd5b22/testutils/go.mod#L15

elmiko: we don't need the updated cluster-api for our tests, but we didn't want to change the pin because we don't own the cluster-capi-operator. cc @damdo and @mdbooth

PR #488, opened 2026-07-14
NO-JIRA: update test for new pod scaleup delay

elmiko:
This change moves the waiting loop for the workload pod to help reduce the possibility of races with the apiserver.


elmiko:
running the autoscaler tests again just to see if it flakes

/test e2e-aws-autoscaler

